PR #94, opened 2024-12-30
chore: fix typo in hyperlink leading to 404

leonardmq:
This PR fixes a typo in the hyperlink to the REST API documentation in the README, which was causing it to lead to a 404.


scosman:
thanks! good catch.

PR #97, opened 2025-01-03
New UI to allow using JSON schema directly when setting a task.

scosman:
## What does this PR do?

JSON schema UI

## Related Issues

https://github.com/Kiln-AI/Kiln/discussions/96#discussioncomment-11719364

## Checklists

### DIFF
--- a/app/web_ui/src/lib/utils/json_schema_editor/json_schema_form_element.svelte
+++ b/app/web_ui/src/lib/utils/json_schema_editor/json_schema_form_element.svelte
@@ -1,12 +1,29 @@
 <script lang="ts">
-  import { empty_schema_model } from "./json_schema_templates"
-  import type { SchemaModel } from "./json_schema_templates"
   import FormElement from "../form_element.svelte"
+  import {
+    type SchemaModel,
+    schema_from_model,
+  } from "$lib/utils/json_schema_editor/json_schema_templates"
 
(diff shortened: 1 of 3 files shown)


app/web_ui/src/lib/utils/json_schema_editor/json_schema_form_element.svelte
```diff
@@ -54,88 +71,177 @@
...
+          Add Property
+        </button>
       </div>
-    {/each}
-    <div class="flex place-content-center">
+    </div>
+  {/if}
+{:else}
+  <div class="flex flex-col gap-8 pt-6" {id}>
+    <FormElement
```
leonardmq, line 207: A button to allow switching back to the Visual Editor could be helpful to prevent users from getting themselves stuck on the raw schema form. Otherwise they would need to refresh the page (and possibly lose their form state in doing so - 'Requirements', etc.) to return to the Visual Editor

scosman: Totally, but the visual editor doesn't support many of the features of JSON Schema, so unfortunately it's a 1 way road. Maybe I'll add a warning "This can't be undone".

scosman: nevermind, added this!

Issue #95, opened 2024-12-30
JSONL fine-tuning files escape non-ASCII characters


leonardmq:
Fine-tuning JSONL file downloads escape non-ASCII characters.

This results in JSONL files containing something like:
```json
{
   "messages":[
      {
         "role":"system",
         "content":"\u4f60"
      },
      ...
   ]
}
```

Instead of:
```json
{
   "messages":[
      {
         "role":"system",
         "content":"你"
      },
      ...
   ]
}
```

By default, json.dumps encodes strings into ASCII, escaping non-ASCII characters. The ASCII-encoded strings are then written to the file using UTF-8 encoding. While the JSON remains valid, opening the file in an editor or loading it into certain tools displays the escaped string instead of the actual characters.

**Proposed fix:** setting `ensure_ascii=False` in the `json.dumps` in `DatasetFormatter` [here](https://github.com/Kiln-AI/Kiln/blob/94668be4693a06ab54c2e7dda3529179b3e52b34/libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py#L185) resolves the issue:
```py
        with open(output_path, "w", encoding="utf-8") as f:
            for run_id in self.dataset.split_contents[split_name]:
                ...
                f.write(json.dumps(example, ensure_ascii=False) + "\n")
```

In some cases, I noticed that a “sample repair” run might also return similarly escaped strings in the dashboard UI. I am unsure if this is just caused by the LLM returning escaped data or if there is another encoding issue elsewhere in the codebase. I could take a closer look to see if there may be some other issue on this front.

If this proposed fix for fine-tuning JSONL files works for you, I’d be happy to open a PR.

scosman:
Ahhh the wonderful world of encodings. 

Thanks for the bug! I’ll take a look shortly.

leonardmq:
Took some time to investigate this a bit more and found that the escaping problem is also present in the other dataset formats (not just the JSONL file download). The impact varies depending on the output format used in the Task.

## Plain text

Unicode characters are escaped (current):
```json
{
  "messages": [
    { "role": "system", "content": "# Instructions here  \n\n" },
    { "role": "user", "content": "\u4eca\u5929..." },
    { "role": "assistant", "content": "\u6211..." }
  ]
}
```

After setting `ensure_ascii=False` in the `json.dumps` calls in `DatasetFormatter`:
```json
{
  "messages": [
    { "role": "system", "content": "# Instructions here  \n\n" },
[...]

PR #102, opened 2025-01-11
fix: unicode is escaped in structured output

leonardmq:
## What does this PR do?

This PR disables escaping unicode characters into ASCII in the structured output formatter which seemed to be causing unicode characters to be passed on to OpenAI as escaped strings in toolcall arguments during fine-tuning. Additionally adds some tests to demonstrate the issue - the newly added tests fail when `ensure_ascii=False` is not present.

## Related Issues

Follow up on: https://github.com/Kiln-AI/Kiln/commit/daf18e62ac87821c23437a1e9c5d0eee8cb1bf29

Fixes: https://github.com/Kiln-AI/Kiln/issues/95 - as it pertains to escaping unicode in structured output arguments.

As illustrated in this [comment](https://github.com/Kiln-AI/Kiln/issues/95#issuecomment-2581395722), fine-tuning jobs for Structured Output tasks currently appear to use up many more `Trained tokens` on OpenAI when the arguments are `json.dumps` without `ensure_ascii`. This suggests that that unicode text contained in toolcall `arguments` is not being decoded by OpenAI and that the fine-tuning is being done over the unicode escape literals, rather than the unicode characters themselves.

The scope of this particular problem is limited to the structured output formatter (`OPENAI_CHAT_TOOLCALL_JSONL`). The reason being that there is double serialization happening there (`arguments` is being treated as a string). Other formatters do not exhibit this issue.

There may be a broader discussion to be had about unicode escaping in general as I have noticed other similar issues but they are hard to debug.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py
+++ b/libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py
@@ -68,7 +68,7 @@ def generate_chat_message_toolcall(
                         "function": {
                             "name": "task_response",
                             # Yes we parse then dump again. This ensures it's valid JSON, and ensures it goes to 1 line
-                            "arguments": json.dumps(arguments),
+                            "arguments": json.dumps(arguments, ensure_ascii=False),
                         },
                     }
                 ],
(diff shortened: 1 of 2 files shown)


scosman:
Looks great, thanks!

PR #103, opened 2025-01-12
fix: unicode converted into escape sequences in repair prompt

leonardmq:
## What does this PR do?

This PR fixes a bug where Unicode characters in structured output are passed back as escaped ASCII literals in the repair prompt. Since the structured output being repaired is serialized into a string, the LLM does not decode it and instead consumes the escape literals rather than the actual characters.

As LLMs have some understanding of escape sequences, this bug breaks the repair output in semi-sporadic, subtle and unpredictable ways - for example, it may inject random characters (often newlines) into the repaired output, or respond with escape sequences instead of the characters.

## Related Issues

This is part of a broader issue related to unicode escaping: https://github.com/Kiln-AI/Kiln/issues/95

## Checklists


scosman:
Thanks @leonardmq !

PR #110, opened 2025-01-15
fix: unicode in list_json_schema_for_task

leonardmq:
## What does this PR do?

This PR fixes:
- bug: a task with structured input will have its schema ASCII-escaped in the sample input generation task (due to `list_json_schema_for_task`), which would cause issues if the schema contains Unicode characters
- test: two existing tests had seemingly contradictory naming, and one test seemed to accidentally be setting `output_json_schema = None` instead of `input_json_schema = None`

## Related Issues

Part of a group of bugs caused by escaped Unicode: https://github.com/Kiln-AI/Kiln/issues/95

This seems like this was the last unicode-related issue I found - the remaining `json.dumps` are not serializing user-supplied data that could contain non-ASCII.

This seems

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/data_gen/data_gen_task.py
+++ b/libs/core/kiln_ai/adapters/data_gen/data_gen_task.py
@@ -163,7 +163,7 @@ def list_json_schema_for_task(task: Task) -> str:
         "required": ["generated_samples"],
     }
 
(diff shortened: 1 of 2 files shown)


scosman:
@leonardmq huge thank you for tracking these down!

PR #131, opened 2025-01-27
EULA + privacy policy improvements

scosman:


scosman:
Privacy policy not complete. Merged EULA separately but PP needs work.

Issue #192, opened 2025-02-11
[Feature Request] Log instructions for docs


scosman:
It would be good to have instructions in our docs for how to get Kiln error logs. Unfortunately I have no idea where/if pyinstaller logs errors to. It's easy to get these in development mode, but most users won't have those.

We should either add:
 - instructions for how to get the logs on each platform (Win/Mac/Linux)
 - code to save them to files (error level) if pyinstaller doesn't do this by default, including ensuring we don't take up too much disk.

Help would be great on this one.

scosman:
Fixed by @leonardmq in https://github.com/Kiln-AI/Kiln/pull/227!

PR #217, opened 2025-02-22
chore: set up logging

leonardmq:
## What does this PR do?

Add config for logging to file / console / both in the uvicorn server. The files are rotated when they reach a max size and default settings can be overwritten via environment variables.

Add new environment variables:
- `KILN_LOG_LEVEL`: Sets logging level (defaults to "INFO")
- `KILN_LOG_FILE`: Custom path for log file (defaults to "~/Kiln Projects/logs/kiln_desktop.log")
- `KILN_LOG_MAX_BYTES`: Maximum log file size before rotation (defaults to 20MB)
- `KILN_LOG_BACKUP_COUNT`: Number of rotated log files to keep (defaults to 3)
- `KILN_LOG_DESTINATION`: Where to send logs - "file", "console", or "all" (defaults to "all")

You can test out this config on the dev server by passing in the `log_config()` here:
```py
# dev_server.py

from app.desktop.log_config import log_config 

// ...

if __name__ == "__main__":
    uvicorn.run(
        "app.desktop.dev_server:dev_app",
        host="127.0.0.1",
        port=8757,
        reload=True,

        log_config=log_config(), # add this
    )
```

## Related Issues

- https://github.com/Kiln-AI/Kiln/issues/192

## Checklists

**I did not add tests**


scosman:
Oh awesome! I'll take a look at this next week but super excited. Thanks @leonardmq !

scosman:
Few questions:

 - Does this capture all `prints` or should we move those to a proper log call with a associated level?
 - Does this work out of the box for fast API errors?
 - Can we make the default level ERROR? Any reason to keep it at info?

leonardmq:
> * Does this capture all `prints` or should we move those to a proper log call with a associated level?

The `print` statements are not going into the file. We need to use the logger for that.

I can do that - should be easy to set up. There are two ways we can go about it:
1. **Option 1**:  Log everything via the root logger (declare `logger = logging.getLogger()` somewhere and use it everywhere)
2. **Option 2**: Declare a logger in each file / module and give it a name (declare `logger = logging.getLogger(__name__)` in each file or module)

With Option 2, we could then also print out the module name in each log as such (here `kiln_server.task_api`), which seems useful to help narrow down where it is coming from:
```log
2025-02-22 18:31:21,610.610 - INFO - kiln_server.task_api - task_from_id: 283184597669, 195277936324
```

Do you prefer per-module / per-file logger or just using the root logger (in which case no module name in the logs)?

> * Does this work out of the box for fast API errors?

Yes it should. 

Here is a sample of the log file - with some errors (raised from inside FastAPI routes) and their traces, as well as some access logs: [kiln_desktop.log](https://github.com/user-attachments/files/18924347/kiln_desktop.log)

> * Can we make the default level ERROR? Any reason to keep it at info?

No particular reason to default to `INFO` - we can change the default to log only `ERROR` (or maybe `WARNING`).

Alternatively we can pass in an env var to overwrite the log level. Not sure if you package up environment variables into the app (not familiar with the details of how the desktop build / bundling works).

Side note - this logging should be used in the prod server, but it is probably preferable to not use it in the dev server, because it gets rid of the colors so the console logs look a little less nice.

leonardmq:
I have swapped out the `print` statements for logger calls with one logger per file and the log file includes the corresponding module name on each log.

Here's what the final form of the log file looks like: [kiln_desktop.log](https://github.com/user-attachments/files/18924786/kiln_desktop.log)

I have included some cases of errors thrown and unhandled (propagating all the way to FastAPI handlers), some handled ones logged with `logger.error("Some message", exc_info=True)`, and some access logs. This was using the dev server.

---

There were a few `print` statements that were made in tests. I have replaced those with logger calls for consistency, and I confirm that they work fine when run with the `-s` flag:
```py
uv run python3 -m pytest . -s
```

Might be worth noting that the logger used in tests uses the default config and not the custom `log_config` one - which seems fine since the log config is more of a prod-specific config.

app/desktop/log_config.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import List
+
+
+class LogDestination(Enum):
+    CONSOLE = "console"
+    FILE = "file"
+    ALL = "all"
+
+
+# TODO: consolidate this with kiln_server.server.project_api.default_project_path
```
scosman: Let's use the settings path? More hidden, which limits logs to more technical users 😆 

Already exposed nicely here: 

```
from kiln_ai.utils.config import Config
Config.settings_path()
```
```

File: app/desktop/log_config.py [hunk 2]
```diff
@@ -0,0 +1,103 @@
...
+def get_default_project_path() -> str:
+    return os.path.join(Path.home(), "Kiln Projects")
+
+
+def get_log_level() -> str:
```
scosman, line 14: You were right - let's default to WARNING

libs/core/kiln_ai/adapters/test_ollama_tools.py
```diff
@@ -1,16 +1,19 @@
...
     json_response = '{"models":[{"name":"scosman_net","model":"scosman_net:latest"},{"name":"phi3.5:latest","model":"phi3.5:latest","modified_at":"2024-10-02T12:04:35.191519822-04:00","size":2176178843,"digest":"61819fb370a3c1a9be6694869331e5f85f867a079e9271d66cb223acb81d04ba","details":{"parent_model":"","format":"gguf","family":"phi3","families":["phi3"],"parameter_size":"3.8B","quantization_level":"Q4_0"}},{"name":"gemma2:2b","model":"gemma2:2b","modified_at":"2024-09-09T16:46:38.64348929-04:00","size":1629518495,"digest":"8ccf136fdd5298f3ffe2d69862750ea7fb56555fa4d5b18c04e3fa4d82ee09d7","details":{"parent_model":"","format":"gguf","family":"gemma2","families":["gemma2"],"parameter_size":"2.6B","quantization_level":"Q4_0"}},{"name":"llama3.1:latest","model":"llama3.1:latest","modified_at":"2024-09-01T17:19:43.481523695-04:00","size":4661230720,"digest":"f66fc8dc39ea206e03ff6764fcc696b1b4dfb693f0b6ef751731dd4e6269046e","details":{"parent_model":"","format":"gguf","family":"llama","families":["llama"],"parameter_size":"8.0B","quantization_level":"Q4_0"}}]}'
     tags = json.loads(json_response)
-    print(json.dumps(tags, indent=2))
+    logger.info(json.dumps(tags, indent=2))
```
scosman: I should probably just delete this 😀

libs/core/kiln_ai/adapters/test_prompt_builders.py
```diff
@@ -269,7 +272,7 @@ def test_few_shot_prompt_builder(tmp_path):
                 rating=TaskOutputRating(value=4 + (i % 2), reason="Good joke"),
             ),
         )
-        print("RATING", "Joke Initial Output ", i + 1, " - RATED:", 4 + (i % 2), "\n")
+        logger.info("RATING: Joke Initial Output %s - RATED: %s", i + 1, 4 + (i % 2))
```
scosman: I should delete the ones in this file too.

scosman (commented):
Okay, this is awesome. Thanks again Leo!

Some small tweaks suggested, but LGTM after these?

app/desktop/log_config.py
leonardmq: Makes sense, I have updated to match - I had to make a small change to `settings_path()` because it was the path of the settings file (rather than the path to its directory)

app/desktop/log_config.py
leonardmq: Updated to default to `WARNING`.

Wondering what you think of the `KILN_LOG_DESTINATION` defaulting to `all` (log to both `console + file`) - might be better to default to `file` only?

libs/core/kiln_ai/adapters/test_ollama_tools.py
leonardmq: Deleted that line - I wondered why it was there lol

libs/core/kiln_ai/adapters/test_prompt_builders.py
leonardmq: Deleted that line

scosman:
Good to merge @leonardmq ?

leonardmq:
> Good to merge @leonardmq ?

Ok for me @scosman

app/desktop/log_config.py
scosman: Let's keep all! Most folks won't run from console, so it doesn't matter. Those that do probably will like it.

scosman (approved):

scosman:
LFTM. FYI merging into a staging branch first to trigger full CI. Not sure why it isn't running here.

PR #226, opened 2025-02-26
Eval config eval

scosman:
New system to evaluate the evaluators agains human baselines.


scosman:
Created for incorrect branch. See https://github.com/Kiln-AI/Kiln/pull/228

PR #238, opened 2025-03-08
feat: import samples from CSV file

leonardmq:
## What does this PR do?

This PR adds a `Upload dataset` feature to let users import samples / runs from a CSV file, and paves the way for more formats.

Changes:
- add button on the `Datasets` page that triggers a modal
- add `POST /api/projects/{project_id}/tasks/{task_id}/runs/bulk_upload` API route
- add dataset import handling
- add tests for the core import

Here's a screen recording showing the feature - note I added some intentional sleep on the server side to demonstrate the loading state, but otherwise it completes instantaneously:

https://github.com/user-attachments/assets/db1d9d83-02af-477a-bafa-c7c31b7a44ff

### Structure

The CSV structure must have a header row at the top, and fields must be escaped where necessary; the supported fields are:
- `input`: the run input - can be serialized JSON
- `output`: the run output - can be serialized JSON
- `reasoning` (optional): if using intermediate output
- `tags` (optional): comma-separated list of tags

Example - note that the ordering of columns does not matter:
```csv
input,output,reasoning,tags
"My name is John","{""joke"": ""They call me John Wick... but my bank calls me 'insufficient funds'.""}","The user introduced themselves, so we play along with a joke on the name 'John'.","funny"
"My name is Jane","{""joke"": ""Jane? Like Tarzan's Jane? Hope you're ready for some wild adventures!""}","Making a playful reference to the famous Tarzan and Jane pairing.","hilarious"
"My name is Jack","{""joke"": ""Jack? Hope you don't lose your spot on the door like in Titanic.""}","Classic reference to 'Jack' from Titanic, a common cultural joke.","funny,movie"
"My name is Jill","{""joke"": ""Jill? Where's Jack? I hope you two didn't tumble down a hill!""}","Classic nursery rhyme reference to Jack and Jill.",
"My name is Jack","{""joke"": ""Jack? Like Jack Sparrow? Where’s the rum?""}","Pirates of the Caribbean joke for name recognition.",
```

### Tagging

In addition to the tags that may optionally be provided in the CSV file, the import will assign two tags to the `TaskRuns` created: `imported`, and `imported_{timestamp}`.

### Invalid upload

If the file provided is invalid, the dataset importer will not create any tasks, and will raise a `KilnInvalidImportFormat` exception that refers to the particular row that caused the exception to be raised. 

The `KilnInvalidImportFormat` is handled at the API controller layer, and returned to the frontend with a `422` and then frontend displays it:
image

## Related Issues

None.

## Checklists


leonardmq:
I was not sure if these files belong in the `adapters/` directory.

As the `dataset_import.py` file is already a bit bloated with CSV utils and validation helpers, I was thinking about restructuring this:
- add a folder `data_import/` and have one file per importer (e.g. `csv_importer.py`, etc.)
- add an `AbstractImporter` with a `validate()` and `create_runs()` method, and have each importer be an implementation of that class (instead of each importer being just a function), which could help avoid having the fairly large `Importer` functions
- add a `data_import_registry.py` wrapper where we register the mapping from import type to the corresponding importer

In terms of the UI, I kept the logic fairly self-contained in the `upload_dataset_dialog.svelte` component so it should be easy to move around if needed.

The design idea with this is that more importers could be easily added later - both UI and backend should support that fairly easily. In terms of potential formats that could be supported, maybe the OpenAI JSONL format used for fine-tuning could be a useful one.

Let me know what you think

scosman:
Amazing. I'll take a look next week!

scosman:
UI is nice and clean and follows Kiln style. Some small UI tweaks would be good.

- The “Upload Dataset” button should be the the AppPage buttons section - it has a standard function for adding page actions with consistent UI (it will be higher up, aligned to title). 
- Let’s remove the radio button since we only support CSV
- “Selected: filename” seems redundant since you can see it right above 
- Let’s put a little helper text in the modal/dialog explaining the fields
  - “Upload a CSV to add each row to your dataset. The CSV must have a header row. The the following columns are supported: input, output, reasoning (optional), tags (optional, comma separated string).”

What do you think?

Looking at code next!

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,139 @@
+<script lang="ts">
```
scosman: Nit: we can move this into `/dataset/[project_id]/[task_id]`. The Lib folder is mostly for components we want to reuse, and this is fairly single purpose.

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,139 @@
...
+  export let onImportCompleted: () => void
+
+  const dataset_type_accepts = {
+    csv: [".csv"],
+  }
+
+  let error: KilnError | null = null
+  let loading = false
+  let selected_file: File | null = null
+  let dataset_type: "csv" | null = "csv"
```
scosman: let's remove this an associated radio/UI? I don't see us adding any more file formats for a long time, and if we do we'll figure out UI then.

libs/server/kiln_server/run_api.py
```diff
@@ -257,6 +274,61 @@ async def edit_tags(
             )
         return {"success": True}
 
+    @app.post("/api/projects/{project_id}/tasks/{task_id}/runs/bulk_upload")
+    async def bulk_upload(
+        project_id: str,
+        task_id: str,
+        dataset_type: DatasetImportFormat = Form(...),
```
scosman: I'm sure uploading files from a browser is a pain, so if you tried stuff and found this was best way let me know and disregard this.

Usually for a post you can create a new pydantic object with all the fields. Might clean up the typing issues in the JS OpenAPI client (the @ts-expect-error)?

```
class BulkUploadRequest(BaseModel):
   dataset_type: DatasetImportFormat
   file: UploadFile = File(...)
```

Slight preference for keeping the API json based, and not using FormData, but if that's not possible with browser uploads let me know.

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,139 @@
...
+        throw error
+      }
+
+      selected_file = null
+
+      onImportCompleted()
+
+      return true
+    } catch (e) {
+      console.error("Upload failed:", e)
```
scosman: remove the log since we have UI for errors?

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,139 @@
...
+
+  function handleCancel() {
+    close()
+    return true
+  }
+</script>
+
+<Dialog
+  bind:this={dialog}
+  title="Upload Dataset"
```
scosman: "Upload CSV to Dataset" is a bit clearer - filetype and no risk people will think it will replace existing dataset.

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,139 @@
...
+        {#if selected_file}
+          <p class="mt-2 text-sm text-gray-500">
+            Selected: {selected_file.name}
+          </p>
+        {/if}
+      {/if}
+    {/if}
+  </div>
+  {#if error}
+    <div class="text-error text-sm">
```
scosman: nit: `pt-2`

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/+page.svelte
```diff
@@ -484,6 +496,11 @@
   sub_subtitle_link="https://docs.getkiln.ai/docs/organizing-datasets"
   no_y_padding
 >
+  <div class="flex flex-row justify-end">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/+page.svelte]: dupe of other comment, but yo can put this action into the AppPage action_buttons, and it will make the layout consistent with other pages.

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
+import csv
```
scosman: I think this file can be in "utils". It's not really an adapter. Not that this folder is super clean 😓

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: I was thinking the dropdown could be used to add support for the JSONL format that fine-tuning are often exported to (for OpenAI for example). I can remove it from the UI but keep the extensible logic on the backend in case we decide to support more formats later.

What do you think? Remove from UI but keep extensible on the backend side?

scosman:
FYI - I have a review in progress with a few more smaller comments. Away from laptop but will try to finish asap

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
...
+        dataset_name: str,
+    ) -> int: ...
+
+
+class CSVRowSchema(BaseModel):
+    """Schema for validating rows in a CSV file."""
+
+    input: str = Field(description="The input to the model")
+    output: str = Field(description="The output of the model")
+    reasoning: str | None = Field(
```
scosman: P2: we could also add chain_of_thought.

We store these under different keys in a TaskRun intermediate_output, since they are different things.

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
...
+        validate_default=True,
+    )
+
+    @field_validator("tags", mode="before")
+    def split_tags(cls, value: str | None) -> list[str]:
+        # Handle missing tags column or empty value
+        if not value:
+            return []
+        # Handle string values (comma-separated tags)
+        if isinstance(value, str):
```
scosman: I'd move this out of a model validator and into where this is set. A List[str] shouldn't ever be assigned a str, and would be good for type catcher to catch/fix that. Make the caller call strip/split and assign the right type.

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
...
+    ]
+
+
+class KilnInvalidImportFormat(Exception):
+    """Raised when the import format is invalid"""
+
+    def __init__(self, message: str, row_number: int | None = None):
+        self.row_number = row_number
+        if row_number is not None:
+            message = f"Error in row {row_number}: {message}"
```
scosman: 👍

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
...
+    row_number: int,
+    session_id: str,
+) -> TaskRun:
+    """Validate and create a TaskRun from a CSV row, without saving to file"""
+
+    # first we validate the row from the CSV file
+    try:
+        validated_row = CSVRowSchema.model_validate(row)
+    except ValidationError as e:
+        logger.warning(f"Invalid row {row_number}: {row}", exc_info=True)
```
scosman: we have this same `except ValidationError` logic twice. Remove it from this method, wrap the call to create_task_run_from_csv_row with a try, and put this exception logic there? If we add future exceptions (or accidental) in this method, it would catch them as well. No need for any try/except in this method.

This also means we don't need row_number as a parameter (only the caller needs to know about the context).

libs/core/kiln_ai/adapters/dataset_import.py
```diff
@@ -0,0 +1,230 @@
...
+    if validated_row.tags:
+        tags.extend(validated_row.tags)
+
+    # now we create the task run, which may raise pydantic validation errors
+    try:
+        run = TaskRun(
+            parent=task,
+            input=validated_row.input,
+            input_source=DataSource(
+                type=DataSourceType.file_import,
```
scosman: We have validators for the DataSource mode that validate properties. We should add file_name there and make it required, and also make the other properties not_allowed_for DataSourceType.file_import

libs/server/kiln_server/run_api.py
```diff
@@ -257,6 +274,61 @@ async def edit_tags(
...
+            importer = DatasetFileImporter(
+                task,
+                ImportConfig(
+                    dataset_type=dataset_type,
+                    dataset_path=file_path,
+                    dataset_name=file_name,
+                ),
+            )
+            imported_count = importer.create_runs_from_file()
+        except KilnInvalidImportFormat as e:
```
scosman, line 306: Can we just remove the try/except here?

We can let the other exception bubble up. FastAPI will catch exceptions, log and return a 500. Nice to keep the code short/clean and get same result.

I think it will also catch the 422 automatically before reaching this since we use an enum field (it validates input and raises an error). But either way the exception would bubble up and be logged/returned without the extra code.

leonardmq: The intent here was to translate the `KilnInvalidImportFormat` error into a 422. This error is raised during `import_csv(...)` if the file structure or a row fails parsing - so it is different from the `DatasetImportFormat` enum validation in the route (assuming that is the enum you were referring to).

But I should remove the generic `except Exception as e` further down, since FastAPI will handle it as a 500, like you mentioned. Were you suggesting just that, or removing the entire try/catch block? If we remove the whole try/catch entirely, invalid file parsing errors would bubble up as a 500, which might not be ideal.

scosman: Keeping the 422 makes sense. Returning a 4xx when the inputs are wrong is a good call.

libs/server/kiln_server/run_api.py
leonardmq: I have tried a few combinations but could not get the client SDK to work. OpenAPI describes the file as `string` with `binary` format. The `openapi-typescript` codegen expects a `string` for such schemas. 

According to [this issue](https://github.com/openapi-ts/openapi-typescript/issues/1214), using their `transform` API to specify how to type-map a multipart form upload seems to be the recommended approach. The docs about this here: https://openapi-ts.dev/node#example-blob-types . It seems like it would require some changes to how the codegen is currently being used as we use it as CLI in `generate_schema.sh` and the dynamic `transform` looks like it runs differently from the CLI.

If we want the API to be truly JSON, we could also serialize the file to base64, but it seems inefficient as files may be large

libs/core/kiln_ai/adapters/dataset_import.py
leonardmq: ✅  - just an FYI - moving this to `utils/` creates a circular import when exporting it in `utils/__init__.py`, since `basemodel.py` depends on `utils`. So, I still moved `dataset_import.py` into `utils/`, but left it out of the `__init__.py` to avoid the issue. Let me know if this works for you, or if you’d prefer we put it elsewhere. It might be best to keep `utils/` free from core dependencies.

libs/core/kiln_ai/adapters/dataset_import.py
leonardmq: ✅ 

(to be fair, the typing did catch this initially and I kept trying different things to get around it thinking there was a legitimate way to do a transform like this (😵‍💫). Unwrapping at the caller makes a lot more sense)

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: ✅

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: ✅ - removed from the UI and the API, but kept most of the logic on the backend

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: ✅

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: ✅

app/web_ui/src/lib/ui/upload_dataset_dialog.svelte
leonardmq: ✅

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/+page.svelte]: ✅

libs/core/kiln_ai/adapters/dataset_import.py
leonardmq: ✅ - makes a lot more sense; I think the original intent was to figure out if the error was from the CSV format itself that was problematic, or the actual semantics of it (rejected by the task run creation validation), but I guess we get it in the stack trace anyway

libs/server/kiln_server/run_api.py
leonardmq: ✅

leonardmq:
> * Let’s put a little helper text in the modal/dialog explaining the fields
>   
>   * “Upload a CSV to add each row to your dataset. The CSV must have a header row. The the following columns are supported: input, output, reasoning (optional), tags (optional, comma separated string).”

I added a helper with the text explaining this and changed to `font-light` to align with similar UIs elsewhere in the code:

image

scosman:
Awesome, will try to 👀  and merge soon!

leonardmq:
@scosman - thanks for the review - and there are two comments (marked with 👀 ) that I have not addressed yet but am planning to in the next couple of days:
- adding `chain_of_thought` -> sounds easy to add, but I want to test that out
- adding the validators for the `DataSource` -> I found where this is going on and get the idea, but need to look into the details

libs/core/kiln_ai/adapters/dataset_import.py
leonardmq: ✅

libs/core/kiln_ai/adapters/dataset_import.py
leonardmq: ✅ - added `chain_of_thought` (also in the modal UI that lists out the fields)

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte
```diff
@@ -0,0 +1,135 @@
...
+          <ul class="mb-4 ml-4 mt-1 list-disc">
+            <li><code>input</code> - Required</li>
+            <li><code>output</code> - Required</li>
+            <li><code>reasoning</code> - Optional</li>
+            <li><code>chain_of_thought</code> - Optional</li>
+            <li><code>tags</code> - Optional, comma separated string</li>
+          </ul>
+          <p class="mb-4">
+            See an example <a
+              href="#link-to-example-on-github"
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte]: @scosman 

I have a couple of example datasets that demonstrate the structure, I was thinking we can put them somewhere in the repo and link to them from this UI. The CSV structure is easy enough, but might not be obvious how JSON should be presented in it or how to escape the fields.

Is there any directory where we could put those example files?

leonardmq:
All the comments should be addressed now.

I also modified my dialog UI to rely more on the `Dialog` component for the loading state and error handling. My component was handling it itself before this commit: https://github.com/Kiln-AI/Kiln/pull/238/commits/8952ac19b038e12a43aeccdc3da787cca299703d - but it should be more uniform now

libs/server/kiln_server/run_api.py
scosman: Makes sense. No need to spend any more time on this. This stuff is always a bit of a pain. Thanks for trying

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte]: I'd say your UI is pretty great. I don't think people will need examples, but if they do the docs are the most discoverable place. I'll add a "Importing CSV" section to [this page](https://docs.getkiln.ai/docs/organizing-datasets), if you want to send some examples I can include them.

scosman:
Really fantastic work @leonardmq ! Merging.

libs/core/kiln_ai/adapters/dataset_import.py
scosman: Ahh, makes sense. Let's do a new /importers folder in adapters? Folks might want to use this in the python API and it will be excluded from docs and harder to import if we don't include it in __init__.py

I already merged. I can do this. Or you. Whoever gets to it 😆

Issue #248, opened 2025-03-16
[Bug] Run imported from CSV cannot be repaired


leonardmq:
**Describe the bug**
The run repairing flow does not work on CSV-imported task runs. This is because the `file_import` data source does not have a `prompt_id`, so it fails to create the repair prompt.

Error shown in the UI is: `Unexpected error: Prompt builder 'None' is not a valid prompt builder`.

In terms of a fix, not immediately clear where the best place for the user to indicate the prompt type should be - possibly this is a piece of information that they should include in the CSV as a field with possibly a fallback to basic zero shot prompt type if left blank. Alternatively in the CSV import modal, through a dropdown but it does not really work if the samples they import span different types of prompts.

**Checks**
- [x] I've read the [troubleshooting guide](https://docs.getkiln.ai/docs/troubleshooting-and-logs)
- [x] - n/a - I've tried to reproduce the problem using another model, and confirmed it's not an issue specific to the model I've chosen.
- [x] I've searched [the docs](https://docs.getkiln.ai) for a solution
- [x] I've searched for existing Github issues/discussions

**To Reproduce**

1. Go to `Dataset`
2. Click on `Upload File`
3. Choose a valid file in the modal and click `Upload`
4. Click on one of the newly added rows
5. Rate it `2` stars
6. Fill in the `Repair Instructions` text area in `Repair Output`
7. Click `Attempt Repair`
8. UI shows an error: `Unexpected error: Prompt builder 'None' is not a valid prompt builder`

**Expected behavior**
Clicking `Attempt Repair` on a CSV-imported task run should attempt to repair it. 

**Screenshots**
<img width="1221" alt="Image" src="https://github.com/user-attachments/assets/a511bded-807f-410e-a68e-6cd947527c22" />

**Error Logs**
```log
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/xxx/Kiln/.venv/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 406, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/xxx/Kiln/.venv/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
[...]

PR #254, opened 2025-03-18
fix: cannot repair CSV-imported runs

leonardmq:
## What does this PR do?

Add fallback to use simple prompt builder during repair if the `TaskRun` does not have a `prompt_id` or `prompt_builder_name` - as in the case of a CSV-imported `TaskRun`. This fixes part of the issue that makes CSV-imported `TaskRun` not repairable.

## Related Issues

https://github.com/Kiln-AI/Kiln/issues/248

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/repair/repair_task.py
+++ b/libs/core/kiln_ai/adapters/repair/repair_task.py
@@ -1,11 +1,10 @@
 import json
-from typing import Type
 
 from pydantic import BaseModel, Field
 
 from kiln_ai.adapters.prompt_builders import (
     BasePromptBuilder,
-    SavedPromptBuilder,
+    PromptGenerators,
     prompt_builder_from_id,
 )
 from kiln_ai.datamodel import Priority, Project, Task, TaskRequirement, TaskRun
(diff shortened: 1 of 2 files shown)


leonardmq:
@scosman - There now is another problem that might require some UI. The `Repair` API requires a `model_name` and `provider`. The UI for a `synthetic` or `human` runs uses the original run's model / provider; but CSV-imported (as well as future manual runs) ones do not have such properties. 

image

Probably need an extra piece of UI to let the user select the model they want to use for the repair. It may make sense also in general because some repairs might be better done with a different model than the one used to generate the original (broken) output.

I was thinking a similar form similar to that one:
image

Maybe in the sidebar, and call it `Repair Options` - and save the selection in local storage to avoid requiring the user to pick a model on each repair.

scosman:
Oh right, good catch.

Hmmm. I'm torn on just disabling repair for imports or a UI like you suggest.  Simplicity is good....

leonardmq:
> Oh right, good catch.
> 
> Hmmm. I'm torn on just disabling repair for imports or a UI like you suggest. Simplicity is good....

I was thinking a UI like this could work reasonably well - default it to the run's model / provider if there is one, but let the user pick another model if they want to or need to:

image

But also agree possibly confusing. Maybe disable the repair for imports for now, then allow manual repairing for imports when manual repair is added; and then reassess?

---

I opened a [separate PR](https://github.com/Kiln-AI/Kiln/pull/259) for disabling the repair UI. Does not hurt to have it disabled regardless of what the permanent solution eventually is.

leonardmq:
@scosman 

Possibly best to close that one now that there is a temporary mitigation measure in place (showing the user that manually imported runs cannot be repaired)?

scosman:
@leonardmq we want to close this correct? We just disabled repair on inputs and are leaving it at that IIRC?

leonardmq:
@scosman - Yes, we should close it because the extra work and UI implications to get it to make it useful is probably not worth it for now. 

Can I close it or do you prefer closing yourself?

scosman:
No preference but I got it!

PR #259, opened 2025-03-19
fix: hide repair for imported runs

leonardmq:
## What does this PR do?

Follow up on https://github.com/Kiln-AI/Kiln/pull/254 which partially resolved the issue preventing the repair UI from working on `file_import` runs. However, additional downstream issues were discovered.

This PR disables the repair UI for runs that are neither `synthetic` nor `human`. Instead of an unusable repair form, users will see a warning indicating that repair is unavailable. Even if this UI is replaced later with a working repair UI, this PR already improves the current behavior. This PR also hides the empty fields in the run `Parameters` block in the sidebar.

The warning is only displayed if the user has rated the run `< 5 stars`. The rationale for showing a warning rather than nothing is that users are more likely to report repairing imported runs as a feature they need if they know it is normally supposed to be there.

image

## Related Issues

https://github.com/Kiln-AI/Kiln/issues/248

## Checklists

### DIFF
--- a/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
+++ b/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
@@ -49,6 +49,7 @@
       )
     }
 
+    const model_id = run?.output?.source?.properties?.model_name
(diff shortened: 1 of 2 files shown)


scosman:
👍 Thanks!

Issue #115, opened 2025-01-15
Allow manual data entry/correction


scosman:
From reddit: https://www.reddit.com/r/LocalLLaMA/comments/1i1ffid/comment/m78a9zr/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

PR #261, opened 2025-03-20
feat: allow manual repair edit

leonardmq:
## What does this PR do?

This PR adds support for manually editing a task run repair. After `Attempt repair`, the user has a new button (`Edit`) that turns the repair output display into a text area that allows users editing the repair attempt.

This PR also fixes the missing validation of the repair output in the repair API. There was a bug that allowed repair outputs that do not match the structured output schema, which might have caused downstream problems due to corrupt shapes.

This is an acceptable solution but it has a major limitation due to requiring the user to first `Attempt (automatic) repair`. In the case of a `TaskRun` accessed from the `Dataset` page, the user does not have the ability to select the model used during the repair - it is by default the same as the one that generated the original run output. If a run used a small model (e.g. the small Llama models), the user may be stuck as the small model may fail to even pass validation during `Attempt repair`, which means the user then never gets access to the `Edit`.

New `Edit` button:
image

`Edit` mode:
image

Error state (if JSON does not match schema):
image

## Related Issues

Manual edit requested: https://github.com/Kiln-AI/Kiln/issues/115

## Checklists

### DIFF
--- a/app/desktop/studio_server/repair_api.py
+++ b/app/desktop/studio_server/repair_api.py
@@ -85,7 +85,8 @@ def connect_repair_api(app: FastAPI):
             validate_schema(json.loads(output), task.output_json_schema)
 
         # manually edited runs are human but the user id is not set
-        if input.repair_run.output.source.type == DataSourceType.human:
+        source = input.repair_run.output.source
+        if not source or source.type == DataSourceType.human:
             input.repair_run.output.source = DataSource(
                 type=DataSourceType.human,
                 properties={"created_by": Config.shared().user_id},
--- a/app/desktop/studio_server/repair_api.py
+++ b/app/desktop/studio_server/repair_api.py
@@ -1,7 +1,10 @@
+import json
+
 from fastapi import FastAPI, HTTPException
 from kiln_ai.adapters.adapter_registry import adapter_for_task
 from kiln_ai.adapters.repair.repair_task import RepairTaskRun
 from kiln_ai.datamodel import TaskRun
+from kiln_ai.datamodel.json_schema import validate_schema
 from kiln_server.run_api import model_provider_from_string, task_and_run_from_id
 from pydantic import BaseModel, ConfigDict, Field
 
@@ -73,6 +76,12 @@ def connect_repair_api(app: FastAPI):
         project_id: str, task_id: str, run_id: str, input: RepairRunPost
     ) -> TaskRun:
         task, run = task_and_run_from_id(project_id, task_id, run_id)
+
+        # the repair must match the task's output schema if structured output is enabled
+        if task.output_json_schema:
+            output = input.repair_run.output.model_dump().get("output", "")
+            validate_schema(json.loads(output), task.output_json_schema)
+
         # Update the run object atomically, as validation will fail setting one at a time.
         updated_data = run.model_dump()
         updated_data.update(
--- a/app/desktop/studio_server/repair_api.py
+++ b/app/desktop/studio_server/repair_api.py
@@ -79,11 +79,6 @@ def connect_repair_api(app: FastAPI):
     ) -> TaskRun:
         task, run = task_and_run_from_id(project_id, task_id, run_id)
 
-        # the repair must match the task's output schema if structured output is enabled
-        if task.output_json_schema:
-            output = input.repair_run.output.model_dump().get("output", "")
-            validate_schema(json.loads(output), task.output_json_schema)
-
         # manually edited runs are human but the user id is not set
         source = input.repair_run.output.source
         if not source or source.type == DataSourceType.human:
--- a/app/desktop/studio_server/repair_api.py
+++ b/app/desktop/studio_server/repair_api.py
@@ -5,6 +5,8 @@ from kiln_ai.adapters.adapter_registry import adapter_for_task
 from kiln_ai.adapters.repair.repair_task import RepairTaskRun
 from kiln_ai.datamodel import TaskRun
 from kiln_ai.datamodel.json_schema import validate_schema
+from kiln_ai.datamodel.task_output import DataSource, DataSourceType
+from kiln_ai.utils.config import Config
 from kiln_server.run_api import model_provider_from_string, task_and_run_from_id
 from pydantic import BaseModel, ConfigDict, Field
 
@@ -81,6 +83,13 @@ def connect_repair_api(app: FastAPI):
         if task.output_json_schema:
             output = input.repair_run.output.model_dump().get("output", "")
             validate_schema(json.loads(output), task.output_json_schema)
+
+        # manually edited runs are human but the user id is not set
+        if input.repair_run.output.source.type == DataSourceType.human:
+            input.repair_run.output.source = DataSource(
+                type=DataSourceType.human,
+                properties={"created_by": Config.shared().user_id},
+            )
 
         # Update the run object atomically, as validation will fail setting one at a time.
         updated_data = run.model_dump()
--- a/app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
+++ b/app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
@@ -2,26 +2,17 @@
   import FormElement from "$lib/utils/form_element.svelte"
   import type { Task, TaskRun } from "$lib/types"
   import FormContainer from "../../../lib/utils/form_container.svelte"
+  import { createKilnError, KilnError } from "../../../lib/utils/error_handlers"
+  import { client } from "../../../lib/api_client"
 
   export let on_submit: (run: TaskRun) => void
   export let on_cancel: () => void
 
+  export let project_id: string
   export let task: Task
+  export let run: TaskRun
   export let repair_run: TaskRun
-
-  function validate_output(text: string) {
-    if (task.output_json_schema) {
-      try {
-        // TODO: validate against the JSON schema instead of just checking if it is valid JSON
-        JSON.parse(text)
-        return true
-      } catch (err) {
-        return false
-      }
-    }
-
-    return true
-  }
+  export let repair_instructions: string
 
   function prettify_output(text: string) {
     if (task.output_json_schema) {
@@ -34,11 +25,12 @@
     return text
   }
 
-  $: repair_output_edited_valid = validate_output(repair_output_edited)
   let repair_output_edited = prettify_output(repair_run.output.output) || ""
 
-  function handle_submit() {
-    const human_edited_run: TaskRun = {
+  let post_repair_error: KilnError | null = null
+  let post_repair_submitting = false
+  async function handle_submit() {
+    repair_run = {
       ...repair_run,
       output: {
         ...repair_run.output,
@@ -47,12 +39,53 @@
           type: "human",
           properties: {
             // the user name will be replaced with the actual user name on the server side
-            created_by: "user",
+            created_by: "unknown",
           },
         },
       },
     }
-    on_submit(human_edited_run)
+
+    post_repair_error = null
+    post_repair_submitting = true
+
+    try {
+      if (!repair_run) {
+        throw new KilnError("No repair to edit", null)
+      }
+      if (!task.id || !run.id) {
+        throw new KilnError(
+          "This task run isn't saved. Enable Auto-save. You can't edit repairs for unsaved runs.",
+          null,
+        )
+      }
+      const {
+        data, // only present if 2XX response
+        error: fetch_error, // only present if 4XX or 5XX response
+      } = await client.POST(
+        "/api/projects/{project_id}/tasks/{task_id}/runs/{run_id}/repair",
+        {
+          params: {
+            path: {
+              project_id: project_id,
+              task_id: task.id,
+              run_id: run.id,
+            },
+          },
+          body: {
+            repair_run: repair_run,
+            evaluator_feedback: repair_instructions || "",
+          },
+        },
+      )
+      if (fetch_error) {
+        throw fetch_error
+      }
+      on_submit(data)
+    } catch (err) {
+      post_repair_error = createKilnError(err)
+    } finally {
+      post_repair_submitting = false
+    }
   }
 
   function handle_cancel() {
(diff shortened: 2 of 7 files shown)


leonardmq:
WIP because I want to add a couple of tests in the repair API

app/desktop/studio_server/repair_api.py
```diff
@@ -73,6 +76,12 @@ async def post_repair_run(
         project_id: str, task_id: str, run_id: str, input: RepairRunPost
     ) -> TaskRun:
         task, run = task_and_run_from_id(project_id, task_id, run_id)
+
+        # the repair must match the task's output schema if structured output is enabled
+        if task.output_json_schema:
```
leonardmq: I have not found a centralized utility that could be used to do the validation for this. The validation in the adapter does not seem to be usable outside the context of a generation run.

Is there a better existing method to check the object against this task's output schema?

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -343,6 +343,25 @@
...
+
+  function hide_repair_edit() {
+    repair_edit_mode = false
+  }
+
+  function apply_manual_edit() {
+    if (!repair_run) {
+      return
+    }
+    repair_run.output.output = repair_output_edited
```
leonardmq: The way it currently works is that the repair run would still be attributed to whichever model generated the repair output that was repaired. Would you rather want the repair run to be different?

scosman:
ohhh, can't wait!

Let's pretty print for the text-area if it's JSON? Makes it easier to edit. Basically: `JSON.stringify(obj, null, 2)`

app/web_ui/src/routes/(app)/run/run.svelte
scosman: yeah, we should attribute it to the person making the edit. A standard Human datasource should work!

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+  export let on_submit: (run: TaskRun) => void
+  export let on_cancel: () => void
+
+  export let task: Task
+  export let repair_run: TaskRun
+
+  function validate_output(text: string) {
+    if (task.output_json_schema) {
+      try {
+        // TODO: validate against the JSON schema instead of just checking if it is valid JSON
```
leonardmq: @scosman - I'd want to validate the schema on the client-side here, but the webapp does not have a library for JSON schema validation. 

Is there any particular library you would prefer to use for JSON schema validation on the client side?

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+  $: repair_output_edited_valid = validate_output(repair_output_edited)
+  let repair_output_edited = prettify_output(repair_run.output.output) || ""
+
+  function handle_submit() {
+    const human_edited_run: TaskRun = {
+      ...repair_run,
+      output: {
+        ...repair_run.output,
+        output: repair_output_edited,
+        source: {
```
leonardmq, line 38: I tried `source: null` and `source: undefined` instead of prefilling like this, but the API validation does not allow that

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+            d="M 27.9999 51.9063 C 41.0546 51.9063 51.9063 41.0781 51.9063 28 C 51.9063 14.9453 41.0312 4.0937 27.9765 4.0937 C 14.8983 4.0937 4.0937 14.9453 4.0937 28 C 4.0937 41.0781 14.9218 51.9063 27.9999 51.9063 Z M 24.7655 40.0234 C 23.9687 40.0234 23.3593 39.6719 22.6796 38.8750 L 15.9296 30.5312 C 15.5780 30.0859 15.3671 29.5234 15.3671 29.0078 C 15.3671 27.9063 16.2343 27.0625 17.2655 27.0625 C 17.9452 27.0625 18.5077 27.3203 19.0702 28.0469 L 24.6718 35.2890 L 35.5702 17.8281 C 36.0155 17.1016 36.6249 16.75 37.2343 16.75 C 38.2655 16.75 39.2733 17.4297 39.2733 18.5547 C 39.2733 19.0703 38.9687 19.6328 38.6640 20.1016 L 26.7577 38.8750 C 26.2421 39.6484 25.5858 40.0234 24.7655 40.0234 Z"
+          /></svg
+        >
+        Structure Valid
+      </div>
+    {:else}
+      <div
+        class="text-xs font-medium text-error flex flex-row mb-2 items-center"
+      >
+        <svg
```
leonardmq: This SVG comes from the `Warning` component

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+  function handle_cancel() {
+    on_cancel()
+  }
+</script>
+
+<div>
+  {#if task.output_json_schema}
+    {#if repair_output_edited_valid}
+      <div class="text-xs font-medium text-gray-500 flex flex-row mb-2">
+        <svg
```
leonardmq: This SVG is the same used elsewhere in the page to show `Structure Valid`.

Maybe worth wrapping each SVG into components and giving them names

leonardmq:
@scosman 

Added the following changes:
- add a `disabled` state on the `FormContainer` to prevent submitting
- prevent submitting form if the `submit` button is not visible
- add server-side validation of the repair output if the task has structured output
- refactor repair edit form:
  - use the `FormContainer`
  - prettify initial value
  - add client-side validation if the task has structured output
  - sets the `human` source (prefilled on frontend, and user id is set on the server side)
- test for manual repair source

---

https://github.com/user-attachments/assets/e24f1673-631a-4f33-8eb6-0bcea4fac52f

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
scosman: We can leave validation to server! We just shouldn't save it without validation, and should make a decent error message there.

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
scosman: totally. I try to do that when I duplicate it.... but as you can tell I'm not religious about it

scosman:
This looks great. Few quick comments, but I'll be able to look in more depth next week!

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
leonardmq: I added validation on the server side, but doing it on the client too would be nice as it allows validating in realtime and doing things like disabling the submit button while invalid. 

The current UI for editing the repair requires that you first save to apply the edit to the unsaved repair, and then you submit the repair. If we do server-only, the validation feedback would not be on this same form. 

I could change it to make the edit form directly submit to the API (instead of applying to the run without submitting) if you think that makes more sense - then server validation errors could come through on this same form

scosman:
@leonardmq is this ready for me to take a look?

leonardmq:
> @leonardmq is this ready for me to take a look?

Yes should be ready - some open questions around desired behavior of the UI, so if changes are wanted there, I'll tweak it after the review

app/desktop/studio_server/repair_api.py
scosman: Good catch.

Best thing is to create a validator in the task_run model. If there's a repair_run.output and task.output_json_schema, validate. Very similar to `validate_output_format` in the TaskRun class.

That will catch it here, and anywhere else.

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
scosman: Yeah - I like the latter. Click once to save, see any errors immediately. No need for the intermediate state.

client-validation - could be nice, but P3. We'll do it later 😀. Let's remove the TODO in code for now? I try to reserve those for bugs vs features.

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
scosman: oh that's annoying, but yeah. Makes sense.

Maybe set created_by: "unknown" just in case.

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
scosman: just use the warning component? P2

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+            d="M128,20.00012a108,108,0,1,0,108,108A108.12217,108.12217,0,0,0,128,20.00012Zm0,192a84,84,0,1,1,84-84A84.0953,84.0953,0,0,1,128,212.00012Zm-12-80v-52a12,12,0,1,1,24,0v52a12,12,0,1,1-24,0Zm28,40a16,16,0,1,1-16-16A16.018,16.018,0,0,1,144,172.00012Z"
+          />
+        </svg>
+        Structure does not match required schema
+      </div>
+    {/if}
+  {/if}
+
+  <FormContainer
+    submit_label="Apply"
```
scosman: "Save Repair (5 stars)" to match the auto repair button (per above comment, we won't separate "save" and submit")

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+      id={"repair_manual_output"}
+      label="Manual Repair"
+      inputType="textarea"
+      tall={true}
+      bind:value={repair_output_edited}
+    />
+
+    <button
+      class="link text-gray-500 text-sm text-right"
+      on:click={handle_cancel}>Cancel edit</button
```
scosman: Just "Cancel"

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
```diff
@@ -0,0 +1,116 @@
...
+  {/if}
+
+  <FormContainer
+    submit_label="Apply"
+    on:submit={handle_submit}
+    disabled={!repair_output_edited_valid}
+  >
+    <FormElement
+      id={"repair_manual_output"}
+      label="Manual Repair"
```
scosman, line 104: info_text="Provide a improvement or correction to the task's output"

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -425,15 +445,24 @@
                 bind:value={repair_instructions}
               />
             </FormContainer>
-          {:else if repair_review_available}
+          {:else if repair_review_available || repair_edit_mode}
```
scosman: Let's just make these two separate blocks? They are quite different. While we're using the repair to seed the manual, the user may completely change it. Text like "The model has attempted to fix the output given" becomes out of date.

We can add a paragraph with explanation text in new block, something like "Manually improve or correct the response."

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -459,24 +488,29 @@
...
+                  {:else}
+                    Accept Repair (5 Stars)
+                  {/if}
+                </button>
+              </div>
+            </div>
+
             {#if accept_repair_error}
-              <p class="text-error font-medium text-sm">
+              <p class="mt-2 text-error font-medium text-sm">
```
scosman, line 519: this is my fault, but error should always be above the buttons. If not, they can appear off screen depending on where the user scrolls to, and make the UI feel like nothing happened.

scosman (changes requested):
Looks good, some small comments. Like your UX suggestion for the save flow.

Issue where if auto-repair fails, you can’t manual repair: I’m okay with this for now. If the selected model fails repeatedly, you have bigger problems. This problem is not being introduced here, repairing would already be failing: Let’s see if this is a common complaint before fixing.

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
leonardmq: ✅  - renamed

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
leonardmq: ✅  - renamed

app/web_ui/src/routes/(app)/run/output_repair_edit_form.svelte
leonardmq: ✅ - edited the text

app/web_ui/src/routes/(app)/run/run.svelte
leonardmq: ✅ - split those two into separate blocks and used `Manually improve or correct the response.` as text at the top

app/web_ui/src/routes/(app)/run/run.svelte
leonardmq: I have not moved this error block because not sure exactly where you would want them to be. But I have moved the errors above the buttons in my manual repair form component

leonardmq:
@scosman 

Pushed changes to align with your review comments:
- Refactored to submit directly from the repair edit form (removing intermediate state)
- Removed all client-side validation—including the `Malformed JSON` warning and the disabled state on the submit button—to keep validation fully server-side for simplicity and consistency
- Updated `validate_repaired_output` in `TaskRun` to call `validate_schema` on the provided repair output (if any), when the task has an output JSON schema; added a couple of tests for this
- Removed TODOs and tweaked some of the UI text

The error message from the validator might still be a bit off, since it comes from `validate_schema`, which currently has a narrow message format. I opened a separate PR to address that: https://github.com/Kiln-AI/Kiln/pull/272 - since fixing it required touching unrelated parts of the codebase.

scosman:
👍👍👍

scosman (approved):
👍

PR #272, opened 2025-04-06
refactor: schema validation error handling

leonardmq:
## What does this PR do?

Refactors `validate_schema` to improve modularity and error handling. Previously, all JSON Schema validation errors were wrapped into a `ValueError` with overly specific messaging, which limited reusability. 

This also led to minor bugs such as:
- `validate_output_format` attempted to catch a JSON Schema validation error that was never actually raised.
- errors related to structured input validation returned messages incorrectly referring to output.

## Related Issues

N/A

## Checklists


libs/core/kiln_ai/datamodel/json_schema.py
```diff
@@ -41,16 +41,33 @@ def validate_schema(instance: Dict, schema_str: str) -> None:
...
     Raises:
         jsonschema.exceptions.ValidationError: If validation fails
-        ValueError: If the schema is invalid
+    """
+    schema = schema_from_json_str(schema_str)
+    v = jsonschema.Draft202012Validator(schema)
+    v.validate(instance)
+
+
+def validate_schema_with_value_error(
```
leonardmq, line 50: Possibly could simply merge this one with `validate_schema` (would avoid having this variant with a very long name) if we don't see a need for the separation

libs/core/kiln_ai/datamodel/task_run.py
```diff
@@ -87,14 +87,19 @@ def validate_input_format(self, info: ValidationInfo) -> Self:
...
-        # validate output
+        # validate input
         if task.input_json_schema is not None:
             try:
-                validate_schema(json.loads(self.input), task.input_json_schema)
+                input_parsed = json.loads(self.input)
             except json.JSONDecodeError:
                 raise ValueError("Input is not a valid JSON object")
-            except jsonschema.exceptions.ValidationError as e:
-                raise ValueError(f"Input does not match task input schema: {e}")
```
leonardmq, line 98: This one was never reached since `jsonschema.exceptions.ValidationError` was never raised by `validate_schema`

Issue #142, opened 2025-01-31
[ux/feature-request] auto populate all subtopics/topics


lucyknada:
as can be seen in the demo video starting at 37 seconds: https://vimeo.com/1044098339 where it had to be fast-forwarded to not show real-time how long that takes to do manually each time.

you have to first generate the topics, then for each one click subtopics, then for each subtopic click create data, ...

this becomes very cumbersome with tens of topics and tens of subtopics per topic to get a baseline before you generate more selectively, ideally there would simply be a single modal where you specify N topics, N subtopics, N data per subtopic and it does it, then you can still add more subtopics/topics/data manually by clicking the usual links on the right side

scosman:
This has been done for a while, finally updating the card!

PR #277, opened 2025-04-12
feat: synthetic data generation cascade down topic tree

leonardmq:
## What does this PR do?

- Add `Add to all` buttons on the `Synthetic data` page to let the user generate samples for every subtopic leaf descendant of a given topic node; I called that "Cascade mode"
- Add optional parallelization of synthetic data generation when in Cascade mode

The parallelization is inspired by what was done for the `Save all` action; the error handling is similar. Tried to keep the workers' errors mapped to their topic so that if we want, we can later add something to `Try again` and retry only the failed topics.

## Related Issues

https://github.com/Kiln-AI/Kiln/issues/142

## Checklists

### DIFF
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
@@ -5,6 +5,7 @@
   import { KilnError } from "../../../../../lib/utils/error_handlers"
   import { client } from "$lib/api_client"
   import { createKilnError } from "$lib/utils/error_handlers"
+  import FormElement from "../../../../../lib/utils/form_element.svelte"
 
   interface TopicNodeWithPath {
     path: string[]
@@ -21,6 +22,8 @@
   export let num_samples_to_generate: number = 8
   export let custom_topics_string: string | null = null
   export let cascade_mode: boolean = false
+  export let generate_samples_mode: "parallel" | "sequential" = "parallel"
+  let ui_show_errors = false
 
   export let on_completed: () => void
 
@@ -52,9 +55,16 @@
     }
   }
 
+  type GenerateSampleResponse = {
+    error: KilnError | null
+  }
+
   let sample_generating: boolean = false
   let sample_generation_error: KilnError | null = null
-  async function generate_samples(topic: TopicNodeWithPath) {
+  let generate_samples_sub_errors: KilnError[] = []
+  async function generate_samples(
+    topic: TopicNodeWithPath,
+  ): Promise<GenerateSampleResponse> {
     try {
       if (!model) {
         throw new KilnError("No model selected.", null)
@@ -102,17 +112,19 @@
         model_provider,
       )
     } catch (e) {
-      // TODO: handle errors better and distinguish between sub-errors and
-      // orchestration errors
       if (e instanceof Error && e.message.includes("Load failed")) {
         sample_generation_error = new KilnError(
           "Could not generate samples, unknown error. If it persists, try another model.",
           null,
         )
-      } else {
-        sample_generation_error = createKilnError(e)
+
+        return { error: sample_generation_error }
       }
+
+      return { error: createKilnError(e) }
     }
+
+    return { error: null }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte
```diff
@@ -174,122 +175,17 @@
...
     // @ts-expect-error dialog is not a standard element
     modal?.showModal()
   }
 
-  function add_synthetic_samples(
-    samples: unknown[],
-    model_name: string,
-    model_provider: string,
-  ) {
-    // Add ignoring dupes and empty strings
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte:193]: Was there a plan to skip the duplicates?

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte
```diff
@@ -482,50 +397,19 @@
...
+    {id}
+    {data}
+    {path}
+    {project_id}
+    {task_id}
+    {human_guidance}
+    {model}
+    {num_samples_to_generate}
+    {custom_topics_string}
+    on_completed={handleGenerateSamplesCompleted}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte:410]: Wondering if there’s a more idiomatic way in Svelte to control modal visibility (when the modal decides to close itself) than passing in a callback from the parent component like this

leonardmq:
The `Add data to all` button is added on the root of the tree as well as on each topic (that has 1 or more child topics):
image

The `Add data to all` button is hidden on topics that have no children:
image

The `Add data to all` modal is the same as `Add data` with the text slightly altered to align with what it does, and adds a parallelization field:
image

If the generation is successful, it will close the modal.

If the generation runs into errors, the errors from each worker are collected and displayed at the end (specific error message a little clunky - might want to rephrase):

image

In the case of a non-cascade (i.e. single topic) generation failure, the error is shown directly like this:
image

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
```diff
@@ -0,0 +1,339 @@
...
+<dialog id={`${id}-generate-samples`} class="modal">
+  <div class="modal-box">
+    <form method="dialog">
+      <button
+        class="btn btn-sm text-xl btn-circle btn-ghost absolute right-2 top-2 focus:outline-none"
+        >✕</button
+      >
+    </form>
+    <h3 class="text-lg font-bold">
+      Generate Samples{cascade_mode ? " (cascade)" : ""}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte:253]: Just "Generate Data" - not Samples, no need for (cascade)

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
```diff
@@ -0,0 +1,339 @@
...
+          {/if}
+        {/if}
+
+        <button
+          class="btn mt-6 {custom_topics_string ? '' : 'btn-primary'}"
+          on:click={generate_samples}
+        >
+          Generate {num_samples_to_generate} Samples
+          {#if cascade_mode}
+            (for each subtopic)
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte:330]: "For Each Topic" (no brackets, caps)

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
```diff
@@ -0,0 +1,339 @@
...
+              Failed to generate samples for {topics_failed_to_generate[0].topic.path.join(
+                " → ",
+              )}. Running again may resolve transient issues.
+              <div>
+                {topics_failed_to_generate[0].error?.getMessage()}
+              </div>
+            </div>
+          {:else}
+            <div class="text-error font-light text-sm mt-4">
+              {topics_succeeded_to_generate_count} topics generated successfully
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte:300]: "N topics failed. Running again may resolve transient issues."

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte:193]: I guess 😆 🤷 . Okay to delete comment.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generated_data_node.svelte:410]: The modal can hide itself without removing itself from the DOM (.close). The callback is still needed since we have data to work with.

scosman (approved):
Some small string changes to make, but I'll throw those in after merging.

Issue #298, opened 2025-05-02
[Bug] 1Password overlays buttons on hidden modal fields


leonardmq:
**Describe the bug**

Screenshots further down below.

UI bug:
- 1password Chrome extension overlays its buttons on form fields contained in modals that are present in the DOM but not visible

On the `connect_providers.svelte` page, for example, it seems 1Password matches on a text input inside the hidden `openai_compatible_dialog` modal.

A workaround is to use the 1Password opt-out attribute ([docs](https://releases.1password.com/b5x/beta/8.10.52-14/)):
> When designing a website to work with 1Password, you can now use “data-op-ignore” as an alias for the “data-1p-ignore” attribute.

Although, it would probably be cleaner to remove those modals from the DOM when not shown.

**Checks**

- [x] I've read the [troubleshooting guide](https://docs.getkiln.ai/docs/troubleshooting-and-logs)
- [x] I've tried to reproduce the problem using another model, and confirmed it's not an issue specific to the model I've chosen.
- [x] I've searched [the docs](https://docs.getkiln.ai) for a solution
- [x] I've searched for existing Github issues/discussions

**To Reproduce**

If you have 1password (or some other password manager that may behave similarly):
1. Go to http://localhost:5173/settings/providers or on a detailed eval page (http://localhost:5173/evals/A/B/C)
2. See 1password button overlay floating in the middle of the page

If you click in the page, the button disappears; it comes back again if you hard refresh the page, or if you click on another page in the sidebar, and come back.

**Expected behavior**

Password managers do not overlay their buttons on top of fields in closed modals.

**Screenshots**

Connect providers screen:
<img width="912" alt="Image" src="https://github.com/user-attachments/assets/90aba282-6e48-4e9e-9a1e-7369cacc7a92" />

Evals detailed UI:
<img width="1037" alt="Image" src="https://github.com/user-attachments/assets/041702c1-9eb4-4e86-9c49-551c2a070aed" />

**Error Logs**

N/A

**System Information:**

- OS: MacOS
- Browser: Chrome
- Kiln app Version: `0.15.0` (was happening on all previous versions from the past months too)

**Additional context**

N/A

PR #299, opened 2025-05-02
fix: disable 1password on text input field

leonardmq:
## What does this PR do?

Workaround:
- add `data-op-ignore="true"` field on `FormElement` text input to prevent 1password from overlaying its buttons on top of form fields contained in modals that are present in the DOM but not visible.

A cleaner fix would be to remove the `dialog` from the DOM when not shown, but this workaround should be sufficient for now.

## Related Issues

Fixes: https://github.com/Kiln-AI/Kiln/issues/298

## Checklists


scosman:
Haha. I literally made the same fix on my branch yesterday!

Issue #21, opened 2024-11-06
Add RAG


scosman:
A issue to track requests and details relating to RAG.

scosman:
Closing as completed! @leonardmq

PR #304, opened 2025-05-09
RAG: base extractor and gemini

leonardmq:
## What does this PR do?

This PR adds a `BaseExtractor` abstract class and the `GeminiExtractor` that implements it.

## Related Issues

Part of the overarching RAG issue: https://github.com/Kiln-AI/Kiln/issues/21

## Checklists

## Summary by CodeRabbit

- **New Features**
  - Introduced a flexible framework for file content extraction, supporting plain text and markdown formats.
  - Added a Gemini-powered extractor for documents, images, audio, and video, with configurable prompts and robust error handling.

- **Bug Fixes**
  - Improved error handling for file reading and MIME type detection.

- **Tests**
  - Added comprehensive test suites for extractors and file utility functions to ensure correct behavior and error handling.

- **Chores**
  - Updated dependencies to include support for Google Gemini API integration.


leonardmq:
@coderabbitai review

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,135 @@
+from enum import Enum
+
+from google import genai
+from google.genai import types
+from pydantic import Field
```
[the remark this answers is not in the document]
leonardmq: Coderabbit's comment seems to be referring to an old version of Google's Gen AI SDK that has since been deprecated: https://github.com/google-gemini/deprecated-generative-ai-python/blob/main/docs/api/google/generativeai.md

The import paths I had here seem correct

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+
+class GeminiExtractor(BaseExtractor):
+    def __init__(self, gemini_client: genai.Client, config: GeminiExtractorConfig):
+        super().__init__(config)
+        self.gemini_client = gemini_client
+
+        # hack to access the concrete config subclass here without failing type checking
+        self.config = config
```
leonardmq, line 97: The `config` should normally be accessible to this instance, but the typechecks fail. Not sure if I am doing anything wrong, or if I should just split up the config between `base` and concrete explicitly instead of having the concrete config extend the `base` config and use it for both?

Shadowing the `config` in this instance like I did here works ok, but seems clunky

libs/core/kiln_ai/tests/data/lenna.png
scosman, line 1: We should use a different image. https://en.wikipedia.org/wiki/Lenna#Criticism

I usually use these: https://github.com/MohamedBakrAli/Kodak-Lossless-True-Color-Image-Suite

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ExtractionFormat(str, Enum):
+    TEXT = "text/plain"
+    MARKDOWN = "text/markdown"
+
+
+class BaseExtractorConfig(BaseModel):
```
scosman, line 17: Thinking ahead to serialization this might cause some issues. Our existing configs (EvalConfig) refer to a serializable Kiln model (not just a in memory model like this). 

1) This data class is fine as a concept, but naming wise it sounds like EvalConfig, which is actually a serializeable Kiln model. Maybe just "ExtractorConfigInstance" to differentiate? We're going to need a "ExtractorConfig" Kiln model as well for serialization. 

2) When it comes time to make the serializable version (ExtractorConfig -> KilnBaseModel), be sure to follow the pattern from EvalConfig. use a "type" + "properties dict" pattern instead. See EvalConfig properties and config_type for example. Since they will be persisted and reloaded (potentially by another version of Kiln) the models shouldn't have changing/different properties or deserialization will fail and cause unknown bugs.

3) You might want to simplify and only use the KilnBaseModel based ExtractorConfig instead of also having EvalConfigInstance (this class). It's less OOP, but that's fine. You can write validators for each type (Gemini), but you'd have to do that anyways. You can add accessors for typing accessing properties. It will probably be easier to test if you just use the one model everywhere.

Rough idea for a single ExtractorConfig base model w serialization:

 - Make a single ExtractorConfig in the /datamodel folder, inheriting from KilnBaseModel (later we will parent it, but don't need to that yet)
 - Properties needed by all configs are independent properties on the model (eg output format)
 - It has an "extractor_type" string (Gemini, etc)
 - Properties that are specific to a the extractor_type are in a properties dict 
 - Write a validator for each extractor_type to make sure necessary properties are there. Ignore types you don't recognize from validation perspective (backwards compat)
 - Write accessors for properties dict, for typing
 - Remove the Gemini extractor, and just have the init check the ExtractorConfig.type == "gemini"

This can be a refactor we do after if you like.

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    Base class for all extractor configs.
+    """
+
+    passthrough_mimetypes: list[str] = Field(
```
scosman: `list[ExtractionFormat]`?

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+        description="If the mimetype is in this list, the extractor will not be used and the text content of the file will be returned as is. Only text mime types are supported for passthrough.",
+    )
+
+    output_format: ExtractionFormat = Field(
+        default=ExtractionFormat.MARKDOWN,
+        description="The format to use for the output.",
+    )
+
+    @model_validator(mode="after")
+    def validate_passthrough_mime_types(self) -> Self:
```
scosman: If we use the type above, and remove this. Stricter type checking is good (fixed list, not all `text/`)

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+class ExtractionOutput(BaseModel):
+    is_passthrough: bool = Field(
+        default=False, description="Whether the extractor returned the file as is."
+    )
+    content_format: ExtractionFormat = Field(
+        description="The format of the extracted data."
+    )
+    content: str = Field(description="The extracted data.")
```
scosman, line 50: I was thinking we'd store this in a separate file. A big doc as a JSON string would work but would be painful for PRs.

Can be a later refinement, not blocking this PR.

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Should be subclassed, and the _extract method implemented.
+    """
+
+    def __init__(self, config: BaseExtractorConfig):
+        self.config = config
+
+    @abstractmethod
+    def _extract(
+        self, file_info: FileInfoInternal, custom_prompt: str | None
```
scosman: Don't take a prompt here. Put it in the config. That way it's saved so the config is consistent always (even if we change the default prompt, old configs keep their exact behaviour).

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,183 @@
...
+        This method must be implemented by subclasses to define how content is extracted.
+        The optional custom prompt is provided by the user if they want to override the
+        default prompt for this extractor.
+        """
+        pass
+
+    def extract(
+        self,
+        file_info: FileInfo,
+        custom_prompt: str | None = None,
```
scosman: ditto on prompt.

scosman (changes requested):
WIP review

libs/core/kiln_ai/tests/data/lenna.png
leonardmq: ✅  - Replaced the image with the parrots from the linked dataset

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅ - removed the validator

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅ - moved it to the config passed via constructor (not sure yet whether it belongs more on the subclass or base config; put it in subclass config for now)

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅ - moved it to the config passed via constructor (not sure yet whether it belongs more on the subclass or base config; put it in subclass config for now)

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: I'll remove the `BaseExtractorConfig` (and so probably also `GeminiExtractorConfig`) and replace with `ExtractorConfig` similar to the `EvalConfig` in `/datamodel/eval.py`; seems like there would not be much of a point in keeping this one (and needing to do all the extra work of mapping from the `ExtractorConfig` to this) once we have `ExtractorConfig`

leonardmq:
@coderabbitai review

libs/core/kiln_ai/adapters/extractors/test_gemini_extractor.py
```diff
@@ -0,0 +1,372 @@
...
+    # mock the bytes loading
+    mock_load_file_bytes = MagicMock()
+    mock_load_file_bytes.return_value = b"test content"
+    mock_gemini_extractor_with_kind_prompts._load_file_bytes = mock_load_file_bytes
+
+    # test the extract method
+    with pytest.raises(Exception):
+        mock_gemini_extractor_with_kind_prompts.extract(
+            FileInfoInternal(path="test.pdf", mime_type="application/pdf"),
+        )
```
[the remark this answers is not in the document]
leonardmq: @scosman - I was wondering if it would be worth creating our own error classes for specific flows / services (e.g. `ExtractorError` like suggested here) or if raising `ValueError` anywhere is good enough for the time being

libs/core/kiln_ai/adapters/extractors/test_gemini_extractor.py
```diff
@@ -0,0 +1,372 @@
...
+    )
+
+    # check the gemini client was called with the correct arguments
+    mock_gemini_client.models.generate_content.assert_called_once_with(
+        model="fake-model",
+        contents=[
+            types.Part.from_bytes(data=b"test content", mime_type="application/pdf"),
+            PROMPTS_FOR_KIND[Kind.DOCUMENT],
+        ],
+    )
```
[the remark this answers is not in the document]
leonardmq: The `assert` works here because `types.Part.from_bytes()` returns a `Part` object that inherits from `pydantic.BaseModel`, which implements `__eq__` to do value-based (field-by-field) comparison. So as long as the field values are equal, even distinct instances will compare equal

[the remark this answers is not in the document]
leonardmq: See `types.Part` definition here: 
- https://github.com/googleapis/python-genai/blob/b9c9e32efaefdab54414a2bc1285b1014e7385fd/google/genai/types.py#L616

```py
class Part(_common.BaseModel):
  ...
```

See `_common.BaseModel` definition here:
- https://github.com/googleapis/python-genai/blob/b9c9e32efaefdab54414a2bc1285b1014e7385fd/google/genai/_common.py#L199

```py
class BaseModel(pydantic.BaseModel):
  ...
```

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+    ExtractionFormat,
+    ExtractionOutput,
+    FileInfoInternal,
+)
+
+
+class Kind(Enum):
+    DOCUMENT = "document"
+
+    # NOTE: maybe remove the these ones for now as we don't use them and the mimetypes list for them
```
scosman: Looks like making the list took some work. Fine to keep it.

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+        Determines if the file should be returned without extraction based on its MIME type.
+
+        Returns:
+            True if the MIME type is in the passthrough list; otherwise, False.
+        """
+        return mime_type.lower() in {
+            mt.lower() for mt in self.config.passthrough_mimetypes
+        }
+
+    def _load_file_bytes(self, path: str) -> bytes:
```
scosman: Necessary? We already have load_file_bytes. This is just transforming the exception into another exception? The path will be added via `Error extracting {file_info.path}` above

Also: let's limit docstrings on private methods (not SDK methods), and reserve them for things that need explanation? Code will get long with these. I'd likely forget to update them as well.

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Raises:
+            ValueError: If the file cannot be read as text.
+        """
+        try:
+            return file_utils.load_file_text(path)
+        except Exception as e:
+            raise ValueError(f"Error loading file text for {path}: {e}") from e
+
+    def _get_mime_type(self, path: str) -> str:
```
scosman: ditto

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
scosman: We should eventually move this enum into the model once we have a model for docs. But can come later

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+
+class GeminiExtractorConfig(BaseExtractorConfig):
+    default_prompt: str = Field(
+        description="The default prompt to use for the extractor."
+    )
+    custom_prompt: str | None = Field(
+        default=None,
+        description="A custom prompt provided by the user to use for the extractor. It takes precedence over the prompt_for_kind and default_prompt.",
+    )
+    prompt_for_kind: dict[Kind, str] = Field(
```
scosman, line 76: Let's collapse all 3, just keeping "prompt_for_kind"? Will simplify downstream usage. 

We don't need/want default/custom logic at the config level. The fact that there's a default will be handled in config init, or even just the UI. But the caller always just uses the saved prompt.

We should add a validator which checks all values are set in prompt_for_kind (doc, vid, etc), but the initializer can use defaults for any not set manually, which will keep usage easy.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+        default_factory=dict,
+        description="A dictionary of prompts for each kind of file.",
+    )
+    model: str = Field(description="The model to use for the extractor.")
```
scosman: Nit: model_id and model_provider are the usual keys.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
scosman: You've got the same variable name being used for 2 different types. Base class setting config:BaseExtractorConfig, this class setting config:GeminiExtractorConfig.

This should resolve itself when you move to single Config class, but lets make sure it does and remove this.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+
+class GeminiExtractor(BaseExtractor):
+    def __init__(self, gemini_client: genai.Client, config: GeminiExtractorConfig):
+        super().__init__(config)
+        self.gemini_client = gemini_client
+
+        # hack to access the concrete config subclass here without failing type checking
+        self.config = config
+
+    def _get_kind_from_mime_type(self, mime_type: str) -> Kind:
```
scosman: ditto on internal function docstrings.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+        Returns:
+            The corresponding Kind enum value for the given MIME type.
+
+        Raises:
+            ValueError: If the MIME type is not supported.
+        """
+        for kind, mime_types in MIME_TYPES_SUPPORTED.items():
+            if mime_type in mime_types:
+                return kind
+        raise ValueError(f"Unsupported MIME type: {mime_type}")
```
scosman: Okay, old man programmer thinking: avoid exceptions in the getter, throw them from the caller where we have more context.

A get method returning "None" is a fine response, not an exception. 

One level up, a None Kind becomes an exception. The gemini extractor doesn't support it. By

```
if not Kiln:
  raise ValueError(f"The Gemini document extractor doesn't support: {mime_type}")
```

Leads to more helpful exception messages, and more reuse (_get_kind_from_mime_type can be used in cases where None isn't an exception)

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _get_prompt_for_kind(self, kind: Kind) -> str:
```
scosman: wrapper method not needed for a 1 line call.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        Returns the prompt string for the specified file kind.
+
+        If a prompt is configured for the given kind, it is returned; otherwise, the extractor's default prompt is used.
+        """
+        return self.config.prompt_for_kind.get(kind, self.config.default_prompt)
```
scosman: per other comment, the extractor shouldn't be considering "default". Keep that in the Config init or UI.

libs/core/kiln_ai/adapters/extractors/file_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+"""
+Utility functions for file operations used by the extractors framework.
+
+This module provides common functionality for loading files as bytes or text,
+and for determining MIME types.
+"""
+
+
+def load_file_bytes(path: str) -> bytes:
```
scosman: Why have a wrapper here? It's a 1 line system call. Wrapper adds complexity.

libs/core/kiln_ai/adapters/extractors/file_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    Args:
+        path: Path to the file to be read.
+
+    Returns:
+        The contents of the file as a bytes object.
+    """
+    return pathlib.Path(path).read_bytes()
+
+
+def load_file_text(path: str) -> str:
```
scosman: ditto

libs/core/kiln_ai/adapters/extractors/file_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    Args:
+        path: The path to the file to be read.
+
+    Returns:
+        The contents of the file as a string.
+    """
+    return pathlib.Path(path).read_text(encoding="utf-8")
+
+
+def get_mime_type(path: str) -> str:
```
scosman: See my "old man programmer" comment below, about raising exceptions at the level up.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -0,0 +1,152 @@
...
+                types.Part.from_bytes(
+                    data=self._load_file_bytes(file_info.path),
+                    mime_type=file_info.mime_type,
+                ),
+                prompt,
+            ],
+        )
+
+        if response.text is None:
+            raise ValueError("No text returned from Gemini")
```
scosman: Add a bit more detail in the error message. "No text returned from Gemini when extracting document"

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: The idea with these was to make it easier to target and mock in tests, and not tie them to the current implementation (as I suspect some of it might get pulled into global utils, or we might add HTTP fetch for example), but maybe too boilerplate-y so can remove

libs/core/kiln_ai/adapters/extractors/file_utils.py
leonardmq: ✅  - removed (also we could possibly use the real test files if we want)

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅  - updated string

libs/core/kiln_ai/adapters/extractors/file_utils.py
leonardmq: ✅  - updated to return `None`. It could still throw if the `path` is `None`, but typing and upstream validation should prevent that

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅ - removed `default_prompt`, `custom_prompt` and kept only `prompt_for_kind`

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅ - removed as a result of `default_prompt` going away

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅  - made it return `None` instead

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅  - removed (almost) all the docstrings

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅  - renamed into `model_id`

When we have the `ExtractorConfig`, I will add `model_provider` to it (e.g. `gemini`)

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: ✅  - removed `default_prompt` and `custom_prompt`

> But the caller always just uses the saved prompt.

What does the "saved prompt" refer to in this case? A user-supplied prompt for each `kind` that they would have saved?

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅ - removed

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: ✅ - removed

libs/core/kiln_ai/adapters/extractors/file_utils.py
leonardmq: ✅ - removed

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
scosman: The prompt always comes from the config, so it's always "saved" (since configs are usually persisted prompt_for_kind is "saved", is what I was trying to say). The adapter just uses prompt_for_kind. If there's default/custom logic, that's when then ExtratorConfig is created - but doesn't need to be saved beyond that and the adapter doesn't need to be aware of it.

scosman (approved):

PR #313, opened 2025-05-15
refactor: ExtractionConfig datamodel

leonardmq:
## What does this PR do?

- add datamodel for `ExtractorConfig` (inheriting from `KilnBaseModel`)
- refactor `BaseExtractor` and `GeminiExtractor` to use `ExtractorConfig`
- update tests

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/extraction/__init__.py
+++ b/libs/core/kiln_ai/adapters/extraction/__init__.py
@@ -5,9 +5,10 @@ This package provides a framework for extracting content from files
 using different extraction methods.
 """
 
-from . import base_extractor, gemini_extractor
+from . import base_extractor, gemini_extractor, registry
 
 __all__ = [
     "base_extractor",
     "gemini_extractor",
+    "registry",
 ]
--- a/libs/core/kiln_ai/adapters/extraction/gemini_extractor.py
+++ b/libs/core/kiln_ai/adapters/extraction/gemini_extractor.py
@@ -1,25 +1,14 @@
 import pathlib
-from enum import Enum
 
 from google import genai
 from google.genai import types
-from pydantic import Field
 
-from kiln_ai.adapters.extractors.base_extractor import (
+from kiln_ai.adapters.extraction.base_extractor import (
     BaseExtractor,
-    BaseExtractorConfig,
-    ExtractionFormat,
     ExtractionOutput,
     FileInfoInternal,
 )
-
-
-class Kind(Enum):
-    DOCUMENT = "document"
-    IMAGE = "image"
-    VIDEO = "video"
-    AUDIO = "audio"
-
+from kiln_ai.datamodel.extraction import ExtractorConfig, ExtractorType, Kind
 
 # docs list out supported formats:
 # - https://ai.google.dev/gemini-api/docs/document-processing#supported-formats
@@ -72,29 +61,16 @@ MIME_TYPES_SUPPORTED = {
 }
 
 
-class GeminiExtractorConfig(BaseExtractorConfig):
-    prompt_for_kind: dict[Kind, str] = Field(
-        default_factory=dict,
-        description="A dictionary of prompts for each kind of file.",
-    )
-    model_id: str = Field(
-        description="The model to use for the extractor.",
-        examples=["gemini-2.0-flash"],
-    )
-    output_format: ExtractionFormat = Field(
-        default=ExtractionFormat.MARKDOWN,
-        description="The format to use for the output.",
-    )
-
-
 class GeminiExtractor(BaseExtractor):
-    def __init__(self, gemini_client: genai.Client, config: GeminiExtractorConfig):
-        super().__init__(config)
-        self.gemini_client = gemini_client
+    def __init__(self, gemini_client: genai.Client, extractor_config: ExtractorConfig):
+        if extractor_config.extractor_type != ExtractorType.gemini:
+            raise ValueError(
+                f"GeminiExtractor must be initialized with a gemini extractor_type config. Got {extractor_config.extractor_type}"
+            )
 
-        # TODO: the hack below will go away once we have ExtractorConfig as a dict anyway
-        # hack to access the concrete config subclass here without failing type checking
-        self.config = config
+        super().__init__(extractor_config)
+        self.gemini_client = gemini_client
+        self.gemini_config = extractor_config.gemini_properties()
 
     def _get_kind_from_mime_type(self, mime_type: str) -> Kind | None:
         for kind, mime_types in MIME_TYPES_SUPPORTED.items():
@@ -107,12 +83,12 @@ class GeminiExtractor(BaseExtractor):
         if kind is None:
             raise ValueError(f"Unsupported MIME type: {file_info.mime_type}")
 
-        prompt = self.config.prompt_for_kind.get(kind)
+        prompt = self.gemini_config.prompt_for_kind.get(kind)
         if prompt is None:
             raise ValueError(f"No prompt found for kind: {kind}")
 
         response = self.gemini_client.models.generate_content(
-            model=self.config.model_id,
+            model=self.gemini_config.model_name,
             contents=[
                 types.Part.from_bytes(
                     data=pathlib.Path(file_info.path).read_bytes(),
@@ -128,5 +104,5 @@ class GeminiExtractor(BaseExtractor):
         return ExtractionOutput(
             is_passthrough=False,
             content=response.text,
-            content_format=self.config.output_format,
+            content_format=self.extractor_config.output_format,
         )
--- a/libs/core/kiln_ai/datamodel/extraction.py
+++ b/libs/core/kiln_ai/datamodel/extraction.py
@@ -1,0 +1,88 @@
+import json
+from enum import Enum
+from typing import Any, List
+
+from pydantic import BaseModel, Field, ValidationError, model_validator
+from typing_extensions import Self
+
+from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnBaseModel
+
+
+def format_properties_errors(e: ValidationError) -> str:
+    errors: List[str] = []
+    for error in e.errors():
+        loc = error["loc"][0]
+        msg = error["msg"]
+        errors.append(f"{loc}: {msg}.")
+    return "\n".join(errors)
+
+
+class OutputFormat(str, Enum):
+    TEXT = "text/plain"
+    MARKDOWN = "text/markdown"
+
+
+class ExtractorType(str, Enum):
+    gemini = "gemini"
+
+
+class Kind(str, Enum):
+    DOCUMENT = "document"
+    IMAGE = "image"
+    VIDEO = "video"
+    AUDIO = "audio"
+
+
+class GeminiProperties(BaseModel):
+    prompt_for_kind: dict[Kind, str] = Field(
+        description="A dictionary of prompts for each kind of content to extract.",
+    )
+
+    model_name: str = Field(
+        description="The name of the model to use for this extractor config. ",
+    )
+
+
+class ExtractorConfig(KilnBaseModel):
+    name: str = NAME_FIELD
+
+    output_format: OutputFormat = Field(
+        default=OutputFormat.MARKDOWN,
+        description="The format to use for the output.",
+    )
+    passthrough_mimetypes: list[OutputFormat] = Field(
+        default_factory=list,
+        description="If the mimetype is in this list, the extractor will not be used and the text content of the file will be returned as is.",
+    )
+    extractor_type: ExtractorType = Field(
+        default=ExtractorType.gemini,
+        description="This is used to determine the type of extractor to use.",
+    )
+    properties: dict[str, Any] = Field(
+        default={},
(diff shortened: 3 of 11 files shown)


leonardmq:
@coderabbitai review

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,99 @@
...
+from enum import Enum
+from typing import Any, List
+
+from pydantic import BaseModel, Field, ValidationError, model_validator
+from typing_extensions import Self
+
+from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnBaseModel
+
+
+def format_properties_errors(e: ValidationError) -> str:
```
leonardmq: Maybe can remove this altogether - this seems to cause some error messages to get double wrapped

libs/core/kiln_ai/adapters/extraction/gemini_extractor.py
```diff
@@ -107,12 +83,12 @@ def _extract(self, file_info: FileInfoInternal) -> ExtractionOutput:
         if kind is None:
             raise ValueError(f"Unsupported MIME type: {file_info.mime_type}")
 
-        prompt = self.config.prompt_for_kind.get(kind)
+        prompt = self.gemini_config.prompt_for_kind.get(kind)
         if prompt is None:
             raise ValueError(f"No prompt found for kind: {kind}")
```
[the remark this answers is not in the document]
leonardmq: The string values (e.g. `"document"`)  appear to be coerced into `Kind` during validation of the Pydantic model:
```py
class GeminiProperties(BaseModel):
    prompt_for_kind: dict[Kind, str] = Field(
        description="A dictionary of prompts for each kind of content to extract.",
    )

    model_name: str = Field(
        description="The name of the model to use for this extractor config. ",
    )

    def validate_model_name(self) -> Self:
        for kind in Kind:
            if kind not in self.prompt_for_kind:
                raise ValueError(f"Prompt for kind {kind.value} is required.")
        return self
```

`GeminiExtractor` accesses that model when it does: `self.gemini_config.prompt_for_kind.get(kind)`

Also added `test_extractor_config_kind_coercion` in `test_extraction_model.py` to double check.

leonardmq:
@coderabbitai review

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: Removed the Gemini properties pydantic validation - half-manual nested Pydantic validation in the validator was clunky and errors coming out were a little odd. Replaced with manual validation for now; which is not ideal, but more in line with how it is done in eval. 

Later, could possibly have a registry of models to use for `properties` validation, but since it might be re-used across different models / validators, and propagate errors to the frontend, should be better thought out than what I had here

libs/core/kiln_ai/adapters/extraction/__init__.py
```diff
@@ -5,9 +5,10 @@
 using different extraction methods.
 """
 
-from . import base_extractor, gemini_extractor
+from . import base_extractor, gemini_extractor, registry
```
scosman: Comment not related to this line, but can we rename the folder to "extractors". "Extraction" is the result of an "Extractor". This folder is "extractors"

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+
+def validate_model_name(model_name: Any):
+    if not isinstance(model_name, str):
+        raise ValueError("model_name must be a string.")
+    if model_name == "":
+        raise ValueError("model_name cannot be empty.")
+
+
+class ExtractorConfig(KilnBaseModel):
+    name: str = NAME_FIELD
```
scosman, line 59: Don't need to add it in this PR, but a description will probably be helpful.

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+    output_format: OutputFormat = Field(
+        default=OutputFormat.MARKDOWN,
+        description="The format to use for the output.",
+    )
+    passthrough_mimetypes: list[OutputFormat] = Field(
+        default_factory=list,
+        description="If the mimetype is in this list, the extractor will not be used and the text content of the file will be returned as is.",
+    )
+    extractor_type: ExtractorType = Field(
+        default=ExtractorType.gemini,
```
scosman: nit: I'd remove the default here. Anyone creating one should be forced to choose. If they forget an error is better than defaulting to something. Defaults are optional.

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+        description="This is used to determine the type of extractor to use.",
+    )
+    properties: dict[str, Any] = Field(
+        default={},
+        description="Properties to be used to execute the extractor config. This is config_type specific and should serialize to a json dict.",
```
scosman: config_type != extractor_tyle

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    passthrough_mimetypes: list[OutputFormat] = Field(
+        default_factory=list,
+        description="If the mimetype is in this list, the extractor will not be used and the text content of the file will be returned as is.",
+    )
+    extractor_type: ExtractorType = Field(
+        default=ExtractorType.gemini,
+        description="This is used to determine the type of extractor to use.",
+    )
+    properties: dict[str, Any] = Field(
```
scosman: Maybe replace any with something like `str | float | int | bool` to ensure it's json serializable.

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+            validate_prompt_for_kind(self.properties.get("prompt_for_kind"))
+            validate_model_name(self.properties.get("model_name"))
+            return self
+        else:
+            raise ValueError(f"Invalid extractor type: {self.extractor_type}")
+
+    @model_validator(mode="after")
+    def validate_json_serializable(self) -> Self:
+        try:
+            # This will raise a TypeError if the dict contains non-JSON-serializable objects
```
scosman: See above. I'd make it strictly typed, so we'll catch issues at typecheck time instead of runtime. Then can remove this for perf.

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def validate_prompt_for_kind(prompt_for_kind: Any):
+    # check prompt_for_kind is a dictionary
+    if not isinstance(prompt_for_kind, dict):
+        raise ValueError("prompt_for_kind must be a dictionary.")
+    # check all keys are valid kinds
+    for key, value in prompt_for_kind.items():
+        # type the key to a kind
+        Kind(key)
```
scosman: few lines here need work. Casting to nothing (might raise but won't be informational msg), commented out if.

scosman (changes requested):
Some small comments

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: ✅ - Removed the `json.dumps` validator I had copied over from `EvalConfig` and changed the field to this:
```py
properties: dict[str, str | int | float | bool | dict[str, str]] = Field(
        default={},
        description="Properties to be used to execute the extractor config. This is extractor_type specific and should serialize to a json dict.",
    )
```

Added `dict[str, str]` because we need it for `prompt_for_kind`, unless we want to flatten it out into `prompt_video`, `prompt_audio`, etc. at the root of the `properties`.

I also tried to make a more generalized and reusable `JSON_TYPE` that would support `dict` and `list` of other serializable primitives, but could not get it to work - something always seems to complain if the type is recursive (`ruff`, `python`, assignments, etc.).

An alternative I was going for initially was to have a Pydantic model (or maybe Pydantic's TypedDict) per config properties (e.g. `GeminiExtractorConfigProperties`), and have the `ExtractorConfig` model use the relevant one to validate the `properties` for the particular extractor. I tried that a bit last week but could not get it to work cleanly so reverted to manual to align with how we do it in `EvalConfig`.

How does that sound conceptually (not in this PR, but down the line; lower priority)?

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: ✅  - 🤦

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: ✅ - removed the default

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: ✅  - added an optional `description` field

libs/core/kiln_ai/adapters/extraction/__init__.py
leonardmq: ✅ - renamed `extraction/` folder to `extractors/`

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: ✅ - removed commented out `if` and rewrote error message

`Kind(key)` raised a `ValueError` that seemed reasonable: `'invalid_kind' is not a valid Kind` - but I have now added a `try / except` around to rewrite the message into: `Invalid kind in prompt_for_kind: 'invalid_kind'`

leonardmq:
@coderabbitai review

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,92 @@
...
+            raise ValueError(
+                f"Invalid prompt for kind: '{key}'. Prompt must be a string."
+            )
+
+    # check all kinds are present
+    for kind in Kind:
+        if kind not in prompt_for_kind:
+            raise ValueError(
+                f"Missing prompt for kind: '{kind.value}'. All kinds must be present in prompt_for_kind."
+            )
```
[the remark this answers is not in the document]
leonardmq:
```
# note it is SomeEnum(str, Enum) not SomeEnum(Enum)
class SomeEnum(str, Enum):
    A = "a"
    B = "b"
    C = "c"


def test_enum_validation():
    x = {
        "a": "text",
        "b": "text",
        "c": "text",
    }

    # this passes
    for string in x:
        assert string in SomeEnum

    # this passes
    for v in SomeEnum:
        assert v in x

    # this passes
    for v in SomeEnum:
        assert v.value in x
```

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -0,0 +1,92 @@
...
+    def validate_properties(self) -> Self:
+        if self.extractor_type == ExtractorType.gemini:
+            validate_prompt_for_kind(self.properties.get("prompt_for_kind"))
+            validate_model_name(self.properties.get("model_name"))
+            return self
+        else:
+            raise ValueError(f"Invalid extractor type: {self.extractor_type}")
+
+    def model_name(self) -> str | None:
+        return cast(str, self.properties.get("model_name"))
```
leonardmq, line 89: Could alternatively do proper runtime type checks:
```diff
     def model_name(self) -> str | None:
-        return cast(str, self.properties.get("model_name"))
+        value = self.properties.get("model_name")
+        if value is not None and not isinstance(value, str):
+            raise ValueError(f"model_name must be a string, got {type(value)}")
+        return value

     def prompt_for_kind(self) -> dict[Kind, str] | None:
-        return cast(dict[Kind, str], self.properties.get("prompt_for_kind"))
+        value = self.properties.get("prompt_for_kind")
+        if value is not None and not isinstance(value, dict):
+            raise ValueError(f"prompt_for_kind must be a dict, got {type(value)}")
+        # Convert string keys to Kind enum
+        if value is not None:
+            return {Kind(k): v for k, v in value.items()}
+        return None
```

libs/core/kiln_ai/datamodel/extraction.py
scosman: Ahhh, I see why I was lazy and did the dumps validator. Thanks for looking at this in more details.

I'd avoid the more-typed version as backwards compatibility is always a nightmare on properties fields. I'm also amazed Pydantic doesn't have this built in.

Aligning to EvalConfig is reasonable. Thanks for digging in.

libs/core/kiln_ai/datamodel/extraction.py
scosman: Generally I'd prefer runtime checks and no casts. 

Also: I'm not sure how cast would handle a None, which is a valid return. Might work might not - I'd have to check.

scosman (approved):
One take it or leave it comment.

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: Ok - I'll merge this one and do runtime checks here in the next PR

PR #315, opened 2025-05-16
WIP: document, extraction datamodels

leonardmq:
## What does this PR do?

Add datamodel for `Document` (a file usable downstream in RAG), for `Extraction` (the output of an extractor), and set up the relationships between models.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/datamodel/extraction.py
+++ b/libs/core/kiln_ai/datamodel/extraction.py
@@ -1,11 +1,14 @@
 import json
 from enum import Enum
-from typing import Any, List
+from typing import TYPE_CHECKING, Any, List, Union
 
 from pydantic import BaseModel, Field, ValidationError, model_validator
 from typing_extensions import Self
 
-from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnBaseModel, KilnParentedModel
+from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnParentedModel
+
+if TYPE_CHECKING:
+    from kiln_ai.datamodel.project import Project
 
 
 def format_properties_errors(e: ValidationError) -> str:
@@ -54,7 +57,7 @@ class GeminiProperties(BaseModel):
         return self
 
 
-class ExtractorConfig(KilnBaseModel):
+class ExtractorConfig(KilnParentedModel):
     name: str = NAME_FIELD
 
     output_format: OutputFormat = Field(
@@ -97,6 +100,12 @@ class ExtractorConfig(KilnBaseModel):
 
     def gemini_properties(self) -> GeminiProperties:
         return GeminiProperties(**self.properties)
+
+    # Workaround to return typed parent without importing Project
+    def parent_project(self) -> Union["Project", None]:
+        if self.parent is None or self.parent.__class__.__name__ != "Project":
+            return None
+        return self.parent  # type: ignore
 
 
 class ExtractionSource(str, Enum):
--- a/libs/core/kiln_ai/datamodel/extraction.py
+++ b/libs/core/kiln_ai/datamodel/extraction.py
@@ -5,7 +5,7 @@ from typing import Any, List
 from pydantic import BaseModel, Field, ValidationError, model_validator
 from typing_extensions import Self
 
-from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnBaseModel
+from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnBaseModel, KilnParentedModel
 
 
 def format_properties_errors(e: ValidationError) -> str:
@@ -97,3 +97,22 @@ class ExtractorConfig(KilnBaseModel):
 
     def gemini_properties(self) -> GeminiProperties:
         return GeminiProperties(**self.properties)
+
+
+class ExtractionSource(str, Enum):
+    PROCESSED = "processed"
+    PASSTHROUGH = "passthrough"
+
+
+class Extraction(KilnParentedModel):
+    source: ExtractionSource = Field(
+        description="The source of the extraction.",
+    )
+    extractor_config_id: str = Field(
+        description="The ID of the extractor config that was used to extract the data.",
+    )
+
+    # TODO: add extracted data as attachment
+    # output: KilnModelAttachment = Field(
+    #     description="The attachment that was used to extract the data.",
+    # )
--- a/libs/core/kiln_ai/datamodel/project.py
+++ b/libs/core/kiln_ai/datamodel/project.py
@@ -2,10 +2,18 @@ from pydantic import Field
 
 from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnParentModel
 from kiln_ai.datamodel.document import Document
+from kiln_ai.datamodel.extraction import ExtractorConfig
 from kiln_ai.datamodel.task import Task
 
 
-class Project(KilnParentModel, parent_of={"tasks": Task, "documents": Document}):
+class Project(
+    KilnParentModel,
+    parent_of={
+        "tasks": Task,
+        "documents": Document,
+        "extractor_configs": ExtractorConfig,
+    },
+):
     """
     A collection of related tasks.
 
@@ -25,3 +33,6 @@ class Project(KilnParentModel, parent_of={"tasks": Task, "documents": Document})
 
     def documents(self) -> list[Document]:
         return super().documents()  # type: ignore
+
+    def extractor_configs(self) -> list[ExtractorConfig]:
+        return super().extractor_configs()  # type: ignore
--- a/libs/core/kiln_ai/datamodel/project.py
+++ b/libs/core/kiln_ai/datamodel/project.py
@@ -1,10 +1,11 @@
 from pydantic import Field
 
 from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnParentModel
+from kiln_ai.datamodel.document import Document
 from kiln_ai.datamodel.task import Task
 
 
-class Project(KilnParentModel, parent_of={"tasks": Task}):
+class Project(KilnParentModel, parent_of={"tasks": Task, "documents": Document}):
     """
     A collection of related tasks.
 
@@ -21,3 +22,6 @@ class Project(KilnParentModel, parent_of={"tasks": Task}):
     # Needed for typechecking. TODO P2: fix this in KilnParentModel
     def tasks(self) -> list[Task]:
         return super().tasks()  # type: ignore
+
+    def documents(self) -> list[Document]:
+        return super().documents()  # type: ignore
--- a/libs/core/kiln_ai/datamodel/document.py
+++ b/libs/core/kiln_ai/datamodel/document.py
@@ -1,0 +1,68 @@
+from typing import TYPE_CHECKING, List, Union
+
+from pydantic import BaseModel, Field, model_validator
+from typing_extensions import Self
+
+from kiln_ai.datamodel.basemodel import NAME_FIELD, KilnParentedModel, KilnParentModel
+from kiln_ai.datamodel.extraction import Extraction, Kind
+
+if TYPE_CHECKING:
+    from kiln_ai.datamodel.project import Project
+
+
+class FileInfo(BaseModel):
+    filename: str = Field(description="The filename of the file")
+
+    size: int = Field(description="The size of the file in bytes")
+
+    mime_type: str = Field(description="The MIME type of the file")
+
+    # TODO: add attachment
+    # attachment: KilnModelAttachment = Field(
+    #     description="The attachment to the file",
+    # )
+
+
+class Document(
+    KilnParentedModel, KilnParentModel, parent_of={"extractions": Extraction}
+):
+    name: str = NAME_FIELD
+
+    description: str = Field(description="A description for the file")
+
+    original_file: FileInfo = Field(description="The original file")
+
+    # TODO: move {mime_type:kind} mapping out of GeminiExtractor and into here
+    #   - will also need to have models specify which mimetypes they support
+    # TODO: move Kind enum into this file (instead of in extraction.py)
+    #   - waiting for ExtractionConfig PR merged to do this
+    kind: Kind = Field(
+        description="The kind of document. The kind is a broad family of filetypes that can be handled in a similar way"
+    )
+
+    # NOTE: could extract {tags + validate_tags} into a reusable Taggable model and inherit from that here
+    # and in TaskRun
+    # thoughts?
+    tags: List[str] = Field(
+        default_factory=list,
+        description="Tags for the document. Tags are used to categorize documents for filtering and reporting.",
+    )
+
+    @model_validator(mode="after")
+    def validate_tags(self) -> Self:
+        for tag in self.tags:
+            if not tag:
+                raise ValueError("Tags cannot be empty strings")
+            if " " in tag:
+                raise ValueError("Tags cannot contain spaces. Try underscores.")
+
+        return self
+
+    # Workaround to return typed parent without importing Project
+    def parent_project(self) -> Union["Project", None]:
+        if self.parent is None or self.parent.__class__.__name__ != "Project":
+            return None
+        return self.parent  # type: ignore
+
+    def extractions(self) -> list[Extraction]:
+        return super().extractions()  # type: ignore


libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +64,48 @@ def string_to_valid_name(name: str) -> str:
...
+        if not os.path.isfile(self.path):
+            raise ValueError(f"File is not a file: {self.path}")
+        return self
+
+    def copy_to_parent_and_update_path(self, parent_folder: str | Path) -> Path:
+        if isinstance(parent_folder, str):
+            parent_folder = Path(parent_folder)
+
+        # the attachment is already in the parent folder, so we don't need to copy it
+        if self.path.parent == parent_folder:
```
leonardmq: This mutation of the path feels quite side-effecty and the check to infer whether or not it needs copying seems fragile here, so not a huge fan. But if we do the path relativization at the time we set the path (could be more straightforward), we still need to have the original source path around because we need to copy from it further down the line.

Maybe could have `KilnAttachmentModel` have two properties, but not convinced it would be better:
- `path` -> the final relative path
- `path_source` (or similarly named) -> path that needs to be copied; and we clear it after processing and we don't serialize it

Could also lift this method up into the `KilnBaseModel` and do the operations from outside this model, but seems similar

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,88 @@ def string_to_valid_name(name: str) -> str:
...
+        if isinstance(self.path, str):
+            self.path = Path(self.path)
+        if not os.path.exists(self.path):
+            raise ValueError(f"File does not exist: {self.path}")
+        if not os.path.isfile(self.path):
+            raise ValueError(f"Path is not a file: {self.path}")
+        return self
+
+    @model_serializer
+    def serialize(self, info: SerializationInfo) -> dict[str, Path] | None:
```
leonardmq, line 97: This gets called when `self.model_dump_json(...)` gets called in `BaseModel.save_to_file()`. In order to do the actual attachment saving, we need to know the parent object's path (so that we copy the attachment file into the same folder). We pass that parent path information via the `context` when we call `self.model_dump_json(...)` inside `BaseModel.save_to_file()`.

One problem is that `model_dump_json` also gets called from other places (wherever the model is dumped to JSON, for example in the response serialization middleware in FastAPI); and when those calls happen, we don't get the parent path that we need to figure out how to copy the file.

To distinguish between those innocuous calls and calls intending to actually save, in this method we check the `context`.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,88 @@ def string_to_valid_name(name: str) -> str:
     return valid_name.strip("_").strip()
 
 
+class KilnAttachmentModel(BaseModel):
+    path: Path = Field(description="The path to the attachment")
```
leonardmq, line 72: There are ways to serialize `KilnAttachmentModel` into a `string` (instead of `{ path: str }`) during serialization, but I could not get it to work during deserialization as typechecks then expect to deserialize a complex `KilnAttachmentModel` type, but fail when they find a `str`, seemingly before even trying to call the validator

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,88 @@ def string_to_valid_name(name: str) -> str:
...
+        extension = suffix or mimetypes.guess_extension(mime_type) or ".unknown"
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=extension)
+        if isinstance(data, str):
+            temp_file.write(data.encode("utf-8"))
+        else:
+            temp_file.write(data)
+        temp_file.close()
+        return cls(path=Path(temp_file.name))
+
+    def resolve_path(self, parent_path: Path) -> Path:
```
leonardmq, line 157: Another problem is this `KilnAttachmentModel` has no concept of what its own parent is except during serialization in `save_to_file` because we pass it its parent via the `model_dump_json(context={dest_path}`. After deserialization (when loading from file), it has no concept of its own full path either (only relative, which in practice is just `filename.extension`).

This gets problematic when the caller needs to read the file - it needs to do `attachment.resolve_path(xxx)` to give it the path.

We could resolve the path automatically during deserialization - by checking `loading_from_file == True` and adding the `parent_path` in that deserialization context (from `BaseModel.load_from_file(...)`), but that seems fragile.

(This method is here to highlight this and not scatter this problem to too many places)

libs/core/kiln_ai/adapters/extractors/registry.py
```diff
@@ -1,15 +1,22 @@
+from typing import Callable
+
+from google import genai
```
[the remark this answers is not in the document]
leonardmq: The `import` is correct and working already.

Good idea for the LRU cache, I added this in https://github.com/Kiln-AI/Kiln/pull/315/commits/0873255433bc63aad448480432a4f99b1f52fe43

libs/core/kiln_ai/datamodel/test_attachment.py
```diff
@@ -0,0 +1,353 @@
...
+@pytest.fixture
+def test_media_file(tmp_path) -> Path:
+    test_file_path = tmp_path / "test_model.json"
+    data = {"v": 1, "model_type": "kiln_base_model"}
+
+    with open(test_file_path, "w") as file:
+        json.dump(data, file, indent=4)
+
+    return test_file_path
+
```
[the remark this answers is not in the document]
leonardmq: Good catch - removed the redundant fixture here: https://github.com/Kiln-AI/Kiln/pull/315/commits/b90a3c62b5787728461df03039e4e95408cb654b

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -78,15 +109,88 @@
 
     @model_validator(mode="after")
     def validate_properties(self) -> Self:
-        if self.extractor_type == ExtractorType.gemini:
+        if self.extractor_type == ExtractorType.GEMINI:
             validate_prompt_for_kind(self.properties.get("prompt_for_kind"))
             validate_model_name(self.properties.get("model_name"))
             return self
         else:
             raise ValueError(f"Invalid extractor type: {self.extractor_type}")
```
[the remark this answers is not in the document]
leonardmq: Removed the useless `else` here: https://github.com/Kiln-AI/Kiln/pull/315/commits/90d17ad5a75bfa1e0af99e38980346b0f94411b7

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,88 @@ def string_to_valid_name(name: str) -> str:
...
+        its permanent location when the model is saved.
+        """
+        extension = suffix or mimetypes.guess_extension(mime_type) or ".unknown"
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=extension)
+        if isinstance(data, str):
+            temp_file.write(data.encode("utf-8"))
+        else:
+            temp_file.write(data)
+        temp_file.close()
+        return cls(path=Path(temp_file.name))
```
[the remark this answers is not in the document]
leonardmq: This is true. However, not clear currently how to handle that flow - we persist the files to their permanent locations after doing `.save_to_file()` on our `KilnBaseModel`. The flow may go for a while and relatively deep before we are done with the temp file. Deleting the file after copying it (only if the file is in temp folders) could be one way. Any ideas for other approaches that generalize well?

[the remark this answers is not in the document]
leonardmq: Implemented your approach 1 (and some tests): https://github.com/Kiln-AI/Kiln/pull/315/commits/e94bea3955f639ac6cba34e2f3dfdbdf19d88953

leonardmq:
@coderabbitai summary

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -41,36 +31,39 @@ def __init__(self, extractor_config: ExtractorConfig):
...
+        if isinstance(path, str):
+            path = Path(path)
+
         try:
-            mime_type, _ = mimetypes.guess_type(file_info.path)
             if mime_type is None:
-                raise ValueError("Unable to guess file mime type")
+                mime_type, _ = mimetypes.guess_type(path)
+                if mime_type is None:
+                    raise ValueError("Unable to guess file mime type")
```
scosman: nit: add the path to error message. Easier to debug.

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -89,10 +85,10 @@ def _get_kind_from_mime_type(self, mime_type: str) -> Kind | None:
                 return kind
         return None
 
-    def _extract(self, file_info: FileInfoInternal) -> ExtractionOutput:
-        kind = self._get_kind_from_mime_type(file_info.mime_type)
+    def _extract(self, path: Path, mime_type: str) -> ExtractionOutput:
+        kind = self._get_kind_from_mime_type(mime_type)
         if kind is None:
-            raise ValueError(f"Unsupported MIME type: {file_info.mime_type}")
+            raise ValueError(f"Unsupported MIME type: {mime_type}")
```
scosman: nit: add path here as well

libs/core/kiln_ai/adapters/extractors/registry.py
```diff
@@ -1,15 +1,26 @@
...
+
 from kiln_ai.adapters.extractors.base_extractor import BaseExtractor
 from kiln_ai.adapters.extractors.gemini_extractor import GeminiExtractor
-from kiln_ai.datamodel.extraction import ExtractorType
+from kiln_ai.datamodel.extraction import ExtractorConfig, ExtractorType
+from kiln_ai.utils.config import Config
 from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 
 
+@lru_cache(maxsize=1)
```
scosman: Is caching needed? Consider just creating a client per GeminiExtractor instance in GeminiExtractor.__init__. I'm assuming their SDK is designed for serverless environments, so client setup can't be too heavy. Reuse might cause hard to track down bugs with async calls / parallelism.

If needed, I'd move this into the Gemini class, as it's not registry specific:

```
@lru_cache(maxsize=1)
def get_genai_client() -> genai.Client:
    return genai.Client(api_key=Config.shared().gemini_api_key)

def buildGeminiExtractor(extractor_config)
    return   GeminiExtractor(get_genai_client(), extractor_config)
```

Then just use buildGeminiExtractor here.

libs/core/kiln_ai/adapters/extractors/test_registry.py
```diff
@@ -1,14 +1,39 @@
...
+        )
+        assert isinstance(extractor, GeminiExtractor)
+        assert extractor.model_name == "gemini-2.0-flash"
+        assert extractor.gemini_client is not None
 
 
 def test_extractor_adapter_from_type_invalid():
     with pytest.raises(ValueError):
-        extractor_adapter_from_type("invalid-type")  # type: ignore
+        extractor_adapter_from_type("invalid-type", {})  # type: ignore
```
scosman: the type checker should ignore test files by default so you shouldn't need the comment. I don't care about the code, but might make your life easier if you get the environment setup to not warn you.

libs/core/kiln_ai/datamodel/project.py
```diff
@@ -1,10 +1,18 @@
...
 from kiln_ai.datamodel.task import Task
 
 
-class Project(KilnParentModel, parent_of={"tasks": Task}):
+class Project(
+    KilnParentModel,
+    parent_of={
+        "tasks": Task,
+        "documents": Document,
+        "extractor_configs": ExtractorConfig,
```
scosman, line 13: 👍

scosman (changes requested):
First pass. I still need to look at some of the bigger classes (runner, models), but sending some nits here. It's looking good!

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: Added the path in the message

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
leonardmq: Added the path in the message (and also `model_name`)

libs/core/kiln_ai/adapters/extractors/registry.py
leonardmq: The idea was a singleton (and coderabbit suggested the `@lru_cache(maxsize=1)`) and inject it via the constructor to allow passing mocks that way. 

I have moved these two functions into the `GeminiExtractor` class file, and instantiate a new `genai.Client` per `GeminiExtractor`. I just tried removing `get_genai_client()`, but I actually do need it for a registry test where I mock only that and pass in real args to the `GeminiExtractor` to check the instantiation.

I'll have a pass over the tests in a while to see if some stuff can be consolidated in hindsight - probably some tests are a little much

libs/core/kiln_ai/adapters/extractors/test_registry.py
leonardmq: Removed the type ignore comment - it does not show the warnings now, but it did a few days ago. Maybe some of the new settings in the vs code config got it aligned

libs/core/kiln_ai/adapters/extractors/extractor_runner.py
```diff
@@ -0,0 +1,108 @@
...
+    ):
+        if len(extractor_configs) == 0:
+            raise ValueError("Extractor runner requires at least one extractor config")
+
+        self.documents = documents
+        self.extractor_configs = extractor_configs
+
+    def collect_jobs(self) -> List[ExtractorJob]:
+        # all extractor configs come from the same project
+        project = self.extractor_configs[0].parent_project()
```
scosman: remove this block. `[0]` is ugly, and the "all configs must be part of same project" req isn't documented (and I don't think we need it). Don't worry about the few extra iterations, the jobs are 1000x heavier than iterating the docs

```
for extractor_config in self.extractor_configs:
    project = extractor_config.parent_project()
    if project is None:
       raise ValueError("Extractor runner requires a project for each extractor_config")
    for document in self.documents:       
       // check if already extracted, then queue
```

And add the document loop

libs/core/kiln_ai/adapters/extractors/extractor_runner.py
```diff
@@ -0,0 +1,108 @@
...
+    ExtractorConfig,
+)
+from kiln_ai.utils.async_job_runner import AsyncJobRunner, Progress
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ExtractorJob:
+    item: Document
```
scosman: item -> doc

libs/core/kiln_ai/datamodel/basemodel.py
scosman: I like path -> string. It leaves the option to add more fields.

libs/core/kiln_ai/datamodel/basemodel.py
scosman: this looks like it was fun to figure out. Thanks for unpacking it.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,118 @@ def string_to_valid_name(name: str) -> str:
...
+            except (OSError, FileNotFoundError):
+                pass  # file is already gone, that's fine
+
+        self.path = new_path.relative_to(dest_path)
+        self.is_persisted = True
+
+        return {"path": self.path}
+
+    def copy_file_to(self, dest_folder: Path) -> Path:
+        filename = f"{str(uuid.uuid4().int)[:12]}{self.path.suffix}"
```
scosman, line 134: P2: filename would be nice if it was the attachmetns name. IE if the attachmetn was `extraction: KilnAttachment` it would be "extraction.md". We'll have models with several attachments.

Can we just pass it in context?

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,118 @@ def string_to_valid_name(name: str) -> str:
...
+        # if the path is already relative, we consider it has been copied already
+        if self.is_persisted:
+            return {"path": self.path}
+
+        # copy file and update the path to be relative to the dest_path
+        path_before_copy = self.path
+        new_path = self.copy_file_to(dest_path)
+
+        # check if source was a temp file and clean it up
+        if KilnAttachmentModel.is_temp_file(path_before_copy):
```
scosman: Admirable, but I worry about bugs (someone could create an attachment with a temp file they created, and not expect it to get unlinked). 

Can we not leave it for the system sweeps and lose the complexity? I think this is simple and good enough. They'll be cleared when you reboot.

Suggestion: use python `TemporaryFile` in from_data which will clean up for you, remove this code here. We won't touch files we didn't create.

Edit: your already using NAmedTempoary file! Remove this and is_temp_file! It should work when garbage collected, which is good.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,118 @@ def string_to_valid_name(name: str) -> str:
...
+
+    def copy_file_to(self, dest_folder: Path) -> Path:
+        filename = f"{str(uuid.uuid4().int)[:12]}{self.path.suffix}"
+        target_path = dest_folder / filename
+        shutil.copy(self.path, target_path)
+        return target_path
+
+    @classmethod
+    def is_temp_file(cls, path: Path) -> bool:
+        """Check if the path is in a temporary directory."""
```
scosman: see above

scosman (commented):
WIP review

libs/core/kiln_ai/datamodel/basemodel.py
scosman: Yeah, not ideal, but I like not trying for magic.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -71,6 +68,118 @@ def string_to_valid_name(name: str) -> str:
...
+        its permanent location when the model is saved.
+        """
+        extension = suffix or mimetypes.guess_extension(mime_type) or ".unknown"
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=extension)
+        if isinstance(data, str):
+            temp_file.write(data.encode("utf-8"))
+        else:
+            temp_file.write(data)
+        temp_file.close()
+        return cls(path=Path(temp_file.name))
```
scosman, line 155: So we're using path as an absolute path and a relative path. Which could cause bugs (resolve path won't work after creating an attachmetn from_data for example).

Maybe keep this non-realitive path in a new variable (private attribute in Pydantic lingo). This might also remove the need for is_persisted (if path is set, it is, if it's not, it isn't). When we serialize, we check for this value, and do all the needed work if it's set.

We can update resolve_path to return the new var if present (or resolve from parent and .path if not)

I know you've put a lot of work in, so thoughts on if this is a good idea/worth it welcome.

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq: Got paranoid after reading that temp files can linger for weeks depending on the OS, so I wanted to avoid leaving cleanup to the caller. But maybe that is better and ultimately more of the caller's responsibility. Will remove

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq: I have added support for optionally injecting a `filename_prefix` into the serialization `context` via a `@field_serializer` on the (immediate) parent model that contains the `KilnAttachmentModel`. If no such serializer is defined on the parent structure, we fall back to the original behavior with a random numeric filename.

Like:
```py
class ModelWithAttachmentNameOverride(KilnBaseModel):
    attachment: KilnAttachmentModel = Field(default=None)

    @field_serializer("attachment")
    def serialize_attachment(
        self, attachment: KilnAttachmentModel, info: SerializationInfo
    ) -> dict:
        context = info.context or {}
        context["filename_prefix"] = "attachment_override"
        return attachment.model_dump(mode="json", context=context)
```

For example:
- with prefix: `attachment_xyz_12345678910.pdf`
- without prefix: `12345678910.pdf`

We still need the random numeric suffix to prevent collisions when serializing a property like this: 
```
list_of_attachments: List[KilnAttachmentModel] = Field() # all the elements of the list will receive the same context.filename_prefix so we need the random suffix to avoid collisions
```

Commit: https://github.com/Kiln-AI/Kiln/pull/315/commits/ec2afc1491e39f35fffdbafca86bc49901f8a6dd

---

Other ideas I tried:
- Using reflection to collect the field name and pass it down via context would also be problematic when there are nested constructs. For example, in `Document`, we have `original_file: SomeOtherBaseModelThatContainsTheAttachment`, so to find the attachments and their property name, we would need to recursively traverse the structure until we find all the attachment fields to figure out what names they have. 
- If a `BaseModel` has multiple `KilnAttachmentModel` fields, that also would not work via context since the `KilnAttachmentModel` serializer would just get the whole context and would not know which field name is for itself or for a sibling or some other field in the ancestor structures.
- Passing some metadata inside `Field(xxx)`; the data passed into the `Field`'s constructor is not accessible by the model the `Field` contains

libs/core/kiln_ai/datamodel/basemodel.py
scosman: oh sorry, typed this days ago but didn't submit. Let's separate this from this PR as to not block

scosman (approved):

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq: Agree it is fragile - the `is_persisted` was due to another issue with relative / absolute (after the first `save_to_file`).

I think what you outlined can work. Something along these lines:
- force `parent` to always be passed in the constructor in `KilnAttachmentModel` and in `from_data(...)`
- in the constructor, build the `absolute_path` from the parent if we do not already have a `relative_path`
- on save, we build the `relative_path`
- on deserialization, we set `relative_path` (but not `absolute_path` since we don't know the parent during deserialization)
- in `resolve_path(parent)` we return either `absolute_path` if that is set, or deduce it from `parent`

I will go back to this

PR #323, opened 2025-05-22
Updating model list to include Opus and Sonnet 4

tawnymanticore:
## What does this PR do?
Added support for two new Anthropic models.

## Related Issues

NA

## Checklists

- [X] Tests have been run locally and passed (./checks)
- [] New tests have been added to any work in /lib


tawnymanticore:
Screenshot 2025-05-22 at 2 30 07 PM

I verified both models do create data by building the project and seeing if it pops up in the web app. is there a better strategy to test that the model works besides doing that? what tests would you recommend @scosman ?

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -700,6 +702,49 @@ class KilnModel(BaseModel):
...
+    KilnModel(
+        family=ModelFamily.claude,
+        name=ModelName.claude_sonnet_4,
+        friendly_name="Claude Sonnet 4",
+        providers=[
+            KilnModelProvider(
+                name=ModelProviderName.openrouter,
+                model_id="anthropic/claude-sonnet-4",
+                suggested_for_data_gen=True,
+                suggested_for_evals=True,
```
scosman, line 714: Let's untag 3.7 as suggested_for_evals if we think 4 is better (seems safe).

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -700,6 +702,49 @@ class KilnModel(BaseModel):
...
+            ),
+            KilnModelProvider(
+                name=ModelProviderName.anthropic,
+                model_id="claude-sonnet-4-20250514",
+                structured_output_mode=StructuredOutputMode.json_instructions,
```
scosman: Why not `structured_output_mode=StructuredOutputMode.function_calling` like 3.7?

JSON instructions inject extra copy into the prompt to get JSON, and don't have any structure guarantees. Generally the preference order is json_schema, function_calling, json_object, json_object_and_instruction, json_instruction. The earlier ones are more formal systems of structured data, the later are "ask and hope".

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -700,6 +702,49 @@ class KilnModel(BaseModel):
...
+        family=ModelFamily.claude,
+        name=ModelName.claude_opus_4,
+        friendly_name="Claude Opus 4",
+        providers=[
+            KilnModelProvider(
+                name=ModelProviderName.openrouter,
+                model_id="anthropic/claude-opus-4",
+                structured_output_mode=StructuredOutputMode.json_instructions,
+                suggested_for_data_gen=True,
+                suggested_for_evals=True,
```
scosman: Let's remove the suggested tags for Opus. We try to keep those only to a limited few (3/4 per category). For price reasons I'm not recommending o1/o1-pro, etc. I think Opus falls in that bucket (💸 ).

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -700,6 +702,49 @@ class KilnModel(BaseModel):
...
+    # Claude Opus 4
+    KilnModel(
+        family=ModelFamily.claude,
+        name=ModelName.claude_opus_4,
+        friendly_name="Claude Opus 4",
+        providers=[
+            KilnModelProvider(
+                name=ModelProviderName.openrouter,
+                model_id="anthropic/claude-opus-4",
+                structured_output_mode=StructuredOutputMode.json_instructions,
```
scosman: ditto

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -700,6 +702,49 @@ class KilnModel(BaseModel):
...
+                suggested_for_data_gen=True,
+                suggested_for_evals=True,
+            ),
+            KilnModelProvider(
+                name=ModelProviderName.anthropic,
+                model_id="claude-opus-4-20250514",
+                structured_output_mode=StructuredOutputMode.json_instructions,
+                suggested_for_data_gen=True,
+                suggested_for_evals=True,
+            ),
```
scosman, line 740: ditto to above

scosman (changes requested):
Bunch of small comments -- nothing you could have known before. Just learning the system stuff!

Last thing to check before merging is running the paid tests. In Cursor filter tests to sonnet_4 and run each one, one at a time. When you manually click run on a single test It will use paid credits (whereas it usually will skip paid tests)

tawnymanticore:
> Bunch of small comments -- nothing you could have known before. Just learning the system stuff!
> 
> Last thing to check before merging is running the paid tests. In Cursor filter tests to sonnet_4 and run each one, one at a time. When you manually click run on a single test It will use paid credits (whereas it usually will skip paid tests)

Nice yeah sounds good. updated now. I ran the tests and it was super easy, kudos!

All tests passing but one set. Having a bit of trouble with the structured output CoT tests but I got all the other tests to run. I confirmed that the issue is present with Sonnet 3.7 as well
Screenshot 2025-05-23 at 9 45 50 AM

Any idea where to begin debugging this?

libs/core/kiln_ai/adapters/ml_model_list.py
tawnymanticore: ah I did not understand the difference at all, great point

scosman (approved):

scosman:
for tomorrow: feel free to merge anytime after it's approved (which it is)!

PR #329, opened 2025-05-28
small string change to explain that the JSON descriptions are not sent anywhere

tawnymanticore:
## What does this PR do?

Small user experience fix needs to be made here. it's not obvious that the input/output schema description is not sent to a model. wasted some time carefully crafting those descriptions only to realize its more suited to live in the main training prompt.

Used the UI Copy from the other elements. Any suggestions to make it smaller? Ideally it only pops up for Structured JSON
Screenshot 2025-05-28 at 10 17 38 AM

## Checklists

## Summary by CodeRabbit

- **Documentation**
  - Updated the descriptive text in the task creation and editing form to clarify that input and output schema descriptions are for team reference only and not used by the model.

### DIFF
--- a/app/web_ui/src/routes/(fullscreen)/setup/(setup)/create_task/edit_task.svelte
+++ b/app/web_ui/src/routes/(fullscreen)/setup/(setup)/create_task/edit_task.svelte
@@ -365,7 +365,8 @@
         Part {show_requirements ? "3" : "2"}: Input Schema
       </div>
       <div class="text-xs text-gray-500">
-        What kind of input will the model receive?
+        What kind of input will the model receive? This is a description for you
+        and your team, not used by the model.
       </div>
     </div>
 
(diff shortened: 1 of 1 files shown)


scosman:
Okay, my mistake - this is typically passed to the model! We just saw a case of the fine-tune not producing the correct schema. We don't need this warning at the task level, as usually this works correctly.

We're making a choice in the fireworks finetune adapter to just pass json_mode which asks for JSON but doesn't append the schema to the prompt. json_instruction_and_object would do both. 

We could make this an option in the fine-tuning UI, but that would be a separate lower pri task. It's really hard to explain what this is. We could change the default, but including the schema every time seems heavy....

```
# Fireworks doesn't support function calls or json schema, so we'll use json mode at call time
self.datamodel.structured_output_mode = StructuredOutputMode.json_mode
```

PR #331, opened 2025-05-29
WIP: alternative attachment modeling

leonardmq:
## What does this PR do?

Alternative implementation of `KilnAttachmentModel` (from https://github.com/Kiln-AI/Kiln/pull/315) to use less reflection, and use `TypedDict` instead. See which one we prefer.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/datamodel/basemodel.py
+++ b/libs/core/kiln_ai/datamodel/basemodel.py
@@ -7,9 +7,10 @@ from abc import ABCMeta
 from builtins import classmethod
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Type, TypeVar, get_args
+from typing import Annotated, Any, Dict, List, Optional, Type, TypeVar
 
 from pydantic import (
+    AfterValidator,
     BaseModel,
     ConfigDict,
     Field,
@@ -19,7 +20,7 @@ from pydantic import (
     model_validator,
 )
 from pydantic_core import ErrorDetails
-from typing_extensions import Self
+from typing_extensions import Self, TypedDict
 
 from kiln_ai.datamodel.model_cache import ModelCache
 from kiln_ai.utils.config import Config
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -64,43 +65,38 @@ def string_to_valid_name(name: str) -> str:
...
-        if not os.path.exists(self.path):
-            raise ValueError(f"File does not exist: {self.path}")
-        if not os.path.isfile(self.path):
-            raise ValueError(f"File is not a file: {self.path}")
-        return self
 
-    def copy_to(self, dest_folder: str | Path) -> Path:
-        if isinstance(dest_folder, str):
-            dest_folder = Path(dest_folder)
+KilnPath = Annotated[Path, AfterValidator(check_file_exists)]
```
leonardmq, line 76: This is based on an example found here: https://github.com/pydantic/pydantic/discussions/6517#discussioncomment-6392345

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -64,43 +65,38 @@ def string_to_valid_name(name: str) -> str:
...
+    path: KilnPath
 
-        return target_path
 
-    @classmethod
-    def is_attachment_type(cls, annotation: type[Any] | None) -> bool:
-        """Recursively check if KilnAttachmentModel is used anywhere in the type annotation."""
-        if annotation is KilnAttachmentModel:
-            return True
+ATTACHMENT_FIELD = Field(
```
leonardmq, line 83: We could refactor this - possibly to proxy in the `Field` args and merge them together, so the client could still pass in a default factory, override the description, etc. The only thing we care about here is the `json_schema_extra={"attachment_type": "single"}`, which we need to later be able to detect attachments in `collect_attachments_from_fields`

leonardmq:
Found a much better way - closing this

PR #333, opened 2025-05-30
I18N Support

KenRouKoro:
## What does this PR do?

Adding complete I18N support for Kiln-AI's frontend. Currently merged language files include: English, Simplified Chinese, Japanese, French, and Russian.

Added `svelte-i18n` as the I18N infrastructure support.  
All hardcoded UI strings (except for tests) have been converted to use translation keys for localizable text.  

Added a language_switcher.svelte component as the language switcher.

The language switcher is placed at the bottom of the Settings page to maintain consistency with the project's styling.

The CI may flag XSS attack concerns because some translation texts need to render HTML. However, I believe this is not an issue for the current frontend, which only runs in a local environment.

This is a very large commit, and there might be some overlooked hardcoded strings that require further review.  

image
image
image
image
image
image
image

## Related Issues

NONE

I, @KenRouKoro, confirm that I have read and agree to the [Contributors License Agreement](https://github.com/Kiln-AI/Kiln/blob/main/CLA.md).

## Checklists


scosman:
@KenRouKoro Thanks for taking a run at this. I know it's a major effort!

I've been thinking a bunch about internationalization. I don't think we can maintain a full internationalization stack yet. We don't have the language expertise internally, and if just using machine translations I prefer the extension route which does the same thing, and doesn't require the maintenance cost.

I've written up some thinking on it here: https://github.com/Kiln-AI/Kiln/issues/382

Basically:
 - use `translate="no"` to prevent translation of elements which should never be translated (model outputs, etc).
 - Encourage people to use Chrome's translate feature (or another extension)
 - Strategically translate a few key items (sidebar elements, important buttons that don't automatically translate well), without trying to translate everything. Leave 99% to the extension.

What do you think?

KenRouKoro:
> @KenRouKoro Thanks for taking a run at this. I know it's a major effort!
> 
> I've been thinking a bunch about internationalization. I don't think we can maintain a full internationalization stack yet. We don't have the language expertise internally, and if just using machine translations I prefer the extension route which does the same thing, and doesn't require the maintenance cost.
> 
> I've written up some thinking on it here: #382
> 
> Basically:
> 
> * use `translate="no"` to prevent translation of elements which should never be translated (model outputs, etc).
> * Encourage people to use Chrome's translate feature (or another extension)
> * Strategically translate a few key items (sidebar elements, important buttons that don't automatically translate well), without trying to translate everything. Leave 99% to the extension.
> 
> What do you think?

Without considering the use of string keys, most translation plugins work relatively normally, but it's difficult to achieve correct layout and maintain accurate translations (like the classic Microsoft-style Chinese, which is exactly why I'm making this effort). Using translate=no can serve as a temporary measure, but honestly, its effectiveness is not great.
For the PR I submitted, it actually only needs the EN language key to function properly. I've implemented automatic fallback and can wait for the community to submit relevant translation keys.
However, regarding this PR, I have identified some issues that I'm currently fixing. I also plan to implement an independent automated translation system (based on LLM prompt engineering and glossaries rather than just machine translation), and I am still conceptualizing the structure for this at the moment.
Obviously, this will still take some time, so there's no need to rush.

scosman:
I think we have 2 separate issues:

First one is: `translate=no`. This is needed. You can't work on LLM evals if you're seeing content that's not true to the original. Someone else offered to take that, so let's keep that separate from this PR.

Second one is better translations, which you're covering here. I'm open to better translations than the automatic ones. I think I just want to keep the code impact low as the primary devs don't speak Chinese/Japanese/Russian (only English/French). Two ideas to minimize the impact:

1) Keep it focused to strings that are either really important (sidebar items, "OK", etc), or strings that have bugs when automatically translated (people can submit PRs with translations for problematic strings). We set it up so people can fix issues, but we don't aim for 100% coverage in this PR. We know that auto-translate tools will be used along with this. Does that sounds reasonable?

2) We keep the syntax easy so English is still inline as our primary development happens in English. An example: `if (!confirm($_("forms.unsaved_changes_warning")))` is a lot harder to code review and maintain than what we have today. The message isn't inline, and is split across multiple files. Instead something like  `if (!confirm($_("forms.unsaved_changes_warning", "English confirm message here")))` where English is inline, but can be overridden for other languages if needed. 

What do you think?

KenRouKoro:
I basically agree with your two points. I can try to implement the complete functionality of the second point, and rewriting this PR's content may be simpler, but it might take one to two weeks.

scosman:
Sounds great! No rush, and greatly appreciate the effort.

PR #341, opened 2025-06-06
Leonard/document api UI clean

leonardmq:
## What does this PR do?

- Add document and extraction API
- Add document / extraction management UI

## Checklists


app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
leonardmq, line 1: This is a copy / paste from `main`. Will resolve the conflicts when I catch up `rag` with `main` after this gets merged into `rag`

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,390 @@
...
+
+            document = Document(
+                parent=project,
+                name=sanitize_name(name),
+                description=description,
+                kind=kind,
+                original_file=FileInfo(
+                    filename=file.filename or "",
+                    mime_type=content_type,
+                    attachment=KilnAttachmentModel(path=Path(tmp_file.name)),
```
leonardmq: Can just use `KilnAttachmentModel.from_data(await file.read())` instead of writing to filetemp in this controller

scosman:
@coderabbit review

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -206,6 +206,6 @@ async def run_job(self, job: EvalJob) -> bool:
             eval_run.save_to_file()
 
             return True
-        except Exception as e:
-            logger.error(f"Error running eval job for dataset item {job.item.id}: {e}")
+        except Exception:
+            logger.error(f"Error running eval job for dataset item {job.item.id}")
             return False
```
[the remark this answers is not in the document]
scosman: +1

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -73,9 +73,25 @@ def parent_document(self) -> Union["Document", None]:
...
+        if not self.output.is_persisted:
+            return None
+        if not self.path:
+            raise ValueError(
+                "Failed to resolve the path of extraction output attachment because the extraction does not have a path."
+            )
+
+        full_path = self.output.resolve_path(self.path.parent)
+        with open(full_path, "r") as f:
+            return f.read()
```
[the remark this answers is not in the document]
scosman: +1

Also this is synchronous io. Local disk not remote so not a deal breaker (I often leave these), but something to be aware of if you want to do a lot of this. Can async read too.

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -90,7 +106,7 @@
     extractor_type: ExtractorType = Field(
         description="This is used to determine the type of extractor to use.",
     )
-    properties: dict[str, str | int | float | bool | dict[str, str]] = Field(
+    properties: dict[str, str | int | float | bool | dict[str, str] | None] = Field(
         default={},
         description="Properties to be used to execute the extractor config. This is extractor_type specific and should serialize to a json dict.",
     )
```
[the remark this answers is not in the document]
scosman: 🤯 if this is true

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -90,7 +106,7 @@ class ExtractorConfig(KilnParentedModel):
     extractor_type: ExtractorType = Field(
         description="This is used to determine the type of extractor to use.",
     )
-    properties: dict[str, str | int | float | bool | dict[str, str]] = Field(
+    properties: dict[str, str | int | float | bool | dict[str, str] | None] = Field(
```
scosman, line 173: P2 but slight preference to remove value vs setting None.

scosman (changes requested):
WIP review, but will submit these in batches. Nothing major yet, just some nits.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -18,6 +18,7 @@
   import { formatDate } from "$lib/utils/formatters"
   import { goto } from "$app/navigation"
   import DeleteDialog from "$lib/ui/delete_dialog.svelte"
+  import { isMacOS } from "../../../../../../../lib/utils/platform"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: a `lib/utils/platform` here is easier than relative

app/web_ui/src/lib/utils/formatters.ts
```diff
@@ -43,6 +43,16 @@ export function formatDate(dateString: string | undefined): string {
...
 
+export function formatSize(byteSize: number): string {
+  if (byteSize < 1024) {
+    return byteSize + " B"
+  }
+  if (byteSize < 1024 * 1024) {
+    return (byteSize / 1024).toFixed(2) + " KB"
+  }
+  return (byteSize / 1024 / 1024).toFixed(2) + " MB"
+}
```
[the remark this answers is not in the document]
scosman: nice...

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+run_extractor_locks: Dict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
+
+
+def open_folder(path: str | Path) -> None:
+    if sys.platform.startswith("darwin"):
+        subprocess.run(["open", path], check=True)
+    elif sys.platform.startswith("win"):
+        os.startfile(path)  # type: ignore[attr-defined]
+    else:
+        subprocess.run(["xdg-open", path], check=True)
```
scosman, line 49: I have the same code in `def open_logs_folder`. Sorry, I think you had it first but I beat you into main. Can you pull it into a util and delete the dupe as part of PR when you merge main.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+
+
+# TODO: extract out into common utils
```
scosman, line 52: +1, looks like exact copy.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+    )
+    passthrough_mimetypes: list[OutputFormat] = Field(
+        description="The mimetypes to pass through to the extractor",
+        default_factory=list,
+    )
+    extractor_type: ExtractorType = Field(
+        description="The type of the extractor",
+    )
+    properties: dict[str, str | int | float | bool | dict[str, str] | None] = Field(
+        default={},
```
scosman: default_factory? Ask Cursor :P

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+    )
+
+    @model_validator(mode="before")
+    def set_default_name(cls, values: dict) -> dict:
```
scosman: I don't love using before validators to set defaults. I do it only when we're loading from a file (only way I know to do that).

Do this in the default_factory instead? Or just leave it optional, you set a default in the the API call itself.

P2 -- Also: I'm starting realize the "memorable_names" are hard to use. If these are prominent in the UI, considering make a descriptive name string "Gemini 2.5 Pro to Markdown - Daring Badger" (properties + memorable name at end). Ignore this part of these aren't important in the UI.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+    created_by: str
+    source: str
+    extractor_config_id: str
+    output_content: str
+    extractor: ExtractorSummary
+
+
+class CreateExtractorConfigRequest(BaseModel):
+    name: str | None = Field(
+        description="The name of the extractor config",
```
scosman: Use `NAME_FIELD` here. Will make the error message nice if there's invalid data.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+    def set_default_name(cls, values: dict) -> dict:
+        if values.get("name") is None:
+            values["name"] = generate_memorable_name()
+        else:
+            values["name"] = string_to_valid_name(values["name"])
+        return values
+
+    @model_validator(mode="after")
+    def set_default_properties(self):
+        if not isinstance(self.properties, dict):
```
scosman: We have typing, not sure you need this.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            raise ValueError("Properties must be a dictionary")
+
+        match self.extractor_type:
```
scosman: Can we just remove this logic (and gemini_properties_with_defaults)? I'm assuming creating and ExtractorConfig with an invalid extractor_type or invalid extractor_properties already raises an appropriate error? Let that bubble up instead of duplicating the logic. This logic should be in ExtractorConfig, not the API. Don't want to mange all the providers at this level.

libs/server/kiln_server/document_api.py
scosman: Also: maybe just add the properties you expect here instead of a properties dictionary.

It feels like we should either:

 - have a custom request object, with strong types for model_name, prompt_document, etc. This will give us a really nice API, and really good error messages.
 - Remove CreateExtractorConfigRequest and just use ExtractorConfig, without creating a new wrapper.

My preference is for the first. Frontend code will be nice and typed too.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            name=string_to_valid_name(name),
+            description=description,
+            kind=kind,
+            original_file=FileInfo(
+                filename=file.filename or "",
+                mime_type=content_type,
+                attachment=KilnAttachmentModel.from_data(file_data, content_type),
+                size=len(file_data),
+            ),
+        )
```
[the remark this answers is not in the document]
scosman: I disagree w CR here -- see comment above.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        content_type = file.content_type or ""
+        if content_type.startswith("image/"):
+            kind = Kind.IMAGE
+        elif content_type.startswith("video/"):
+            kind = Kind.VIDEO
+        elif content_type.startswith("audio/"):
+            kind = Kind.AUDIO
+        else:
+            kind = Kind.DOCUMENT
+
```
scosman, line 223: if not file.filename raise 422? Seems important.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        document = Document(
+            parent=project,
+            name=string_to_valid_name(name),
```
scosman: `name or file.filename` to give it a default?

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+
+        return document
+
+    @app.get("/api/projects/{project_id}/documents")
+    async def get_documents(
+        project_id: str,
+    ) -> list[Document]:
+        project = project_from_id(project_id)
+
+        # NOTE: maybe add cache here (readonly=True flag)?
```
scosman: yes, please do that!

readonly=false is only needed if you want to mutate the object passed back. We're immediately returning it without mutations.

Background: readonly=true returns the python object from the cache, so it's risky if you make a change but don't save that change to disk. Readonly=false makes a copy, which is needed for mutation, but not here.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+                    "failed_documents": failed_documents,
+                    "error": "Documents not found",
+                },
+            )
+        return {"success": True}
+
+    @app.post("/api/projects/{project_id}/create_extractor_config")
+    async def create_extractor_config(
+        project_id: str,
+        request: CreateExtractorConfigRequest,
```
scosman, line 318: see above comment

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        extractor_config.save_to_file()
+
+        return extractor_config
+
+    @app.get("/api/projects/{project_id}/extractor_configs")
+    async def get_extractor_configs(
+        project_id: str,
+    ) -> list[ExtractorConfig]:
+        project = project_from_id(project_id)
+        return project.extractor_configs()
```
scosman: readonly

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        project = project_from_id(project_id)
+        document = Document.from_id_and_parent_path(document_id, project.path)
+        if not document:
+            raise HTTPException(
+                status_code=404,
+                detail="Document not found",
+            )
+
+        extractor_configs: dict[str, ExtractorConfig] = {}
+        for extraction in document.extractions():
```
scosman: readonly

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+                str(extraction.extractor_config_id), project.path
+            )
+            if extractor_config:
+                extractor_configs[str(extraction.extractor_config_id)] = (
+                    extractor_config
+                )
+
+        return [
+            ExtractionSummary(
+                **extraction.model_dump(exclude={"output"}),
```
scosman: I'd just map them manually. ** could have unintended side effects

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+                output_content=extraction.output_content() or "",
+                extractor=ExtractorSummary(
+                    **extractor_configs[str(extraction.extractor_config_id)].model_dump(
```
scosman: Ditto

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+                        exclude={"properties"}
+                    ),
+                ),
+            )
+            for extraction in document.extractions()
```
scosman: Weird to iterate extractions twice. Why not just iterate once above, building a list of ExtractionSummary. Would be easier to read

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            str(extraction.extractor_config_id), project.path
+        )
+        if not extractor_config:
+            raise HTTPException(
+                status_code=404,
+                detail=f"Extractor config {extraction.extractor_config_id} not found",
+            )
+
+        return ExtractionSummary(
+            **extraction.model_dump(exclude={"output"}),
```
scosman: ditto - maybe a helper since you do it several times.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            output_content=extraction.output_content() or "",
+            extractor=ExtractorSummary(
+                **extractor_config.model_dump(exclude={"properties"}),
+            ),
+        )
+
+    @app.get("/api/projects/{project_id}/documents/{document_id}/download")
+    async def download_document_file(
```
scosman, line 464: nice touch!

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+    async def delete_document(project_id: str, document_id: str) -> dict:
+        project = project_from_id(project_id)
+        document = Document.from_id_and_parent_path(document_id, project.path)
+        if not document:
+            raise HTTPException(
+                status_code=404,
+                detail=f"Document {document_id} not found",
+            )
+
+        document.delete()
```
scosman, line 538: This working well? I think this is our first delete endpoint!?! Deleting the whole directory I'd guess? that works if that's the case.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        document_count_successful = 0
+        for document in project.documents():
+            extractions = document.extractions()
+            if any(
+                extraction.extractor_config_id == extractor_config_id
+                for extraction in extractions
+            ):
+                document_count_successful += 1
+
+        # NOTE: could make sense to persist failed extractions (with some property like "status" / "failed_reason") to
```
scosman: Assuming you can press the "extract" button and get an error there, I'd say safe to delete the comment. We'll add it if people ask.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        extractor_config = ExtractorConfig.from_id_and_parent_path(
+            extractor_config_id, project.path
+        )
+        if extractor_config is None:
+            raise HTTPException(
+                status_code=404,
+                detail="Extractor config not found",
+            )
+
+        document_count_total = len(project.documents())
```
scosman: 1) readonly
2) do `documents = project.documents(readonly=true)` and reuse it for perf reasons.

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+        document_count_successful = 0
+        for document in project.documents():
+            extractions = document.extractions()
```
scosman: readonly

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            ):
+                document_count_successful += 1
+
+        # NOTE: could make sense to persist failed extractions (with some property like "status" / "failed_reason") to
+        # be able to surface failures here (as opposed to pending extractions), and also show the actual error
+        # as some may be due to some provider-specific rejection (e.g. "file too large", "wrong codec", etc.) we cannot resolve
+        return ExtractionProgress(
+            document_count_total=document_count_total,
+            document_count_successful=document_count_successful,
+            current_extractor_config=extractor_config,
```
scosman: "current" isn't quite the right name. Maybe just "extractor_config"

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+            raise HTTPException(
+                status_code=404,
+                detail=f"Document {document_id} not found",
+            )
+
+        if extractor_config_ids is None:
+            extractor_config_ids = [
+                extractor_config.id
+                for extractor_config in [
+                    ec for ec in project.extractor_configs() if not ec.is_archived
```
scosman: readonly

app/web_ui/src/routes/(app)/docs/fileicon.svelte
```diff
@@ -0,0 +1,236 @@
...
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    ><g id="SVGRepo_bgCarrier" stroke-width="0"></g><g
+      id="SVGRepo_tracerCarrier"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+    ></g><g id="SVGRepo_iconCarrier">
+      <path
+        d="M18.505 22H5.495C5.225 22 4.995 21.78 4.995 21.5V3.5C4.995 3.23 5.215 3 5.495 3H18.505C18.775 3 19.005 3.22 19.005 3.5V21.51C18.995 21.78 18.775 22 18.505 22Z"
+        stroke="#0F0F0F"
```
scosman: The key with SVGs is to always replace the colors with `currentColor`, that way the take on the Kiln color automatically (usually default text-color). You need to check stroke and fill and replace any that aren't none with `currentColor`

libs/server/kiln_server/document_api.py
leonardmq: I saw that (and copied your Windows and Linux support). I also pasted another chunk from main, though I can’t recall what but should hit a merge conflict (intentionally).

I am planning to bring `rag` up to date with `main` and resolve merge conflicts after this giant one is merged. I tried rebasing `rag` the other day while other PRs were in flight, but it caused a mega giant diff, so I want to catch it up with few PRs out in the open.

I have rebased rag a couple of times in the past weeks (there were no conflicts). But maybe it’s cleaner to open a PR from `main` into `rag` so we can review and track the merge together?

app/web_ui/src/lib/utils/formatters.ts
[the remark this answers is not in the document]
leonardmq: Updated to use Coderabbit's suggested implementation, and also added some tests for it, in this commit: https://github.com/Kiln-AI/Kiln/pull/341/commits/60cc664b2c65165421a8fac017cd81a1edadbf90

libs/core/kiln_ai/adapters/eval/eval_runner.py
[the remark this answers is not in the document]
leonardmq: Not really sure what happened there because I merged a commit to add the trace in the eval runner, on `main` some days ago: https://github.com/Kiln-AI/Kiln/pull/345/files#diff-8e2333c3f886505b14fb2eda850397a24fe79217bd975fd98fc87d2d3eb6f69d

Reverted the change I made in this file: https://github.com/Kiln-AI/Kiln/pull/341/commits/0d084b00a4941ca37c985d214102ab28b2ae0895 - this way, we make no changes to that in this PR and integrate the log improvement that is on `main`

libs/core/kiln_ai/datamodel/extraction.py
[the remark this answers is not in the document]
leonardmq: Refactored to use the suggested implementation, except we `raise ValueError(xxx)` on every `Exception`. If we get an exception on trying to read the file that is supposed to be there, that is an unexpected breaking error. The `None` return case here has the semantic meaning of there not being any output content to retrieve.

Refactored the error handling here: 
- https://github.com/Kiln-AI/Kiln/pull/341/commits/8be1f7197c1c244d8f293280239197a67866b577

Further refactor to make the read `async`: 
- https://github.com/Kiln-AI/Kiln/pull/341/commits/cc00957bcbc2da5b57dc7f69eddd175a1bf201d1

It seems that `anyio.Path` has the same interface as `pathlib.Path`, but supports `async` operations. 

Should we always be using `anyio.Path` instead of `pathlib.Path` everywhere? If so, then I might do a separate PR to replace all instances of `pathlib.Path` with `anyio.Path` and make the reads async.

libs/core/kiln_ai/datamodel/extraction.py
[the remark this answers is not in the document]
leonardmq: Coderabbit gave me similar comments in the past but then say it is handled fine in pydantic V2. 

For the sake of clarity though, I switched to `default_factory=dict` as that seems more standard - will use that going forward.

Commit: https://github.com/Kiln-AI/Kiln/pull/341/commits/abcca679591b9952cb6222a995abf09f7bd9a4f7

libs/server/kiln_server/document_api.py
```diff
@@ -0,0 +1,695 @@
...
+
+
+def open_folder(path: str | Path) -> None:
+    if sys.platform.startswith("darwin"):
+        subprocess.run(["open", path], check=True)
+    elif sys.platform.startswith("win"):
+        os.startfile(path)  # type: ignore[attr-defined]
+    else:
+        subprocess.run(["xdg-open", path], check=True)
+
```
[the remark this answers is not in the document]
leonardmq: P2: nice to have but not in this PR as I'd need to test.

Blocking the event loop for this folder opening operation is fine for now - opening the folder would release in a matter of milliseconds and the operation will be user-triggered (no concurrency or rapid-fire calls to it).

libs/server/kiln_server/document_api.py
leonardmq: Updated here: https://github.com/Kiln-AI/Kiln/pull/341/commits/97387d51d55d4ee7d5965505a737c696bac31936

Also updated the `patch` structure.

Aside about the `NAME_FIELD`: the `NAME_FIELD` is fine to use for the extractor name, but for documents, I ran into a situation where it gets too restrictive because we only allow a subset of characters that filesystems actually allow - so when we seed the name from the filename of the file the user is uploading, we can fail validation, and if we `string_to_name` it, we make it useless if the filename was originally in Chinese / Japanese / non-ASCII language.

I opened a PR to update the `NAME_FIELD` a few days ago to allow the subset of unicode characters allowed in filenames on Mac / Windows / Linux:
- https://github.com/Kiln-AI/Kiln/pull/364

libs/server/kiln_server/document_api.py
leonardmq: Removed here: https://github.com/Kiln-AI/Kiln/pull/341/commits/61923f9d6e7866377810209b1319d8c440b883a9

Also not sure why I had this here 🤷

libs/server/kiln_server/document_api.py
leonardmq: As far as I have tested, it does. 

The deletion you implemented does `shutil.rmtree(dir_path)` on the model's path, so everything down that folder also gets deleted so when you delete `document`, its `extractions` also get deleted, as they should

libs/server/kiln_server/document_api.py
leonardmq: Removed comment here: https://github.com/Kiln-AI/Kiln/pull/341/commits/c5545da7ac145eabaad5cc45a2b1331946d4db29

The main issue on the UX front is that we don't surface the error messages in the UI - so the user knows `n` errors occurred during the job, but they don't know what / why / which subtasks failed. There is the same quirk with evals. 

As an example of where this can get problematic is I picked Gemini for G-Eval (which it cannot actually work with because it does not support logprobs). It would fail every single task, and tell me `n errors` without the details of what was failing. I retried a few times, but every time it was managing to make some expensive Gemini call (for COT) before failing on a different second call it was trying to make. I can imagine similar scenarios with extractions, where for example some file is gigantic and Gemini hits output limits or a connection timeout and causes a failure and then on every retry, we retry to parse that file again - just a hypothetical, not sure if this exact scenario can happen.

I made changes to pass on the error logs from the `AsyncJobRunner` on to the frontend and show them in the progress dialog, but I ended up removing it because it was a bit clunky and expanded the scope of this PR too much

libs/server/kiln_server/document_api.py
leonardmq: Renamed here: https://github.com/Kiln-AI/Kiln/pull/341/commits/b7360277269f9dbe7ae2b540118ad46b663db934

libs/server/kiln_server/document_api.py
leonardmq: Thank you for the explanation - I was not sure if I should be using it due to the `KilnAttachmentModel` statefulness, but it seems to be fine since it follows the same lifecycle as the object it is in / gets saved at the same time.

Updated the `readonly` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826

libs/server/kiln_server/document_api.py
leonardmq: Updated the `readonly` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826

libs/server/kiln_server/document_api.py
leonardmq: Updated the `readonly` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826

libs/server/kiln_server/document_api.py
leonardmq: Updated here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826 

and removed an intermediate var afterwards here: https://github.com/Kiln-AI/Kiln/pull/341/commits/5831e6efa560ca84e815c33384a115b1bca039d6

libs/server/kiln_server/document_api.py
leonardmq: Updated the `readonly` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826

libs/server/kiln_server/document_api.py
leonardmq: Updated the `readonly` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/392dc21e748f1ffed9f4349f7ca2e3cf19c22826

libs/server/kiln_server/document_api.py
leonardmq: This is indeed very bizarre (I think because I initially was doing something else in there).

Refactored here: https://github.com/Kiln-AI/Kiln/pull/341/commits/6799ecaa2612b723e6bbb3673de307d052346a65

(Will also add the manual mapping to the output type, in a separate commit)

libs/server/kiln_server/document_api.py
leonardmq: Made a helper to manually map to `ExtractionSummary` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/a7b0326e6f65fe72d2be9bed89c3221f447de1db

libs/server/kiln_server/document_api.py
leonardmq: Agreed that is better.

Made a helper to manually map to `ExtractionSummary` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/a7b0326e6f65fe72d2be9bed89c3221f447de1db

libs/server/kiln_server/document_api.py
leonardmq: Made a helper to manually map to `ExtractionSummary` here: https://github.com/Kiln-AI/Kiln/pull/341/commits/a7b0326e6f65fe72d2be9bed89c3221f447de1db

libs/server/kiln_server/document_api.py
leonardmq: Can I leave the comment and do it after merging this to `rag` and catching it up with `main` to avoid risking merge conflicts on this one?

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: Updated the super long import paths to `$lib` in all the components under `docs/` here:
- https://github.com/Kiln-AI/Kiln/pull/341/commits/fa4a1b9013249d42664610ec5e7d6520fe3c1062

libs/server/kiln_server/document_api.py
leonardmq: Added fallback here: https://github.com/Kiln-AI/Kiln/pull/341/commits/ae30f0ccb5d95b357cb66a257dec73012f948fdc

Side-note - for non-ascii based languages (e.g. Chinese, Japanese, etc.), `string_to_valid_name` would wipe the whole name; for example currently:
- `机器学习_朴素贝叶斯.pdf` would become just `pdf`

https://github.com/Kiln-AI/Kiln/pull/364 would allow for unicode characters

libs/server/kiln_server/document_api.py
leonardmq: Added `422` on `file.filename` missing: https://github.com/Kiln-AI/Kiln/commit/ae30f0ccb5d95b357cb66a257dec73012f948fdc

libs/server/kiln_server/document_api.py
scosman: The diff looks clean here, odd it's not clean. If you want to rebase, merge main in, or clean up after - any are fine

scosman:
on the run tab I'm getting an infinite loop of requesting `"GET /api/available_models`, failing, and retrying. Clean working tree with no other content.

Update: issue is a local fine-tune I have which is newer than the data model expects. Not related to this PR be general running new data on old build issue. I guess I'll fix it since I can repo?

libs/server/kiln_server/document_api.py
leonardmq: Seems like these models where we have a `properties` object whose shape depends on another field would be a good use case for discriminated unions? 

https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators

I have not used this yet so not sure what to expect so would rather do in a background lower-priority PR. If we can clean up the validation of these overall, that would be a nice gain

libs/server/kiln_server/document_api.py
leonardmq: Moved into `default_factory` on the field. However, I was using `NAME_FIELD` and does not seem like there is a way to set a `default_factory` on it - so had to copy / paste the `Field` definition (with `default_factory` added in). 

Commit: https://github.com/Kiln-AI/Kiln/pull/341/commits/7ecccd2bf27e2669bb9d0695c588cb98be39fea2#diff-59c66efe8ce4c5dbda13c7f5773ac9e5c36e4a5137217350e2c5957cd8d3d1fd

(Actually forgot to remove it, and removed it here: https://github.com/Kiln-AI/Kiln/pull/341/commits/32a6502aaaa203206ed231ed171f096a91500d77)

Aside: the ASCII filename PR here: https://github.com/Kiln-AI/Kiln/pull/364 should make it possible to alter the name field on the caller since the validator for it was moved into an annotated type

libs/server/kiln_server/document_api.py
leonardmq: Removed this and moved validation / defaults into the datamodel: https://github.com/Kiln-AI/Kiln/pull/341/commits/7ecccd2bf27e2669bb9d0695c588cb98be39fea2

scosman (approved):

PR #345, opened 2025-06-08
chore: improve logs around litellm and eval failure

leonardmq:
## What does this PR do?

1. clarifies the log message in `log_failure_event()` method, which warns that a sync call was used, because that call can also be triggered by *async* calls that fail early due to client-side errors like parameter validation inside LiteLLM (`litellm.UnsupportedParamsError`)
2. log stack trace (`exc_info=True`) in the `eval_runner`, otherwise we get the error message without its origin
3. removed an unused import

Example of when `await litellm.acompletion()` can call the sync `log_failure_event` logging method:
- while running a `g_eval` using a Gemini model, LiteLLM raises `litellm.UnsupportedParamsError` because `top_logprobs` is not supported for that model (they raise in `_check_valid_arg()`)

## Summary by CodeRabbit

- **Documentation**
  - Improved comments and error messages to clarify the handling of synchronous and asynchronous error cases.

- **Chores**
  - Removed an unused import to streamline the codebase.

- **Bug Fixes**
  - Enhanced error logging to include detailed exception tracebacks for better diagnostics during evaluation jobs.

### DIFF
--- a/app/desktop/log_config.py
+++ b/app/desktop/log_config.py
@@ -165,9 +165,11 @@ class CustomLiteLLMLogger(CustomLogger):
         )
 
     def log_failure_event(self, kwargs, response_obj, start_time, end_time):
+        # This logging method is supposed to be called by Litellm in synchronous error cases (Kiln should use async calls instead)
+        # but it appears to also be getting called in async calls that fail early (e.g. UnsupportedParamsError).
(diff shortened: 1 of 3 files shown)


leonardmq:
@coderabbitai summary

PR #356, opened 2025-06-09
super small change, should check the input schema not the output

tawnymanticore:
## What does this PR do?
Small bug, we have a lot of test coverage on JSON as input in our toy tests but when pipe-cleaning an example without JSON as input, it does not work! its due to this check here where we are loading the input if the output is JSON

## Checklists


scosman (approved):
Good catch!

PR #358, opened 2025-06-10
Add chat formatter utility with tests

scosman:
## Summary
- add base ChatFormatter abstraction and individual formatters per strategy
- expose `get_chat_formatter` helper via adapters package
- update unit tests for new formatter classes

## Testing
- `uvx ruff check --select I`
- `uvx ruff format --check .`
- `uv run pyright . --stats`
- `uv run python3 -m pytest --benchmark-quiet -q libs/core/kiln_ai/adapters/chat/test_chat_formatter.py`

------
https://chatgpt.com/codex/tasks/task_e_68485c4e38288332890092f9c28521d7

### DIFF
--- a/libs/core/kiln_ai/adapters/__init__.py
+++ b/libs/core/kiln_ai/adapters/__init__.py
@@ -17,6 +17,7 @@ The eval submodule contains the code for evaluating the performance of a model.
 """
 
 from . import (
+    chat,
     data_gen,
     eval,
     fine_tune,
@@ -28,6 +29,7 @@ from . import (
 
 __all__ = [
     "model_adapters",
+    "chat",
     "data_gen",
     "fine_tune",
     "ml_model_list",
--- a/libs/core/kiln_ai/adapters/chat/__init__.py
+++ b/libs/core/kiln_ai/adapters/chat/__init__.py
@@ -1,0 +1,8 @@
+from .chat_formatter import (
+    ChatFormatter,
+    ChatMessage,
+    ChatStrategy,
+    get_chat_formatter,
+)
+
+__all__ = ["ChatFormatter", "ChatMessage", "ChatStrategy", "get_chat_formatter"]
--- a/libs/core/kiln_ai/adapters/chat/chat_formatter.py
+++ b/libs/core/kiln_ai/adapters/chat/chat_formatter.py
@@ -1,0 +1,159 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Literal, Optional
+
+from kiln_ai.adapters.model_adapters.base_adapter import COT_FINAL_ANSWER_PROMPT
+
+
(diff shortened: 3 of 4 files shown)


libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -199,26 +199,40 @@ def build_prompt(self) -> str:
...
+                user_input=input,
+                thinking_instructions=cot_prompt,
+            )
         elif cot_prompt:
             # 2: Unstructured output with COT
             # Two calls to separate the thinking from the final response
-            return "cot_two_call", cot_prompt
+            # TODO non legacy mode
+            return get_chat_formatter(
+                strategy=ChatStrategy.two_message_cot_legacy,
```
tawnymanticore: QQ so why is this still legacy? can confirm that this is working as expected. ChatStrategy.two_message_cot_legacy is a double system prompt and ChatStrategy.two_message_cot is the correct behavior

tawnymanticore:
some issues on train/test time maybe. here is a sample from training a Reasoning model (what was sent to FW):

```{
  "messages": [
    {
      "role": "system",
      "content": "..."
    },
    {
      "role": "user",
      "content": "The input is:\n<user_input>\n ... \n</user_input>\n\n..."
    },
    {
      "role": "assistant",
      "content": "..."
    },
    {
      "role": "user",
      "content": "Considering the above, return a final result."
    },
    {
      "role": "assistant",
      "content": "{\"difficulty_factor\": 1.4}"
    }
  ]
}
```
and here is the log when sending an eval request in the Run Methods of this same reasoning model when using the Fine-tuned Prompt as the prompt. this looks like the legacy config to me:
```
{
  "model": "...",
  "messages": [
    {
      "role": "system",
      "content": "...."
    },
    {
      "role": "user",
      "content": "...."
    },
    {
      "role": "system",
      "content": "..."
    },
    {
      "role": "assistant",
      "content": "..."
    },
    {
      "role": "user",
      "content": "Considering the above, return a final result."
    }
  ],
}
```


edit: can confirm that the expected behavior is observed with `ChatStrategy.two_message_cot`. nit: although curious why this param matters? I thought the model should use the training prompt no matter what if I select Fine-Tuned Prompt?

tawnymanticore:
vanilla models all look OK though

**Vanilla models**
In a brand new Task (not clone, but New Task, then copy paste prompt/requirements etc just in case)
- Basic and CoT with GPT4o mini
    - Works
    - Note: correct CoT without 2 system prompts is without using legacy mode
- R1 32B
    - Basic works
    - CoT works

In a Legacy Task
- Basic and CoT with GPT4o mini
    - Works
    - Note: correct CoT without 2 system prompts is without using legacy mode
- R1 32B
    - Basic works
    - CoT works

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
scosman: Fixed

PR #361, opened 2025-06-11
Add confirm dialog for model warnings

leonardmq:
## Summary

Goal: make sure users are aware that the model they pick may cause downstream issues (e.g. if they pick Gemini as a Geval evaluator).

Changes:
- warn users via `confirm` dialog when selecting models with known issues
- reset dropdown value if the user clicks `cancel` on the `confirm` dialog
- centralize dropdown warning strings

Confirm dialog:
image

## Summary by CodeRabbit

- **New Features**
  - Added confirmation prompts when selecting models that do not support logprobs, helping users avoid accidental selections.
  - Standardized warning messages and improved visual consistency in the model selection interface.

- **Bug Fixes**
  - Enhanced feedback when choosing unsupported models, reducing user errors.

### DIFF
--- a/app/web_ui/src/routes/(app)/run/available_models_dropdown.svelte
+++ b/app/web_ui/src/routes/(app)/run/available_models_dropdown.svelte
@@ -212,14 +212,14 @@
   />
 
   {#if selected_model_untested}
-    <Warning warning_message={UNTESTED_WARNING} />
+    <Warning large_icon warning_message={UNTESTED_WARNING} />
   {:else if selected_model_unsupported}
     {#if requires_data_gen}
-      <Warning warning_message={DATA_GEN_WARNING} />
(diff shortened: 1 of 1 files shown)


scosman:
I think we should scope this to the logprobs warning. That's a 99.9% chance of failure, but trying a model we haven't tested is pretty normal and shouldn't be showing a ugly browser alert. Thoughts?

leonardmq:
> I think we should scope this to the logprobs warning. That's a 99.9% chance of failure, but trying a model we haven't tested is pretty normal and shouldn't be showing a ugly browser alert. Thoughts?

I have now removed the warnings for the other messages and kept only the one for unsupported logprobs. 

Wasn’t too sure how likely the other issues were to result in hard failures, but agreed it could get tedious if we show disruptive warnings for everything

PR #364, opened 2025-06-13
fix: allow non-ascii in model names

leonardmq:
## What does this PR do?

The `NAME_FIELD` and `SHORT_NAME_FIELD` Pydantic fields we reuse across different `KilnBaseModel` implementations are more restrictive than they need to be as they reject a range of values that would be valid filenames (e.g. non-ASCII letters from other languages like Chinese, Japanese, etc.).

According to: https://en.wikipedia.org/wiki/Filename#Problematic_characters - only a dozen or so special characters are problematic in filenames.

Note: I intentionally kept the validator and sanitization separate. But we could also merge them together to automatically sanitize. We could also use a library like: https://pypi.org/project/pathvalidate/ that validates paths and can normalize them.

## Checklists

## Summary by CodeRabbit

- **Documentation**
  - Updated descriptions for name fields to provide clearer, entity-specific guidance.

- **Bug Fixes**
  - Improved validation and normalization of names to prevent invalid or unsupported characters.

- **Tests**
  - Enhanced and expanded test coverage for name validation and Unicode handling.

### DIFF
--- a/libs/core/kiln_ai/datamodel/basemodel.py
+++ b/libs/core/kiln_ai/datamodel/basemodel.py
@@ -2,22 +2,17 @@ import json
 import os
 import re
 import shutil
+import unicodedata
 import uuid
 from abc import ABCMeta
 from builtins import classmethod
 from datetime import datetime
 from pathlib import Path
-from typing import (
-    Any,
-    Dict,
-    List,
-    Optional,
-    Type,
-    TypeVar,
-)
+from typing import Any, Callable, Dict, List, Optional, Type, TypeVar
 
 from pydantic import (
     BaseModel,
+    BeforeValidator,
     ConfigDict,
     Field,
     ValidationError,
@@ -26,7 +21,7 @@ from pydantic import (
     model_validator,
 )
 from pydantic_core import ErrorDetails
-from typing_extensions import Self
+from typing_extensions import Annotated, Self
 
 from kiln_ai.datamodel.model_cache import ModelCache
 from kiln_ai.utils.config import Config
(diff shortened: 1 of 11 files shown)


libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -44,33 +39,64 @@
...
+            raise ValueError(f"Input should be a valid string, got {type(name)}")
+        if len(name) < min_length:
+            raise ValueError(
+                f"Name is too short. Min length is {min_length} characters, got {len(name)}"
+            )
+        if len(name) > max_length:
+            raise ValueError(
+                f"Name is too long. Max length is {max_length} characters, got {len(name)}"
+            )
+        if string_to_valid_name(name) != name:
```
leonardmq, line 64: How about automatically returning the sanitized `name` here instead of raising an error?

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -44,33 +39,64 @@
...
+            raise ValueError(
+                f"Name is invalid. The name cannot contain any of the following characters: {FORBIDDEN_CHARS}"
+            )
+        return name
+
+    return fn
 
 
 def string_to_valid_name(name: str) -> str:
```
leonardmq, line 73: Alternatively, we could use something like https://pypi.org/project/python-slugify/ to slugify the string into ASCII in here. Would be safer

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -44,33 +39,64 @@
...
+            raise ValueError(f"Input should be a valid string, got {type(name)}")
+        if len(name) < min_length:
+            raise ValueError(
+                f"Name is too short. Min length is {min_length} characters, got {len(name)}"
+            )
+        if len(name) > max_length:
+            raise ValueError(
+                f"Name is too long. Max length is {max_length} characters, got {len(name)}"
+            )
+        if string_to_valid_name(name) != name:
```
leonardmq, line 64: How about automatically returning the sanitized `name` here instead of raising an error?

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq, line 73: We could alternatively use this in the validator to verify and / or normalize the string: https://pypi.org/project/pathvalidate/

Or we could also use something like https://pypi.org/project/python-slugify/ to slugify the string. But that one could alter the string significantly in some cases (e.g. non-Latin characters)

scosman:
Can we add control characters (0x00–0x1F) (non-printable characters)?

Edit: you already did

libs/core/kiln_ai/datamodel/basemodel.py
scosman: I like owning this. We're going beyond path reqs into Kiln reqs here (cross OS, style, etc)

libs/core/kiln_ai/datamodel/basemodel.py
scosman: On the fence. Let's not for now. It would be a great think in the UI, and a bad thing in the library IMO.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -44,33 +39,64 @@
...
+            raise ValueError(
+                f"Name is invalid. The name cannot contain any of the following characters: {FORBIDDEN_CHARS}"
```
scosman: Not your bug, but let's improve this with ", consecutive whitespace/underscores, or leading/trailing whitespace/underscores"

scosman (approved):
on suggested string change, but can merge after.

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/364/commits/862a28f9ceb7e19365a64648bde1aaf756da5242

PR #365, opened 2025-06-13
refactor: explicit input_path versus path distinction and better tests

leonardmq:
Following up on this CR comment from a previous PR on the `KilnAttachmentModel`: 
- https://github.com/Kiln-AI/Kiln/pull/315#discussion_r2130737553

## Summary
- store temporary attachment path in `_absolute_path`
- treat attachments as persisted once `path` is set
- update attachment serializer and resolver
- adapt attachment tests for new behaviour

## Testing
- `uvx ruff check --select I`
- `uvx ruff format --check .`
- `uv run pyright .`
- `uv run python3 -m pytest --benchmark-quiet -q .`

## Summary by CodeRabbit

- **Refactor**
  - Improved attachment handling by distinguishing between absolute input paths and relative persisted paths, enhancing validation, serialization, and file operations for attachments.

- **Tests**
  - Updated tests to use dynamic mock file factories instead of static files, improving flexibility and isolation.
  - Added fixtures and utilities for generating mock files and attachments of various MIME types.
  - Expanded Gemini extractor tests to cover multiple MIME types with granular test cases.

- **Bug Fixes**
  - Ensured output content is always accessible for persisted attachments.

- **Documentation**
  - Added detailed README documenting sources and licensing of test assets.
  - Added new test asset documents explaining thermodynamics of ice cubes cooling drinks.

- **Chores**
  - Removed unused imports and duplicate logger definitions for cleaner code.


scosman:
@coderabbit review

leonardmq:
@coderabbitai summary

libs/core/kiln_ai/tests/fixtures/attachment.py
```diff
@@ -0,0 +1,72 @@
+import shutil
```
leonardmq: These are global fixtures that are very convenient to have when dealing with test files. 

I would like to expand it to include more file samples (one of each file type we support); but realizing that we probably need to take these out of `libs/core/kiln_ai/` otherwise they get packaged up into the library and will bloat it a lot?

Would moving them into `libs/tests/xxx` solve this?

libs/core/kiln_ai/adapters/extractors/test_extractor_runner.py
```diff
@@ -70,9 +61,7 @@ def mock_document(test_pdf_file, mock_project) -> Document:
             filename="test.pdf",
             size=100,
             mime_type="application/pdf",
-            attachment=KilnAttachmentModel(
-                path=test_pdf_file,
-            ),
+            attachment=KilnAttachmentModel.from_file(test_pdf_file),
```
leonardmq, line 64: The intended usage is for callers to use `KilnAttachmentModel.from_file(path)` or `KilnAttachmentModel.from_data(bytes)` to create an attachment.

Have not looked into whether we can then make `input_path` and `path` private but still serializable - but that would be the ideal there so users are not exposed to the internal `path` / `input_path` clunkiness

leonardmq:
@coderabbitai full review

libs/core/kiln_ai/adapters/extractors/test_gemini_extractor.py
```diff
@@ -257,10 +253,11 @@ async def test_extract_image(model_name, test_data_dir):
...
-async def test_extract_video(model_name, test_data_dir):
+async def test_extract_video(model_name, mock_file_factory):
+    test_video_file = mock_file_factory(MockFileFactoryMimeType.MP4)
     extractor = paid_gemini_extractor(model_name=model_name)
-    output = await extractor.extract(
-        path=str(test_data_dir / "big_buck_bunny_sample.mp4"),
+    output = extractor.extract(
+        path=str(test_video_file),
         mime_type="video/mp4",
     )
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/365/commits/954e44e4c0636f519afb9734775a9264252446da

libs/core/kiln_ai/adapters/extractors/test_gemini_extractor.py
```diff
@@ -270,10 +267,11 @@
...
-async def test_extract_audio(model_name, test_data_dir):
+async def test_extract_audio(model_name, mock_file_factory):
+    test_audio_file = mock_file_factory(MockFileFactoryMimeType.OGG)
     extractor = paid_gemini_extractor(model_name=model_name)
-    output = await extractor.extract(
-        path=str(test_data_dir / "poacher.ogg"),
+    output = extractor.extract(
+        path=str(test_audio_file),
         mime_type="audio/ogg",
     )
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/365/commits/954e44e4c0636f519afb9734775a9264252446da

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -69,28 +69,62 @@ def string_to_valid_name(name: str) -> str:
...
+    path: Path | None = Field(
+        default=None,
+        description="The path to the attachment relative to the parent model's path",
+    )
 
-    is_persisted: bool = Field(
-        default=False,
-        exclude=True,
-        description="Whether the attachment is persisted to its permanent location on disk. This is set automatically when the attachment is loaded from a file.",
+    input_path: Path | None = Field(
```
leonardmq, line 77: @coderabbitai - I'd ideally want callers to not be able to access `path` and `input_path` directly, and instead go through the `from_data()` and `from_file` class methods to instantiate a `KilnAttachmentModel`.

However, `path` must still be serializable, and both `input_path` and `path` should be accessible within the class's methods.

Is there a way to do that in `pydantic` (V2)?

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -152,18 +192,37 @@ def from_data(
         else:
             temp_file.write(data)
         temp_file.close()
-        return cls(path=Path(temp_file.name))
+        return cls(input_path=Path(temp_file.name))
+
+    @classmethod
+    def from_file(cls, path: Path | str) -> Self:
```
leonardmq, line 198: The intended usage now is for callers to use `KilnAttachmentModel.from_file(path)` or `KilnAttachmentModel.from_data(bytes)` to create an attachment - and not set `input_path` or `path` manually anywhere. Encapsulates the clunky statefulness

libs/core/kiln_ai/datamodel/basemodel.py
leonardmq: This almost works, but fails during deserialization because `model_validate` is getting called on the parent, and that seems to trigger a call to the `__init__` in this model, which thens hits `raise ValueError("Use from_file() or from_data() class methods to create attachments")`

Might revisit in the future

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -132,8 +134,6 @@ def parent_document(self) -> Union["Document", None]:
         return self.parent  # type: ignore
 
     async def output_content(self) -> str | None:
-        if not self.output.is_persisted:
```
leonardmq, line 135: No longer need any such check since the `resolve_path` call further down will return a path both when the attachment is persisted and when it has not been persisted yet

leonardmq:
Total size for the `assets/` directory is `1.1MB`

If we want to delete the old files in the repo's history to save some space, the files to delete are:
- libs/core/kiln_ai/tests/data/1706.03762v7.pdf
- libs/core/kiln_ai/tests/data/big_buck_bunny_sample.mp4
- libs/core/kiln_ai/tests/data/kodim23.png
- libs/core/kiln_ai/tests/data/poacher.ogg

libs/core/kiln_ai/tests/data/1706.03762v7.pdf
scosman, line 1: great choice

scosman (approved):

PR #375, opened 2025-06-20
Design work for Docs/RAG

scosman:
See check-ins for detailed comments. A bunch of UI string/style changes across the new UI. More visually consistent with the rest of the app.

## Summary by CodeRabbit

- **New Features**
  - Added user-friendly file type labels and improved file type validation for document uploads, with clear warnings for unsupported files.
  - Introduced new utility functions for formatting file sizes, displaying MIME types, and capitalizing strings.
  - Enhanced the extraction results display with a dedicated output viewer and improved property lists.

- **UI Improvements**
  - Updated icons and visual styles across document and extractor sections for a cleaner, modern look.
  - Refined navigation and table layouts, making rows clickable and improving interactivity.
  - Reorganized forms and dialogs for adding documents, CSVs, and extractors, with clearer labels and descriptions.
  - Improved empty state messages and illustrations throughout the app.
  - Renamed menu items and section labels for clarity and consistency.
  - Enhanced extractor list and detail views with better layout and conditional action buttons.

- **Bug Fixes**
  - Fixed file upload button behavior to prevent unsupported files from being uploaded.
  - Prevented event propagation on extractor control buttons to avoid unintended side effects.

- **Tests**
  - Expanded and updated test coverage for formatting utilities.


scosman:
@corerabbit review

scosman:
@CodeRabbit review

app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte
```diff
@@ -120,21 +116,13 @@
         bValue = b.name
         break
       case "kind":
-        aValue = a.kind
-        bValue = b.kind
-        break
-      case "description":
-        aValue = a.description
-        bValue = b.description
+        aValue = a.original_file.mime_type
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte:123]: `kind` and `mime_type` have different purposes and different information value to the user - the `kind` is a bucket of mime types that determines the extraction prompt, which users would need to know about in practice.

Contiguous mimetypes are not necessarily the same `kind` - e.g. `text/csv` and `application/pdf` are both `document` kind and would be running through the same extraction pipeline, but they would be sorted very far apart from each other here as their mimetype prefixes are different.

If we want the same mimetypes to be contiguous, we should sort by `(kind, mime_type)` rather by either one alone, so as not to lose the ability to filter by similarly-processed files (of the same `kind`)

app/web_ui/src/lib/utils/formatters.test.ts
```diff
@@ -44,4 +44,103 @@ describe("formatters", () => {
...
+          expect(mime_type_to_string(input)).toBe(expected)
+        })
+      })
+    })
+
+    describe("generic video types", () => {
+      const videoTestCases = [
+        { input: "video/mp4", expected: "Video (mp4)" },
+        { input: "video/webm", expected: "Video (webm)" },
+        { input: "video/avi", expected: "Video (avi)" },
```
leonardmq, line 104: Not important for this test, but we should hopefully not get `video/avi` - the official `.avi` mimetype is `video/x-msvideo`

app/web_ui/src/lib/utils/formatters.test.ts
```diff
@@ -44,4 +44,103 @@ describe("formatters", () => {
...
+      videoTestCases.forEach(({ input, expected }) => {
+        it(`should return ${expected} for ${input}`, () => {
+          expect(mime_type_to_string(input)).toBe(expected)
+        })
+      })
+    })
+
+    describe("generic audio types", () => {
+      const audioTestCases = [
+        { input: "audio/mp3", expected: "Audio (mp3)" },
```
leonardmq, line 117: Not important for this test, but we should hopefully not get `audio/mp3` - the official `.mp3` mimetype is `audio/mpeg` (https://stackoverflow.com/questions/61363934/chrome-81-mimetype-change-to-file-upload-of-audio-mp3)

app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte
```diff
@@ -611,27 +615,7 @@
...
-                      {:else if document.original_file.mime_type.startsWith("text/")}
-                        Text ({document.original_file.mime_type.split("/")[1]})
-                      {:else if document.original_file.mime_type.startsWith("video/")}
-                        Video ({document.original_file.mime_type.split("/")[1]})
-                      {:else if document.original_file.mime_type.startsWith("audio/")}
-                        Audio ({document.original_file.mime_type.split("/")[1]})
-                      {:else}
-                        {document.original_file.mime_type}
-                      {/if}
+                      {mime_type_to_string(document.original_file.mime_type)}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte:618]: The mapping logic from `mime_type -> kind` dictates the prompt that is used in the extractor. That logic is controlled on the backend - doing it on the UI side is somewhat duplicative of that, but we also lose the info of which prompt was used here. Mime type suffixes are also not very user-friendly - MP3 comes out as `MPEG`; `.mov` comes out as `quicktime`; could be confusing to users. 

Using the `extension` (e.g. `txt`, `mov`, etc.) might convey what type it is more clearly, and would align more closely with how we communicate the allowed filetypes in the upload file dialog here.

I suggest something like `${kind} (${extension..replace(/^\./, "").toUpperCase()})` for all the documents so we preserve info on both what pipeline / prompt it ran with, and what filetype / format the file is, and don't get into the weeds of mapping. For example `Document (PDF), Image (PNG), Document (CSV), Video (MOV)`

app/web_ui/src/routes/(app)/docs/library/[project_id]/[document_id]/+page.svelte
```diff
@@ -159,9 +159,7 @@
           <div class="text-xl font-bold">Document Extractions</div>
           <div class="text-gray-500 text-sm">
             Each
-            <a
-              class="link"
-              href={`http://localhost:5173/docs/extractors/189194447826`}
+            <a class="link" href={`/docs/extractors/189194447826`}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/library/[project_id]/[document_id]/+page.svelte:162]: The path refers to a hardcoded `189194447826` project

app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte:123]: Fixed.

app/web_ui/src/lib/utils/formatters.test.ts
scosman: ah, Cursor wrote that 😀. Test case still holds, even though we won't shouldn't that.

app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/docs/library/[project_id]/+page.svelte:618]: I like that, but do we have extension reliably?

PR #388, opened 2025-06-25
feat: chunking datamodel and chunker

leonardmq:
## What does this PR do?

Set up data model and fixed window chunker to chunk output text from extractions

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Introduced a modular text chunking framework with base and fixed window chunkers.
  * Added configuration options for chunk size and overlap with validation and default values.
  * Enabled management of chunker configurations within projects.
  * Added data models and APIs for chunked documents and their relationships to extractions and projects.
  * Provided a registry to instantiate chunkers by type.

* **Bug Fixes**
  * Enforced validation rules ensuring chunk overlap does not exceed chunk size and values are positive integers.

* **Tests**
  * Added comprehensive unit and integration tests covering chunking logic, configurations, data models, and parent-child relationships.

* **Chores**
  * Added a new dependency: `llama-index-core>=0.12.42`.

### DIFF
--- a/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
@@ -1,0 +1,37 @@
+import logging
+from abc import ABC, abstractmethod
+
+from pydantic import BaseModel, Field
+
+from kiln_ai.datamodel.chunk import ChunkerConfig
+
+logger = logging.getLogger(__name__)
+
+
+class Chunk(BaseModel):
+    text: str = Field(description="The text of the chunk.")
+
+
+class ChunkOutput(BaseModel):
+    chunks: list[Chunk] = Field(description="The chunks of the text.")
+
+
+class BaseChunker(ABC):
+    """
+    Base class for all chunkers.
+
+    Should be subclassed by each chunker.
+    """
+
+    def __init__(self, chunker_config: ChunkerConfig):
+        self.chunker_config = chunker_config
+
+    async def chunk(self, text: str) -> ChunkOutput:
+        if not text:
+            return ChunkOutput(chunks=[])
+
+        return await self._chunk(text)
+
+    @abstractmethod
+    async def _chunk(self, text: str) -> ChunkOutput:
+        pass
--- a/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
@@ -8,12 +8,12 @@ from kiln_ai.datamodel.chunk import ChunkerConfig
 logger = logging.getLogger(__name__)
 
 
-class Chunk(BaseModel):
+class TextChunk(BaseModel):
     text: str = Field(description="The text of the chunk.")
 
 
-class ChunkOutput(BaseModel):
-    chunks: list[Chunk] = Field(description="The chunks of the text.")
+class ChunkingResult(BaseModel):
+    chunks: list[TextChunk] = Field(description="The chunks of the text.")
 
 
 class BaseChunker(ABC):
@@ -26,12 +26,12 @@ class BaseChunker(ABC):
     def __init__(self, chunker_config: ChunkerConfig):
         self.chunker_config = chunker_config
 
-    async def chunk(self, text: str) -> ChunkOutput:
+    async def chunk(self, text: str) -> ChunkingResult:
         if not text:
-            return ChunkOutput(chunks=[])
+            return ChunkingResult(chunks=[])
 
         return await self._chunk(text)
 
     @abstractmethod
-    async def _chunk(self, text: str) -> ChunkOutput:
+    async def _chunk(self, text: str) -> ChunkingResult:
         pass
--- a/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
@@ -4,7 +4,7 @@ from unittest.mock import patch
 import pytest
 from llama_index.core.text_splitter import SentenceSplitter
 
-from kiln_ai.adapters.chunkers.base_chunker import ChunkOutput
+from kiln_ai.adapters.chunkers.base_chunker import ChunkingResult
 from kiln_ai.adapters.chunkers.fixed_window_chunker import FixedWindowChunker
 from kiln_ai.datamodel.chunk import (
     ChunkerConfig,
@@ -35,7 +35,7 @@ async def test_fixed_window_chunker_chunk_empty_text(
     # we should not even be calling the splitter if the text is empty
     chunker = mock_fixed_window_chunker_factory(100, 10)
     with patch.object(SentenceSplitter, "split_text") as mock_split_text:
-        assert await chunker.chunk("") == ChunkOutput(chunks=[])
+        assert await chunker.chunk("") == ChunkingResult(chunks=[])
         mock_split_text.assert_not_called()
 
 
--- a/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
@@ -6,11 +6,7 @@ from llama_index.core.text_splitter import SentenceSplitter
 
 from kiln_ai.adapters.chunkers.base_chunker import ChunkingResult
 from kiln_ai.adapters.chunkers.fixed_window_chunker import FixedWindowChunker
-from kiln_ai.datamodel.chunk import (
-    ChunkerConfig,
-    ChunkerType,
-    FixedWindowChunkerProperties,
-)
+from kiln_ai.datamodel.chunk import ChunkerConfig, ChunkerType
 
 
 @pytest.fixture
@@ -20,13 +16,24 @@ def mock_fixed_window_chunker_factory() -> Callable[[int, int], FixedWindowChunk
             ChunkerConfig(
                 name="test-chunker",
                 chunker_type=ChunkerType.FIXED_WINDOW,
-                properties=FixedWindowChunkerProperties(
-                    chunk_size=chunk_size, chunk_overlap=chunk_overlap
-                ),
+                properties={"chunk_size": chunk_size, "chunk_overlap": chunk_overlap},
             )
         )
 
     return create_chunker
+
+
+async def test_fixed_window_chunker_wrong_chunker_type(
+    mock_fixed_window_chunker_factory,
+):
+    with pytest.raises(ValueError):
+        FixedWindowChunker(
+            ChunkerConfig(
+                name="test-chunker",
+                chunker_type="wrong-chunker-type",
+                properties={"chunk_size": 100, "chunk_overlap": 10},
+            )
+        )
 
 
 async def test_fixed_window_chunker_chunk_empty_text(
--- a/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
@@ -1,0 +1,265 @@
+from typing import Callable
+from unittest.mock import patch
+
+import pytest
+from llama_index.core.text_splitter import SentenceSplitter
+
+from kiln_ai.adapters.chunkers.base_chunker import ChunkOutput
+from kiln_ai.adapters.chunkers.fixed_window_chunker import FixedWindowChunker
+from kiln_ai.datamodel.chunk import (
+    ChunkerConfig,
+    ChunkerType,
+    FixedWindowChunkerProperties,
+)
+
+
+@pytest.fixture
+def mock_fixed_window_chunker_factory() -> Callable[[int, int], FixedWindowChunker]:
+    def create_chunker(chunk_size: int, chunk_overlap: int) -> FixedWindowChunker:
+        return FixedWindowChunker(
+            ChunkerConfig(
+                name="test-chunker",
+                chunker_type=ChunkerType.FIXED_WINDOW,
+                properties=FixedWindowChunkerProperties(
+                    chunk_size=chunk_size, chunk_overlap=chunk_overlap
+                ),
+            )
+        )
+
+    return create_chunker
+
+
+async def test_fixed_window_chunker_chunk_empty_text(
+    mock_fixed_window_chunker_factory,
+):
+    # we should not even be calling the splitter if the text is empty
+    chunker = mock_fixed_window_chunker_factory(100, 10)
+    with patch.object(SentenceSplitter, "split_text") as mock_split_text:
+        assert await chunker.chunk("") == ChunkOutput(chunks=[])
+        mock_split_text.assert_not_called()
+
+
+@pytest.mark.parametrize(
+    "chunk_size,chunk_overlap,expected_chunks",
+    [(12, 6, 43), (256, 12, 2), (1024, 64, 1), (2048, 128, 1)],
+)
+async def test_fixed_window_chunker_concrete_chunker(
+    chunk_size, chunk_overlap, expected_chunks, mock_fixed_window_chunker_factory
+):
+    """
+    This test is to ensure that the chunker can split text (with markdown syntax). The specific values are just an illustration rather than values we
+    particularly care about.
+    """
+    chunker = mock_fixed_window_chunker_factory(chunk_size, chunk_overlap)
+    text_to_chunk = """# How Ice Cubes Make Drinks Colder
+
+## Introduction
+
+When you drop an ice cube into a drink, it does more than just float and look refreshing. It changes the thermal state of the liquid in a precise and physically predictable way. While it may seem like a simple act, the science behind how ice cubes make drinks colder is a fascinating interplay of thermodynamics, phase change, and heat transfer.
+
+## The Science of Cooling
+
+### Heat Transfer Basics
+
+At the core of the process is the concept of **heat exchange**. Heat naturally flows from warmer objects to colder ones until thermal equilibrium is reached. When an ice cube, which is at 0°C (32°F), is placed in a drink that is warmer than that, heat begins to flow from the liquid to the ice. This transfer of energy cools the drink while simultaneously warming the ice.
+
+### Latent Heat of Fusion
+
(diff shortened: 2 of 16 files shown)


leonardmq:
@coderabbitai review

libs/core/kiln_ai/adapters/chunkers/base_chunker.py
```diff
@@ -0,0 +1,37 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Chunk(BaseModel):
+    text: str = Field(description="The text of the chunk.")
+
+
+class ChunkOutput(BaseModel):
+    chunks: list[Chunk] = Field(description="The chunks of the text.")
+
```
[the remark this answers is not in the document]
leonardmq: True - fixed in https://github.com/Kiln-AI/Kiln/pull/388/commits/2fdad718629fa9da6ee325378ca2a61bc78b467e

Renamed `ChunkOutput` into `ChunkingResult` and `Chunk` into `TextChunk`

libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -21,10 +21,12 @@
     KilnParentedModel,
     KilnParentModel,
 )
+from kiln_ai.datamodel.chunk import DocumentChunked
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/388/commits/2fdad718629fa9da6ee325378ca2a61bc78b467e

libs/core/kiln_ai/adapters/chunkers/test_registry.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+from kiln_ai.adapters.chunkers.fixed_window_chunker import FixedWindowChunker
+from kiln_ai.adapters.chunkers.registry import chunker_adapter_from_type
+from kiln_ai.datamodel.chunk import ChunkerConfig, ChunkerType
+
+
+def test_extractor_adapter_from_type():
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/388/commits/2fdad718629fa9da6ee325378ca2a61bc78b467e

libs/core/kiln_ai/adapters/chunkers/test_registry.py
```diff
@@ -0,0 +1,28 @@
...
+        ),
+    )
+    assert isinstance(chunker, FixedWindowChunker)
+    assert chunker.chunker_config.properties.chunk_size == 5555
+    assert chunker.chunker_config.properties.chunk_overlap == 1111
+
+
+def test_chunker_adapter_from_type_invalid():
+    with pytest.raises(ValueError):
+        chunker_adapter_from_type("invalid-type", {})
```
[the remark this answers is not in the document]
leonardmq: Do you think that matters in this particular case?

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,94 @@
...
+
+class ChunkerConfig(KilnParentedModel):
+    name: str = NAME_FIELD
+    description: str | None = Field(
+        default=None, description="The description of the chunker config"
+    )
+    chunker_type: ChunkerType = Field(
+        description="This is used to determine the type of chunker to use.",
+    )
+    properties: FixedWindowChunkerProperties = Field(
```
scosman: Usually we just have a dictionary here, then validators that check the right properties exist for the type. IS there a reason we a Pydantic model? Will be hard to extend when we add types.

See evalConfig for example.

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,94 @@
...
+    @field_serializer("attachment")
+    def serialize_attachment(
+        self, attachment: KilnAttachmentModel, info: SerializationInfo
+    ) -> dict:
+        context = info.context or {}
+        context["filename_prefix"] = "chunk_attachment"
+        return attachment.model_dump(mode="json", context=context)
+
+
+class DocumentChunked(KilnParentedModel):
```
scosman: ChunkedDocument for name?

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,94 @@
...
+    name: str = NAME_FIELD
+    description: str | None = Field(
+        default=None,
+        description="The description of the document chunked.",
+    )
```
scosman: Can we remove name and description? They are defined by the config_id and parent extraction. They don't need names, and I wouldn't want to ask users to add them

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: I wanted to try going with a discriminated union (https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators), but it turns out that only works if the `discriminator` (in our case, `chunker_type`) is inside the unioned object - so that means inside `properties`, not alongside it.

To make discriminated union work, we would need to refactor the model like this:
```py
class Config(KilnBaseModel):
  name: xxx
  properties: Union[
    FixedWindowChunkerProperties,
    AnotherChunkerProperties,
  ] = Field(discriminator="chunker_type")

class FixedWindowChunkerProperties(pydantic.BaseModel):
  chunker_type: Literal["fixed_window"]
  chunk_size: int = Field()
  chunk_overlap: int = Field()

class AnotherChunkerProperties(pydantic.BaseModel):
  chunker_type: Literal["another_chunker"]
  color: str = Field()
  city: str = Field()
```

The main upside (if it worked) is that we would get discriminated typing in the OpenAPI schema and in the UI. It also localizes validation logic into the specific models and avoids bloating the base config validator; validation would be smoother / safer - the setting of `default` values is also nicer. Could also help with backward compatibility glue as we could hide it away in the specific variants without bloating the base config.

That said, since the `discriminator` needs to move into `properties`; refactoring this way would be too inconsistent with the other models (`EvalConfig`, `ExtractorConfig`, ...). So I'll revert back to a plain `dict` and manual validation.

I tried a couple of other ideas before, but found these clunky too:
- plain union (no `discriminator`); that works fine so long as there is no validation error, if there is a validation error, it moves on to the next type in the union and tries that one
- `@model_validator(mode="before")` to cast the incoming `dict`
- one model per `properties` variant and call `TheVariantProperties.model_validate(self.properties)` in a `@model_validator` on the `XXXConfig` model, and have a getter that returns a typed object too; but:
  - `TheVariantProperties.model_validate(self.properties)` inside a `@model_validator` loses the `loc` context (vis a vis the model it is inside of) when raising validation errors
  - wasteful since those objects would only be transient

TLDR: reverting to plain dict, manual validation for now - but would be nice if we can figure out something better

leonardmq: Replaced typed object with `dict` for `ChunkerConfig->properties` in https://github.com/Kiln-AI/Kiln/pull/388/commits/2a3c8e025bfd104ceca17b11440ef4b4358d6452

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Was on the fence with this one and went back and forth; replaced with `ChunkedDocument` (also replaced derivatives like `document_chunked`, `documents_chunked`) in https://github.com/Kiln-AI/Kiln/pull/388/commits/b8349888e102ae69b219acea7d0e604b715e6ce0

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Was on the fence with this as well, the thinking (with no strong conviction) was: 
- the `name` is used in file paths so could have been some value in keeping it (and would have been set on the backend, not user-facing)
- `description` could have been user-facing and used by the user to annotate the particular case (e.g. "Wow this one is weird because it puts the XYZ into chunks")

Removed `name` and `description` in https://github.com/Kiln-AI/Kiln/pull/388/commits/45c339fe34afee7474fb4a1cd751c4c747cbe406

Tangential:
- we currently store `Chunk` inside this `ChunkedDocument`, but `Chunk` itself is not a model
- however, when comes time to generate embeddings for each `Chunk`, a `Chunk` may have many embeddings; so this raises the question of whether `Chunk` should be its own `KilnBaseModel`; 
  - now I think it probably should be its own model as that makes it more conceptually natural to then have `Chunk->Embeddings`; considering each embedding needs to have its own properties - at least a `model_name`; 
  - querying seems probably smoother (e.g. `embedding_for_chunk(chunk_id, model_name)`
  - could store all the embeddings inside the `Chunk` - e.g. `Chunk(attachment, embeddings=[Embedding(model_name=xyz, vector=[1,2,3, ...])])` - but seems bloated and possibly awkward concurrency if we write embeddings in parallel
  - noting: chunking currently is synchronous, fast, and cheap; but we will likely implement semantic chunking down the line, which is asynchronous, slow, and expensive (but better quality); so re-doing independent chunking jobs for each embedding model configuration is not an option

libs/core/kiln_ai/adapters/chunkers/base_chunker.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Base class for all chunkers.
+
+    Should be subclassed by each chunker.
+    """
+
+    def __init__(self, chunker_config: ChunkerConfig):
+        self.chunker_config = chunker_config
+
+    async def chunk(self, text: str) -> ChunkingResult:
```
leonardmq, line 29: We don't need this to be `async` at the moment, but we most certainly will if we implement more advanced chunking (e.g. semantic chunking); so easier to just make it `async` right away

pyproject.toml
```diff
@@ -9,6 +9,7 @@ dependencies = [
     "kiln-ai==0.5.3",
     "kiln-server",
     "kiln-studio-desktop",
+    "llama-index-core>=0.12.42",
```
leonardmq, line 12: Socket.dev report on this dependency:
- https://socket.dev/pypi/package/llama-index-core/overview/0.12.42/tar-gz

Docs on the `SentenceSplitter` used in the `FixedWindowChunker`:
- https://docs.llamaindex.ai/en/stable/api_reference/node_parsers/sentence_splitter/

libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
```diff
@@ -0,0 +1,272 @@
...
+)
+async def test_fixed_window_chunker_concrete_chunker(
+    chunk_size, chunk_overlap, expected_chunks, mock_fixed_window_chunker_factory
+):
+    """
+    This test is to ensure that the chunker can split text (with markdown syntax). The specific values are just an illustration rather than values we
+    particularly care about.
+    """
+    chunker = mock_fixed_window_chunker_factory(chunk_size, chunk_overlap)
+    text_to_chunk = """# How Ice Cubes Make Drinks Colder
```
leonardmq, line 61: The text is from our test Markdown file generated using ChatGPT; here:
- https://github.com/Kiln-AI/Kiln/blob/rag/tests/assets/document_ice_cubes.md

libs/core/kiln_ai/adapters/chunkers/test_fixed_window_chunker.py
```diff
@@ -0,0 +1,272 @@
...
+)
+async def test_fixed_window_chunker_concrete_chunker_zh(
+    chunk_size, chunk_overlap, expected_chunks, mock_fixed_window_chunker_factory
+):
+    """
+    This test is to ensure that the chunker can split Chinese text. The specific values are just an illustration rather than values we
+    particularly care about.
+    """
+    chunker = mock_fixed_window_chunker_factory(chunk_size, chunk_overlap)
+    text_to_chunk = """火山是地表下在岩浆库中的高温岩浆及其有关的气体、碎屑从行星的地壳中喷出而形成的，具有特殊形態的地质结构。
```
leonardmq, line 95: This is an excerpt from the `zh-CN` Wikipedia article about volcanoes:
- https://zh.wikipedia.org/zh-cn/%E7%81%AB%E5%B1%B1

leonardmq:
@coderabbitai - I removed the unused imports from your `Nitpick comments` in https://github.com/Kiln-AI/Kiln/pull/388/commits/9a0d8714e46bff78f2620472ba26c11cb27f1d02

Not sure if you can remember instructions, but if you can, do raise unused imports in PRs as actionable comments, not just nitpick. I'll also open a separate PR to have `ruff` flag unused imports

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,135 @@
...
+        if not isinstance(chunk_overlap, int):
+            raise ValueError("Chunk overlap must be an integer.")
+        if chunk_overlap < 0:
+            raise ValueError("Chunk overlap must be greater than or equal to 0.")
+
+    if chunk_size is not None:
+        if not isinstance(chunk_size, int):
+            raise ValueError("Chunk size must be an integer.")
+        if chunk_size < 0:
+            raise ValueError("Chunk size must be greater than 0.")
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/388/commits/8eac7094ad60d4287c533f0dad8361fce9f43fbe

pyproject.toml
scosman: Good call on this library!

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,135 @@
...
+) -> dict[str, str | int | float | bool]:
+    """Validate the properties for the fixed window chunker and set defaults if needed."""
+    chunk_overlap = properties.get("chunk_overlap")
+    chunk_size = properties.get("chunk_size")
+
+    # avoid a situation where user provides a chunk_overlap > DEFAULT_CHUNK_SIZE
+    if chunk_overlap is not None and chunk_size is None:
+        raise ValueError("Chunk size is required if chunk overlap is provided.")
+
+    if chunk_overlap is not None:
```
scosman: make both field required. While we have a default now, we may change it later, so we always want to save the value used into the model. Better to be explicit for persisted data models.

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,135 @@
...
+            raise ValueError("Chunk size must be greater than 0.")
+
+    if (
+        isinstance(chunk_overlap, int)
+        and isinstance(chunk_size, int)
+        and chunk_overlap >= chunk_size
+    ):
+        raise ValueError("Chunk overlap must be less than chunk size.")
+
+    def default_if_none(value: int | None, default: int) -> int:
```
scosman: Can remove all this logic after above change.

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,135 @@
...
+    def chunk_overlap(self) -> int | None:
+        if self.properties.get("chunk_overlap") is None:
+            return None
+        if not isinstance(self.properties["chunk_overlap"], int):
+            raise ValueError("Chunk overlap must be an integer.")
+        return self.properties["chunk_overlap"]
+
+
+class Chunk(BaseModel):
+    attachment: KilnAttachmentModel = Field(description="The attachment of the chunk.")
```
scosman: name it `content` or `chunk_content`? Attachment is the type, but doesn't say what it is. Same goes for the description "The content of the chunk, stored as an attachment"

libs/core/kiln_ai/datamodel/chunk.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @field_serializer("attachment")
+    def serialize_attachment(
+        self, attachment: KilnAttachmentModel, info: SerializationInfo
+    ) -> dict:
+        context = info.context or {}
+        context["filename_prefix"] = "chunk_attachment"
```
scosman: +1 for `content`, but I like that they will have prefixes

libs/core/kiln_ai/datamodel/chunk.py
scosman: oh, that would be nice. I did always wish for a less hardcoded approach.

libs/core/kiln_ai/datamodel/test_extraction_chunk.py
```diff
@@ -0,0 +1,175 @@
...
+
+            # Verify the complete structure
+            assert config.name == "test-chunker"
+            assert config.chunker_type == ChunkerType.FIXED_WINDOW
+            assert config.chunk_size() == 256
+            assert config.chunk_overlap() == 10
+            assert len(doc.chunks) == 2
+            assert doc.chunks[0].content == attachment
+            assert doc.chunks[1].content == attachment
+
```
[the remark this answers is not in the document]
leonardmq: Why would we need these test files after exiting the context? Seems better to me to auto-delete on context exit in tests

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/388/commits/cae765fa6a1ec0f4145055407a01e8a44255eb41 - removed the defaults (and would be setting those one level up - e.g. API layer)

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/388/commits/94fc9ce5d2a91d8ea46fe599aa8dc57ba57fdf34

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/388/commits/cae765fa6a1ec0f4145055407a01e8a44255eb41

libs/core/kiln_ai/datamodel/chunk.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/388/commits/94fc9ce5d2a91d8ea46fe599aa8dc57ba57fdf34

(also added a test for this particular prefix: https://github.com/Kiln-AI/Kiln/pull/388/commits/e314116a2fb652575fef28bf8add6fa289a78eb9)

PR #389, opened 2025-06-27
chore: flag unused imports

leonardmq:
## What does this PR do?

- Remove all currently unused imports.
- Add check for rule `F401` in `uvx ruff check` script (in `checks.sh`); ref: https://docs.astral.sh/ruff/rules/unused-import/
- Alter one awkward `test_provider_api` test that was mocking `os` relative to the `provider_api.py` file where it was imported but unused
  - should centralize `os` config somewhere to avoid scattered `os.environ` that are difficult to mock / track down

## Checklists

## Summary by CodeRabbit

* **Chores**
  * Cleaned up and consolidated import statements across multiple files to remove unused imports and simplify code structure.
  * Enhanced linting checks to detect unused imports in addition to import sorting.
* **Tests**
  * Updated test fixtures to patch environment variables more directly.
  * Removed unused imports and streamlined import statements in test files.

### DIFF
--- a/app/desktop/studio_server/provider_api.py
+++ b/app/desktop/studio_server/provider_api.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from typing import Any, Dict, List
--- a/app/desktop/studio_server/test_provider_api.py
+++ b/app/desktop/studio_server/test_provider_api.py
@@ -254,7 +254,7 @@ async def test_connect_openrouter():
 
 @pytest.fixture
 def mock_environ():
-    with patch("app.desktop.studio_server.provider_api.os.environ", {}) as mock_env:
+    with patch("os.environ", {}) as mock_env:
         yield mock_env
 
 
--- a/checks.sh
+++ b/checks.sh
@@ -14,8 +14,8 @@ headerStart="\n\033[4;34m=== "
 headerEnd=" ===\033[0m\n"
 
 echo "${headerStart}Checking Python: Ruff, format, check${headerEnd}"
-# I is import sorting
-uvx  ruff check --select I
+# I is import sorting, F401 is unused imports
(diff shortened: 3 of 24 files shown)


checks.sh
```diff
@@ -14,8 +14,8 @@ headerStart="\n\033[4;34m=== "
 headerEnd=" ===\033[0m\n"
 
 echo "${headerStart}Checking Python: Ruff, format, check${headerEnd}"
-# I is import sorting
-uvx  ruff check --select I
+# I is import sorting, F401 is unused imports
+uvx  ruff check --select I,F401
```
leonardmq, line 18: Reference: https://docs.astral.sh/ruff/rules/unused-import/

app/desktop/studio_server/provider_api.py
```diff
@@ -1,5 +1,4 @@
 import logging
-import os
```
leonardmq, line 2: `os` was not being used in this file, but it was being mocked in `test_provider_api.py`, but relative to here

(Another comment about this)

app/desktop/studio_server/test_provider_api.py
```diff
@@ -254,7 +255,7 @@ async def test_connect_openrouter():
 
 @pytest.fixture
 def mock_environ():
-    with patch("app.desktop.studio_server.provider_api.os.environ", {}) as mock_env:
```
leonardmq, line 257: `os` was imported in `provider_api.py`, but not used directly - so the patch here was somewhat misleading. Mocking `os.environ` seems equivalent here (and tests pass).

Worth double checking though.

Longer-term, we should probably centralize all the `os.getenv` / `os.environ` to avoid things like this. Haven't looked at the build, so not sure which env vars need to be packaged into what builds (e.g. lib, desktop, etc.) so not doing that now

scosman (approved):
Thanks! I've wanted this for a while!

PR #390, opened 2025-06-27
refactor: use LiteLLM in extractor

leonardmq:
## What does this PR do?

Refactor to use LiteLLM in document extractor.

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Added support for multimodal and document extraction capabilities in model metadata and selection.
  * Introduced the LiteLLM extractor for document, image, audio, and video content extraction.
  * Enhanced model selection UI with dynamic filtering to show only models supporting document extraction.
  * Added base64 URL encoding utilities for file data handling.

* **Improvements**
  * Expanded supported model providers and models with multimodal capabilities and MIME type details.
  * Strengthened validation and error handling for extractor configuration creation.
  * Added model provider and model name fields to extractor configurations and APIs.
  * Updated UI to display model provider information in extractor details and listings.

* **Bug Fixes**
  * Improved error handling when creating extractor configurations with invalid or unsupported models.
  * Added validation to ensure document paths are set before attachment processing.

* **Tests**
  * Added extensive tests for LiteLLM extractor, encoding utilities, and multimodal model support.
  * Removed tests related to the deprecated Gemini extractor and replaced with LiteLLM tests.
  * Added tests validating model provider lookup and multimodal attribute consistency.
  * Enhanced remote config tests for multimodal field serialization and forward compatibility.
  * Expanded adapter registry tests to verify provider authentication details and configuration consistency.

* **Chores**
  * Removed Gemini extractor and associated code.
  * Updated internal schemas and data models to include multimodal and document extraction fields.
  * Refactored extractor adapter registry to replace Gemini with LiteLLM extractor.
  * Centralized provider authentication detail retrieval and simplified adapter configuration logic.


libs/core/kiln_ai/datamodel/extraction.py
```diff
@@ -78,19 +78,19 @@ class ExtractionModel(BaseModel):
 EXTRACTION_MODEL_LIST = {
     ExtractorType.GEMINI: [
         ExtractionModel(
```
leonardmq: Refactor this - control if model supports all modalities

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -112,23 +115,88 @@ async def _extract(self, path: Path, mime_type: str) -> ExtractionOutput:
...
-        response = await self.gemini_client.aio.models.generate_content(
+        # FIXME: the supports_pdf_input check does not seem to work for gemini models
+        # if not supports_pdf_input(self.model_name):
+        #     raise ValueError(f"Model {self.model_name} does not support PDF input")
+
+        # FIXME: check vision support
+        # supports_vision(model=self.model_name)
+
+        def encode_file(path: Path, mime_type: str) -> dict[str, Any]:
+            def to_base64_url(mime_type: str, bytes: bytes) -> str:
```
leonardmq: There are various other shapes mentioned in the docs to pass so and so modalities, but some do not seem to work (or perhaps work only for some providers).

These few work

libs/core/kiln_ai/adapters/extractors/gemini_extractor.py
```diff
@@ -112,23 +115,81 @@ async def _extract(self, path: Path, mime_type: str) -> ExtractionOutput:
         if prompt is None:
             raise ValueError(f"No prompt found for kind: {kind}")
 
-        response = await self.gemini_client.aio.models.generate_content(
+        # FIXME: the supports_pdf_input check does not seem to work for gemini models
```
leonardmq, line 118: Could not get these `supports_xyz(model)` to work (at least with Gemini). Will need to try again

leonardmq:
Pausing for now - getting back to it after LiteLLM embedding adapter

libs/core/kiln_ai/datamodel/extraction.py
leonardmq: Let's merge the extraction models with the list of Kiln models:
- add field like `mime_types_supported: List[str]` on each model
- add flags like `supports_doc_extraction`

Then update APIs that serve the models to filter for specific flags; then update UI to use these models in the Extractor part of the UI.

leonardmq:
All tests are passing, ready for review 

(CI broken for unrelated reasons)

app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte
```diff
@@ -57,10 +39,11 @@
           body: {
             name: name || null,
             description: description || null,
-            extractor_type: extractor_type as unknown as ExtractorType,
-            output_format: output_format as unknown as OutputFormat,
+            extractor_type: "litellm",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte]: we should remove this from the API. the model_name and provider_name will look up appropriate adapter on the backend. The front end shouldn't be aware of it.

libs/core/kiln_ai/adapters/extractors/base_extractor.py
```diff
@@ -9,7 +9,16 @@
...
 
 
+class ExtractionInput(BaseModel):
+    path: Path | str = Field(description="The absolute path to the file to extract.")
+    mime_type: str = Field(description="The mime type of the file.")
+
+
 class ExtractionOutput(BaseModel):
+    """
+    The output of an extraction. This is the data that will be saved to the database.
```
scosman: nit: data store, not database :)

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -190,6 +195,8 @@ class KilnModelProvider(BaseModel):
     uncensored: bool = False
     suggested_for_uncensored_data_gen: bool = False
     tuned_chat_strategy: ChatStrategy | None = None
+    multimodal_capable: bool = False
+    multimodal_mime_types: List[KilnMimeType] | None = None
```
scosman: This needs to be `List[str]` for backwards compatibility with the remote config server. When we add a new type, the old clients should ignore it, not crash when we try to parse it.

Make it ignore unknown types downstream if needed, but don't use an enum here.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -2554,6 +2748,17 @@ def get_model_by_name(name: ModelName) -> KilnModel:
     raise ValueError(f"Model {name} not found in the list of built-in models")
 
 
+def built_in_models_from_provider(
```
scosman, line 2751: How did we not have this???

scosman (changes requested):
Partial review.

libs/core/kiln_ai/adapters/ml_model_list.py
scosman: A good test to add is one for forwards compatibility. See remote_config tests, add a new one where we set a unknown mime type and make sure it's fine.

scosman: Note: we can keep the enum for the front end and API (will actually be very helpful there). Just not in the ml_model_list.

libs/server/kiln_server/document_api.py
```diff
@@ -134,6 +136,42 @@ class CreateExtractorConfigRequest(BaseModel):
         default_factory=dict,
     )
 
+    @model_validator(mode="after")
+    def validate_properties(self):
+        if self.extractor_type == ExtractorType.LITELLM:
```
scosman: ditto

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+    ],
+    Kind.AUDIO: [
+        "audio/wav",
+        "audio/mpeg",  # this is the official MP3 mimetype, audio/mp3 is often used but not correct
+        "audio/ogg",
+    ],
+}
+
+
+def encode_file(path: Path, mime_type: str) -> dict[str, Any]:
```
scosman: maybe encode_file_litellm_format? or a comment on why this is what it is. I assume LiteLLM format but function name/code don't tell me that?

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+        if model_provider_name is None:
+            raise ValueError(
+                "properties.model_provider_name is required for LitellmExtractor"
+            )
+
+        model_name = extractor_config.model_name()
+        if model_name is None:
+            raise ValueError("properties.model_name is required for LitellmExtractor")
+
+        # TODO: some models supports only a few kinds - maybe drop Kind altogether
```
scosman: Let's make it a point to resolve all TODOs before merging. I'll add a CI system to check this soon (just need to resolve some I added in the first months).

Forces us to make a decision on them, and live with decision. Can leave comments explaining why when it's helpful, but "TODO"s should be done before merge. TODO gives a false sense of "we'll come back and fix this so leaving it suboptimal is okay" which is a trap 😀 

I don't feel strongly on this one, except to remove the TODO and make a final call.

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+        if kind is None:
+            raise ValueError(
+                f"Unsupported MIME type: {extraction_input.mime_type} for {extraction_input.path}"
+            )
+
+        prompt = self.prompt_for_kind.get(kind)
+        if prompt is None:
+            raise ValueError(f"No prompt found for kind: {kind}")
+
+        # FIXME: check modality support - maybe later
```
scosman: Ditto: fix before merging/CR or remove FIXME comment. If we want to merge, we should be happy with current state.

Make this an issue if you want to track it, but code isn't the place.

However, feels like we're using ml_model_list for this function, so can just drop comment?

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+                    "role": "user",
+                    "content": [
+                        {"type": "text", "text": prompt},
+                        encode_file(
+                            Path(extraction_input.path), extraction_input.mime_type
+                        ),
+                    ],
+                }
+            ],
+            api_key=Config.shared().gemini_api_key,
```
scosman: CodeRabbit is right here! 

Need to get the right key. See how we do this in litellm_adapter, extract logic to a shared place if needed.

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
leonardmq: This is the format that LiteLLM takes - as for why it takes that format or distinguished between `image` and such, I don't know. Officially their docs supports PDFs with the `file_url` format, and images with the `image_url` format. In practice, image works with both formats; and not only vision models allow images to come through, and the format documented for PDFs takes in everything as well.

Renamed and added a couple of comments in: https://github.com/Kiln-AI/Kiln/pull/390/commits/441ae91743dea8d0b208acedd0b283316e80f534

libs/core/kiln_ai/adapters/extractors/base_extractor.py
leonardmq: Edited in: https://github.com/Kiln-AI/Kiln/pull/390/commits/5c41a4173bfe45e6fffd72118c4f43c9813b099f

app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte]: Removed in: https://github.com/Kiln-AI/Kiln/pull/390/commits/814860cfdc118ab95144857dd9eca159a1d9121f 

(and rebuilt SDK)

libs/server/kiln_server/document_api.py
leonardmq: Removed in: https://github.com/Kiln-AI/Kiln/pull/390/commits/814860cfdc118ab95144857dd9eca159a1d9121f 

(and rebuilt SDK)

libs/core/kiln_ai/adapters/ml_model_list.py
leonardmq: Updated in: https://github.com/Kiln-AI/Kiln/pull/390/commits/b3aabd2871068290639ac06481d9cb3365026feb

1. Changed to `List[str]`
2. Added a check to skip mimetypes that are not in the current `KilnMimeTypes` enum to ensure forward compat during remote config deserialization
3. Added forward compat test

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
leonardmq: Removed in: https://github.com/Kiln-AI/Kiln/pull/390/commits/8b61070f2c944d23effc6a3db9b0efe66c0b3171

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
leonardmq: This one was actually addressed in this PR - forgot to remove it. 

Removed in: https://github.com/Kiln-AI/Kiln/pull/390/commits/8b61070f2c944d23effc6a3db9b0efe66c0b3171

Some of those are useful because there are interdependencies between the PRs; for example, there are a handful of `FIXME` around the `NAME_FIELD`, where I have a fix pending (the `FilenameString` that also allows unicode in filenames); quite helpful when rebasing. But will open Issues instead going forward.

libs/core/kiln_ai/adapters/remote_config.py
```diff
@@ -43,6 +44,18 @@ def deserialize_config(path: str | Path) -> KilnRemoteConfig:
...
+            # Skip the mimetypes that are not supported - this ensure old clients do not
+            # break when new mime types are added over the air.
+            multimodal_mime_types = provider.get("multimodal_mime_types")
+            if multimodal_mime_types is not None:
+                provider["multimodal_mime_types"] = [
+                    mime_type
+                    for mime_type in multimodal_mime_types
+                    if mime_type in KilnMimeType
+                ]
+
```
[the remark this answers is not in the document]
leonardmq: How ironic is it on a scale from 1 to 10 that this forward compatibility fix broke backward compatibility?

app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte]: Currently, `model_provider_name` and `model_name` are in the `properties` dict, because it was assumed that future extractors might not be models (e.g. `docling`).

But if we force every extractor into `model_name` / `model_provider_name` (even if it is not a model with a provider), then should we also move `model_name` and `model_provider_name` out of `properties` dict and as normal typed fields on the `ExtractorConfig`?

https://github.com/Kiln-AI/Kiln/blob/70b1415c387041503d5f79a8d3c2c8727ac50193/libs/core/kiln_ai/datamodel/extraction.py#L156

@scosman

libs/core/kiln_ai/adapters/ml_model_list.py
leonardmq: And I broke support for Python 3.11 in my forward compatibility fix; which Coderabbit found to be "deliciously ironic": https://github.com/Kiln-AI/Kiln/pull/390#discussion_r2223896478

libs/core/kiln_ai/adapters/remote_config.py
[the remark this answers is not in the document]
scosman: 👏

app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/extractors/[project_id]/create_extractor/+page.svelte]: Refactored in: https://github.com/Kiln-AI/Kiln/pull/390/commits/e053195bd1de933ff61a272ce89d85a01e0e1b0c

Now `model_name` and `model_provider_name` are normal typed properties in the `ExtractorConfig` instead of being in the `properties` dict (also had to update the UI to align with this - mostly just changing API calls and a few references to `xxx.properties["model_name"]` and the likes)

libs/core/kiln_ai/adapters/test_adapter_registry.py
```diff
@@ -232,3 +270,420 @@ async def test_fine_tune_provider(mock_config, basic_task, mock_finetune_from_id
...
+        patch("kiln_ai.adapters.provider_tools.Config") as provider_tools_mock,
+        patch("kiln_ai.adapters.test_adapter_registry.Config") as test_mock,
+    ):
+        provider_tools_mock.shared.return_value = config_instance
+        test_mock.shared.return_value = config_instance
+
+        yield provider_tools_mock
+
+
+def create_config_declarative(
```
leonardmq, line 445: Regression test

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
leonardmq: Whoops. Fixed that and refactored in: https://github.com/Kiln-AI/Kiln/pull/390/commits/bfa2afdd0db36b87988cd9e4518b491f67acbfb6

Extracting the logic out was a little risky, so added a regression test to check that the the `adapter_for_task` produces the same adapters as the original switch

libs/core/kiln_ai/adapters/provider_tools.py
```diff
@@ -443,3 +440,77 @@ class ModelProviderWarning:
...
+                "vertex_location": Config.shared().vertex_location,
+            }
+        case ModelProviderName.together_ai:
+            return {"api_key": Config.shared().together_api_key}
+        case ModelProviderName.azure_openai:
+            return {
+                "api_key": Config.shared().azure_openai_api_key,
+                "api_version": "2025-02-01-preview",
+                "base_url": Config.shared().azure_openai_endpoint,
+            }
```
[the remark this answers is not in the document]
leonardmq: The shared config is typically for user-facing customizable values. We currently do not expose this, and the API version is quite tied to the rest of the code (a different version would likely break other modules), so in my opinion it is fine to keep it hardcoded as part of the provider auth

libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
leonardmq: Then renamed that into `provider connection details` instead of `provider auth` (since it is more than just auth), in: https://github.com/Kiln-AI/Kiln/pull/390/commits/96acc6e3b492fb850fa77353ddd5e123b51eb001

scosman:
Feedback around refactor

- get_provider_auth_details needs some work. We need something more type safe.
  - name: 
    - not "auth" related, doesn't mention lite_llm which it very much is coupled to.
    - maybe llm_core_config_for_provider?
  - make the return type typed. New LiteLlmCoreConfig with, default_headers, base_url, additional_body_options (or None). 
  - return None if it’s not a supported liteLLM provider instead of raising an error (ModelProviderName.kiln_fine_tune, ModelProviderName.kiln_custom_registry, etc). The caller can raise the error if it’s None (those two strings aren’t particularly good)
  - Keep the raise_exhaustive check. I want to force people to handle all cases (you have this, but just don’t remove it)
  - caller’s code is now more clean: it will now only create a LiteLllm adapter if llm_core_config_for_provider is not None. We might add others later so this is good. Can raise if it’s None (replacing two prior raises in the switch)
- lite_llm_config_for_openai_compatible
  - most of this code should be moved into your new llm_core_config_for_provider (all the code for api_key and base_url), in the switch statement for openai_compatible.
  - updated_run_config_properties part is specific to LiteLLMAdapter and should be moved out of the provider_tools since it isn’t general, and can put that where you have “ openai compatible providers are handled separately since they have custom config logic” comment
- General CR:
  - If you don’t expect the branch `ModelProviderName.openai_compatible` to ever be called, raise an error instead of returning empty. Will help catch errors much sooner. However, we are going to populate this branch so moot point here.

leonardmq:
@scosman - implemented all the changes, but I had to update a lot of tests, some of which were quite unrelated (mostly needed updating due to some test-specific mocking), so need to run coverage to check everything touched is covered

libs/core/kiln_ai/adapters/adapter_registry.py
```diff
@@ -9,191 +7,56 @@
...
+        raise ValueError(
+            "Fine tune or custom openai compatible provider is not a core provider. The underlying provider should be used when requesting the adapter litellm config instead."
+        )
+
+    return LiteLlmConfig(
+        run_config_properties=run_config_properties,
+        base_url=config.base_url,
+        default_headers=config.default_headers,
+        additional_body_options=config.additional_body_options or {},
+    )
```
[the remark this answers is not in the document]
leonardmq: 1. I think the point with the `llm_core_config_for_provider` suggestion was to have something more tied to `litellm`, so `lite_llm_core_config_for_provider` should be fine (and maybe better / more specific) too as far as I understand
2. We return `LiteLlmCoreConfig` in `lite_llm_core_config_for_provider`; but we return `LiteLlmConfig` in `litellm_core_provider_config` (that last part is unchanged), which is a higher level / different construct concerned with the run properties (and pieces from the `LiteLlmCoreConfig`)
3. I think the feedback was to return `None` from inside the `switch` in `lite_llm_core_config_for_provider`, and raise an error at the caller (which is here inside `litellm_core_provider_config`)
4. The feedback on `lite_llm_config_for_openai_compatible` is saying to move code out of there, which has been done (now moved into the `litellm_core_provider_config` and into `lite_llm_core_config_for_provider`)

scosman (approved):
Whoo!

PR #402, opened 2025-07-03
feat: chunk embedding

leonardmq:
## What does this PR do?

Changes:
- add datamodel: 
  - `EmbeddingConfig`: the configuration managing the embedding process (e.g. what model, etc.)
  - `ChunkEmbeddings`: the container that we will store the embeddings into (as JSON), where `ChunkEmbedding.embeddings[i]` will be the embedding that corresponds to `ChunkedDocument.chunks[i]`
- add `BaseEmbeddingAdapter` and basic LiteLLM implementation of the adapter (`LitellmEmbeddingAdapter`)
- add `built_in_embedding_models` list that describes the different available models, providers, and their features (e.g. OpenAI's `text-embedding-3` family supports custom number of dimensions) and limitations (max input tokens, etc.), similar to `ml_model_list`
- add API endpoints to retrieve embedding models

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Added support for embedding models, including new API endpoints to list and retrieve available embedding models and their metadata.
  * Introduced data models and adapters for embedding configurations, embedding vectors, and chunk-level embeddings.
  * Enabled integration of embedding models from providers such as OpenAI and Gemini, with support for custom dimensions and model metadata.
  * Extended remote configuration to include embedding models alongside standard models.
  * Exposed embedding model list as part of the public adapters API.

* **Bug Fixes**
  * Removed unused imports in test modules to improve code cleanliness.

* **Tests**
  * Added comprehensive unit and integration tests for embedding adapters, embedding data models, and embedding model management.
  * Introduced tests for new API endpoints related to embedding models.
  * Extended remote configuration tests to cover embedding models.

* **Documentation**
  * Updated API schema definitions to include embedding model endpoints and related data structures.

* **Refactor**
  * Improved provider resolution logic for embedding and language models to streamline model ID handling.
  * Refactored remote config handling to unify model and embedding model loading and serialization.


leonardmq:
This is WIP because I need to double check some things, add a couple of `paid` tests as well, but the overall shape is roughly done

leonardmq:
@coderabbitai review

libs/core/kiln_ai/adapters/embedding/embedding_model_list.py
```diff
@@ -0,0 +1,96 @@
...
+        friendly_name="text-embedding-004",
+        providers=[
+            KilnEmbeddingModelProvider(
+                name=ModelProviderName.openai,
+                model_id="text-embedding-004",
+                supported_dimensions=[768],
+                max_input_tokens=2048,
+            ),
+        ],
+    ),
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/402/commits/6a3cb058c1b36e1a83d90ca855a8eb2a13cef2ae

libs/core/kiln_ai/adapters/embedding/embedding_model_list.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class EmbeddingModelName(str, Enum):
+    """
+    Enumeration of specific model versions supported by the system.
+    """
+
+    # Embedding models
+    openai_text_embedding_3 = "openai_text_embedding_3_small"
+    gemini_text_embedding_004 = "gemini_text_embedding_004"
```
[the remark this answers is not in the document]
leonardmq: Fixed in https://github.com/Kiln-AI/Kiln/pull/402/commits/6a3cb058c1b36e1a83d90ca855a8eb2a13cef2ae

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,125 @@
...
+    Enumeration of supported embedding model families.
+    """
+
+    # for bespoke proprietary models, the family tends to be the same
+    # as provider name, but it does not have to be
+    openai = "openai"
+    gemini = "gemini"
+
+
+class EmbeddingModelName(str, Enum):
```
leonardmq, line 20: These names are awkward, but to keep things aligned with the `ml_model_list`, the model names should be globally unique

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,125 @@
...
+    model_id: str | None = None
+
+    max_input_tokens: int | None = None
+
+    n_dimensions: int | None = Field(
+        default=None,
+        description="The number of dimensions in the output embedding. If the model allows picking between different options, provide the list of available dimensions.",
+    )
+
+    supports_custom_dimensions: bool = Field(
```
leonardmq, line 49: OpenAI’s 3rd-generation embedding models support shortening the embedding dimensionality. 

More info:
- https://platform.openai.com/docs/guides/embeddings#:~:text=embeddings%20v3%20launch%20blog%20post
- about the technique itself: https://arxiv.org/abs/2205.13147

Our `KilnEmbeddingModelProvider` setup supports this, though we don’t need to expose it in the UI yet. The default dimensions on these OpenAI models are fairly high (`1536` for small, `3072` for large), so it is likely we will want to shorten them later - indexing high-dimensional vectors is usually not ideal in production (memory, speed)

leonardmq:
@coderabbitai review

app/desktop/studio_server/provider_api.py
```diff
@@ -182,6 +204,58 @@ async def get_available_models() -> List[AvailableModels]:
...
+                        available_models.models.append(
+                            EmbeddingModelDetails(
+                                id=model.name,
+                                name=model.friendly_name,
+                                n_dimensions=provider.n_dimensions or 0,
+                                max_input_tokens=provider.max_input_tokens or 0,
+                                supports_custom_dimensions=provider.supports_custom_dimensions,
+                            )
+                        )
+        return models
```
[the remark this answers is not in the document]
leonardmq: Agreed but would rather do that later in a targeted refactor though to try to avoid potential merge conflicts as this PR is getting merged into a branch that will stay parallel to `main` for a good amount of time still (and the specifics of which might still zigzag for a while)

app/desktop/studio_server/provider_api.py
```diff
@@ -182,6 +204,58 @@
...
+                        (m for m in models if m.provider_id == provider.name), None
+                    )
+                    if available_models:
+                        available_models.models.append(
+                            EmbeddingModelDetails(
+                                id=model.name,
+                                name=model.friendly_name,
+                                n_dimensions=provider.n_dimensions or 0,
+                                max_input_tokens=provider.max_input_tokens or 0,
+                                supports_custom_dimensions=provider.supports_custom_dimensions,
```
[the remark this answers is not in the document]
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/402/commits/af9b636de90a72ef3fe34f6fbc06e75d4e5f3430

I made `n_dimensions` and `max_input_tokens` required fields, so the check / fallback to `0` that you flagged no longer needs to happen

app/desktop/studio_server/provider_api.py
```diff
@@ -182,6 +204,58 @@ async def get_available_models() -> List[AvailableModels]:
...
+    @app.get("/api/providers/embedding_models")
+    async def get_providers_embedding_models() -> ProviderEmbeddingModels:
+        models = {}
+        for model in built_in_embedding_models:
+            models[model.name] = ProviderModel(id=model.name, name=model.friendly_name)
+        return ProviderEmbeddingModels(models=models)
+
+    # returns map, of provider name to list of model names
+    @app.get("/api/available_embedding_models")
+    async def get_available_embedding_models() -> List[AvailableEmbeddingModels]:
```
leonardmq: The logic in this endpoint is copy / paste from the `@app.get("/api/available_models")` endpoint - if we want to refactor (probably should), I'd rather leave that for later after merging so we don't risk the `rag` branch touching too much base code that might change before `rag -> main` merging

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,128 @@
+from enum import Enum
```
scosman, line 1: You should merge main, then integrate this with `remote_config.py`. Should only be a few lines, but will make it so we can remotely update it OTA.

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class KilnEmbeddingModelProvider(BaseModel):
+    name: ModelProviderName
+
+    model_id: str = Field(
+        description="The model ID for the embedding model. This is the ID used to identify the model in the provider's API.",
+    )
+
+    max_input_tokens: int = Field(
```
scosman: Make this optional for future proofing? These data structured will be updated over the air so we have to think ahead about possible values, not just current values. If we ever want to support optional for any future model, we should be optional now, or when we update it, it would break old clients.

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,128 @@
...
+        description="The maximum number of tokens that can be input to the model. If the model allows picking between different options, provide the list of available options.",
```
scosman: "If the model allows picking between different options, provide the list of available options." doesn't make sense here?

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    n_dimensions: int = Field(
+        description="The number of dimensions in the output embedding. If the model allows picking between different options, provide the list of available dimensions.",
```
scosman: you can't provide a list of dimensions in an int.

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+built_in_embedding_models: List[KilnEmbeddingModel] = [
+    KilnEmbeddingModel(
+        family=KilnEmbeddingModelFamily.openai,
+        name=EmbeddingModelName.openai_text_embedding_3_small,
+        friendly_name="text-embedding-3-small",
+        providers=[
+            KilnEmbeddingModelProvider(
+                # TODO: wondering if should use separate enum for the ModelProviderName,
```
scosman: yeah, I'd delete comment. Shared is good?!

app/desktop/studio_server/provider_api.py
```diff
@@ -98,6 +102,20 @@ class AvailableModels(BaseModel):
...
 
+class EmbeddingModelDetails(BaseModel):
+    id: str
+    name: str
+    n_dimensions: int
+    max_input_tokens: int
+    supports_custom_dimensions: bool
+
+
+class AvailableEmbeddingModels(BaseModel):
```
scosman: nit: naming to EmbeddingProvider since this object is a provider (it contains models)? Then the above one can just be `EmbeddingModel`. 

I know this is a copy of my bad names, but no reason to copy to bad parts 😆

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: I think so as well - was mostly wondering because we may have embedding-only providers at some point, but probably not much of a problem even if that happens

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Good catch - I changed halfway through. I originally wanted to provide multiple options for dimensionality for the OpenAI embeddings that can be shortened, but ended up just adding optional custom number of dimensions in properties (and we could do the multiple option some level up when we want to)

app/desktop/studio_server/provider_api.py
scosman: copy/pasted shouldn't cause conflicts (which isn't the case for editing).

General note: following existing style is good, but some of this code was my first ever python code, so always feel free to up the quality bar when touching areas.

app/desktop/studio_server/provider_api.py
```diff
@@ -182,6 +204,58 @@ async def get_available_models() -> List[AvailableModels]:
...
+                provider_name=provider_name_from_id(provider),
+                provider_id=provider,
+                models=[],
+            )
+            for provider in key_providers
+        ]
+
+        for model in built_in_embedding_models:
+            for provider in model.providers:
+                if not provider.model_id:
```
scosman: I'd remove this, nothing to do with embedding models.

libs/core/kiln_ai/adapters/embedding/base_embedding_adapter.py
```diff
@@ -0,0 +1,46 @@
...
+
+from litellm import Usage
+from pydantic import BaseModel, Field
+
+from kiln_ai.datamodel.embedding import EmbeddingConfig
+
+logger = logging.getLogger(__name__)
+
+
+class GeneratedEmbedding(BaseModel):
```
scosman: Just `embedding`?

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+        if dimensions is not None:
+            if not isinstance(dimensions, int) or dimensions <= 0:
+                raise ValueError("Dimensions must be a positive integer")
+
+        return EmbeddingOptions(
+            dimensions=dimensions,
+        )
+
+    # TODO: refactor this to be shared with other implementations of LiteLLM adapters
+    # for example, embedding adapter for LiteLLM, and also Extractor adapter for LiteLLM
```
scosman: This refactor should be part of this PR, not saved for later. No copy/paste within the repo (unless we're editing substantially after the paste)

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -285,6 +285,8 @@ def build_extra_body(self, provider: KilnModelProvider) -> dict[str, Any]:
 
         return extra_body
 
+    # TODO: refactor this to be shared with other implementations of LiteLLM adapters
+    # for example, embedding adapter for LiteLLM, and also Extractor adapter for LiteLLM
```
scosman: Ditto to other comment: this should be part of this PR

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+from kiln_ai.datamodel.embedding import EmbeddingConfig
+from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
+
+MAX_BATCH_SIZE = 2048
+
+
+class EmbeddingOptions(BaseModel):
+    dimensions: int | None = Field(
+        default=None,
+        description="Some models support requesting vectors of different dimensions.",
```
scosman: "The number of dimensions to return for embeddings. Some models support requesting vectors of different dimensions.",

libs/core/kiln_ai/adapters/embedding/base_embedding_adapter.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Base class for all embedding adapters.
+
+    Should be subclassed by each embedding adapter.
+    """
+
+    def __init__(self, embedding_config: EmbeddingConfig):
+        self.embedding_config = embedding_config
+
+    async def embed(self, text: List[str]) -> EmbeddingResult:
```
scosman: naming here is confusing. Strongly implies singular for an array use case. Maybe `generate_embeddings` and `input_texts`? Ditto below for private method

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+        if model_provider_name is None:
+            raise ValueError("Provider must be set")
+
+        model_name = embedding_config.model_name
+        if model_name is None:
+            raise ValueError("Model name must be set")
+
+        super().__init__(embedding_config)
+        self.model_provider_name = model_provider_name
+        self.model_name = model_name
```
scosman: remove these 3, and just access `self.embedding_config.model_provider_name` and similar? Don't want to create duplicate copies, and Base already saves embedding_config

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+        self.model_provider = built_in_embedding_models_from_provider(
+            self.model_provider_name, self.model_name
+        )
+        if self.model_provider is None:
+            raise ValueError(
+                f"Embedding model {self.model_name} not found in the list of built-in models"
+            )
+
+    async def _embed(self, text: List[str]) -> EmbeddingResult:
+        # TODO: providers will throw if the text input is too long - goes over the max tokens for the model
```
scosman: +1 but let's remove the TODO? more likely we'll forget this here forever impacting the code than find and fix it.

Maybe just iterate the texts, check length, and throw a useful error message. Callers should handle it, but that's their job, don't need TODO here.

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+            **self.build_options().model_dump(exclude_none=True),
+        )
+
+        # sanity check to ensure integrity as we should always have as many embeddings
+        # as inputs
+        if len(response.data) != len(text):
+            raise ValueError("Response data length does not match input text length")
+
+        embeddings = []
+        for item in sorted(response.data, key=lambda x: x.get("index")):
```
scosman: 1) naming: avoid names like "item" in favour of descriptive names

2) is typing not possible here? This feels fragile. Would prefer something like below where type-checker would find issues when we upgrade litellm. See example from the LiteLLM Adapter. Very pseudocode, guessing names. 

``` 
from litellm.types.utils import EmbeddingResponse # just guessing the name, but ModelResponse is what it is for LLM calls

...

if (
                not isinstance(response, EmbeddingResponse)
                or [any other checks like embedding isn't null or index isn't null]
):
raise RuntimeError(
    f"Expected ModelResponse with Choices, got {type(response)}."
)

for item in sorted(response.data, key=lambda x: x.inded):
  embeddings.append(GeneratedEmbedding(vector=item/embedding))
```

libs/core/kiln_ai/datamodel/embedding.py
```diff
@@ -0,0 +1,86 @@
...
+
+if TYPE_CHECKING:
+    from kiln_ai.datamodel.chunk import ChunkedDocument
+    from kiln_ai.datamodel.project import Project
+
+
+class EmbeddingConfig(KilnParentedModel):
+    name: str = NAME_FIELD
+    description: str | None = Field(
+        default=None, description="The description of the embedding config"
```
scosman: Names should be descriptive more than just the field names. "A description for your reference, not shared with embedding models."

libs/core/kiln_ai/datamodel/embedding.py
```diff
@@ -0,0 +1,86 @@
...
+    )
+    model_provider_name: ModelProviderName = Field(
+        description="The provider to use to generate embeddings.",
+    )
+    # TODO: should model_name be the EmbeddingModelName enum instead of a string?
+    # in the TaskRunConfigProperties, we store model_name as a plain string:
+    # https://github.com/Kiln-AI/Kiln/blob/b92dde56d9259aa47ba4f71a820f90138bd86c6e/libs/core/kiln_ai/datamodel/task.py#L55
+    # maybe for backward compatibility when we deprecate old models?
```
scosman: yeah, general string needed for backwards compatibility (and remote update). Can delete comment.

libs/core/kiln_ai/datamodel/embedding.py
```diff
@@ -0,0 +1,86 @@
...
+        if self.parent is None or self.parent.__class__.__name__ != "Project":
+            return None
+        return self.parent  # type: ignore
+
+    @model_validator(mode="after")
+    def validate_properties(self):
+        # FIXME: not ideal to import here, but if we import normally, we get a circular import
+        # We should probably move the ml model lists out of the adapters package, since they are
+        # base constructs with no dependencies (and we seem to be getting a circular import due
+        # to the __init__.py rather than due to intrinsic circularity)
```
scosman: Remove this. The model list is dynamic. We'll deprecate them. We'll update it remotely. Changing it should never invalidate an existing model on disk. Data models are forever (get persisted to disk).

I think I saw you had validation on creating a new one via API, which is the right time to check. we don't need a model validator as well.

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
leonardmq: I wanted to do the check here, but then realized that each provider will have its own way of counting tokens, and I could not find a unified way to do it in LiteLLM, so more involved than checking the string length. 

Removed the comment in: https://github.com/Kiln-AI/Kiln/pull/402/commits/90855178064343216100403d90837c231a88390f

Then the best we can probably do here is surface the errors that LiteLLM returns, so the user knows what is going on. The concern though is having a pipeline held up due to just a couple of problematic chunks that for some reason would be too long - hopefully won't happen

libs/core/kiln_ai/datamodel/embedding.py
leonardmq: Removed comment in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

libs/core/kiln_ai/datamodel/embedding.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

I tend to forget that we expose these

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Removed the comment in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Removed the outdated part in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Removed irrelevant part in Removed the outdated part in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
leonardmq: Rewritten in https://github.com/Kiln-AI/Kiln/pull/402/commits/ac271d0336020682083a5e284c1dd2b63e824b28

libs/core/kiln_ai/adapters/embedding/base_embedding_adapter.py
leonardmq: Renamed `embed` into `generate_embeddings`; `_embed` into `_generate_embeddings`; and `text` into `input_texts` in https://github.com/Kiln-AI/Kiln/pull/402/commits/659d9037cf58cce8086fe4ae3f911462e2148f6b

libs/core/kiln_ai/adapters/embedding/base_embedding_adapter.py
leonardmq: Renamed into `Embedding` in https://github.com/Kiln-AI/Kiln/pull/402/commits/3a015d873e8e025a386bfe117f97982fb40017de

Was trying to make it more different than the `Embedding` defined in the datamodel to avoid it risking showing up in autocomplete

libs/core/kiln_ai/datamodel/embedding.py
leonardmq: Removed in: https://github.com/Kiln-AI/Kiln/pull/402/commits/85eb5421476a0f93460a3990366e3a23a1bc8dad

Kept the `model_validator` to verify at least that `dimensions`, if provided, is a positive `int`; or is there a reason to get rid of that check too?

The risk I can see is if `dimensions` were to evolve into meaning something else in the future, but that seems unlikely enough and conceptually different enough that it would warrant a refactor at that time

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
leonardmq: Refactored in: https://github.com/Kiln-AI/Kiln/pull/402/commits/85eb5421476a0f93460a3990366e3a23a1bc8dad

I have put the method in `utils/litellm.py`, not sure if there is a more suitable place.

The refactored version does the mapping from Kiln model to LiteLLM model, and tells the caller whether it is custom or not (different adapters handle that differently; the embedding adapter for example does not support custom models at all; the completion adapter checks that a custom API base URL is set).

Also used `@cached_property` on the caller to avoid needing the private variable to keep track of whether it was already computed / cached.

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
leonardmq: Replied on the other comment: https://github.com/Kiln-AI/Kiln/pull/402#discussion_r2199770741

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
leonardmq: Removed these in: https://github.com/Kiln-AI/Kiln/pull/402/commits/85eb5421476a0f93460a3990366e3a23a1bc8dad and also moved `model_provider` resolution to a `@cached_property` instead of in the constructor

app/desktop/studio_server/provider_api.py
leonardmq: Removed in https://github.com/Kiln-AI/Kiln/pull/402/commits/2fc43d71713caa4c0e8318ae644f584aa24cc8b8

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
leonardmq: The SDK does have an `EmbeddingResponse` type and that is what we get back from the SDK call. However, that type just has `data` as an untyped `List`, so what we get is dictionaries in `response.data`. Confusingly enough, the SDK has an `Embedding` type (also in `litellm.types.utils`), but we don't get that in the response.

I have added a validation / mapping step in https://github.com/Kiln-AI/Kiln/pull/402/commits/9623f35428da1f4c3de45db2349c1f40327698c0

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Made it optional in https://github.com/Kiln-AI/Kiln/pull/402/commits/a141feea84ca091314d6436cd3b4af904f103436

app/desktop/studio_server/provider_api.py
leonardmq: Renamed in https://github.com/Kiln-AI/Kiln/pull/402/commits/a141feea84ca091314d6436cd3b4af904f103436

I have not internalized the naming related to the models - `model_name`, `model_id`, `model_provider.name`, `ModelProviderName`, `KilnModel`, `KilnModelProvider`, etc. Always takes me a while to figure out what model construct / enum / subset we are dealing with, so tried to avoid changes there

leonardmq:
Rebased to pull `main` into here via `rag`

libs/core/kiln_ai/adapters/test_remote_config.py
```diff
@@ -16,68 +17,91 @@
...
+        m.model_dump(mode="json") for m in built_in_embedding_models
+    ]
 
 
 @pytest.mark.asyncio
 async def test_load_remote_models_success(monkeypatch):
-    del os.environ["KILN_SKIP_REMOTE_MODEL_LIST"]
-    original = built_in_models.copy()
-    sample_models = [built_in_models[0]]
+    monkeypatch.delenv("KILN_SKIP_REMOTE_MODEL_LIST", raising=False)
```
leonardmq, line 65: Doing `monkeypatch.delenv` is preferable because it undoes itself after the test execution whereas `os.environ` may pollute other execution by side effect

libs/core/kiln_ai/adapters/test_remote_config.py
```diff
@@ -16,68 +17,91 @@
...
 
     load_remote_models("http://example.com/models.json")
     await asyncio.sleep(0.01)
     assert built_in_models == sample_models
-    built_in_models[:] = original
+    assert built_in_embedding_models == sample_embedding_models
 
 
 @pytest.mark.asyncio
 async def test_load_remote_models_failure(monkeypatch):
```
leonardmq, line 86: The error simulated in this test was never actually hit due to `KILN_SKIP_REMOTE_MODEL_LIST == "true"` resolving to true during execution - found this out when adding the check for logger call

libs/core/kiln_ai/adapters/test_remote_config.py
```diff
@@ -16,68 +17,91 @@
...
+            model_list=sample_models,
+            embedding_model_list=sample_embedding_models,
+        )
 
     monkeypatch.setattr("kiln_ai.adapters.remote_config.load_from_url", fake_fetch)
 
     load_remote_models("http://example.com/models.json")
     await asyncio.sleep(0.01)
     assert built_in_models == sample_models
-    built_in_models[:] = original
```
leonardmq, line 66: I removed this, but am not sure what the intent was here (I get why the assignment in the actual code, but not why we are doing that at this end of this test)

libs/core/kiln_ai/adapters/remote_config.py
```diff
@@ -3,40 +3,76 @@
...
+        "model_list": [m.model_dump(mode="json") for m in models],
+        "embedding_model_list": [m.model_dump(mode="json") for m in embedding_models],
+    }
     Path(path).write_text(json.dumps(data, indent=2, sort_keys=True))
 
 
-def deserialize_config(path: str | Path) -> List[KilnModel]:
+def deserialize_config(path: str | Path) -> KilnRemoteConfig:
     raw = json.loads(Path(path).read_text())
-    model_data = raw.get("model_list", raw if isinstance(raw, list) else [])
```
leonardmq, line 23: There was a fallback check trying to use the JSON as a list of models, but seems like that was a transient thing during development and now there is no scenario where `raw` is a list of models (as opposed to a dict containing the list of models)

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: Integrated with `remote_config.py` - worth double-checking as I get the basic flow, but am not fully familiar with all the interactions there.

Also found some oddities in the tests (unintentional-looking env var states in the failure test case, floating assignment in one test) and in the code code (legacy / dev fallback logic). I left comments on the points I was unsure about in this commit: 
- https://github.com/Kiln-AI/Kiln/pull/402/commits/15e07da2c358cfde33c807d2e0da0696dcd38d7f

libs/core/kiln_ai/adapters/embedding/litellm_embedding_adapter.py
```diff
@@ -0,0 +1,138 @@
...
+            embeddings.append(Embedding(vector=item.get("embedding")))
+
+        return EmbeddingResult(embeddings=embeddings, usage=response.usage)
+
+    def build_options(self) -> EmbeddingOptions:
+        dimensions = self.properties.get("dimensions", None)
+        if dimensions is not None:
+            if not isinstance(dimensions, int) or dimensions <= 0:
+                raise ValueError("Dimensions must be a positive integer")
```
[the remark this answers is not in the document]
leonardmq: We probably could skip this one, but probably does not hurt to do a quick sanity check there?

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
scosman: I'm not seeing the comments - was this all resolved? Code looks fine

app/desktop/studio_server/provider_api.py
scosman: Yeah, it's confusing that 1 model (GPT 4o) can be on many modelProviders (OpenRouter, OpenAI, etc)

libs/core/kiln_ai/adapters/ml_embedding_model_list.py
leonardmq: They are here: 
1. https://github.com/Kiln-AI/Kiln/pull/402#discussion_r2201125330
2. https://github.com/Kiln-AI/Kiln/pull/402#discussion_r2201127989
3. https://github.com/Kiln-AI/Kiln/pull/402#discussion_r2201132101
4. https://github.com/Kiln-AI/Kiln/pull/402#discussion_r2201136117

libs/core/kiln_ai/datamodel/embedding.py
scosman: can keep the type check! Just kill the value checks based on remote config as you did.

scosman (approved):

PR #412, opened 2025-07-09
update prompt so it doesn't confuse examples of the issues as examples of data gen output

scosman:
## Summary by CodeRabbit

* **New Features**
  * Added clarifying notes in guidance templates to specify that provided examples are for understanding the issue and not for data generation.

* **Style**
  * Improved the wording of example sections to enhance clarity for users.

### DIFF
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts
@@ -370,9 +370,7 @@ Here are two examples of topics generated from an example task/system prompt and
       - "Messages including 'help me understand everything about'"
 `
     } else if (task_type == "inputs") {
-      template += `When generating model inputs, generate inputs that are likely to trigger the issue. This may take some creativity, but it's important to make sure the issue is triggered.
-
-Here are two examples of inputs generated from an example task/system prompt, issue description, and a topic path:
+      template += `Here are two examples of inputs generated from an example task/system prompt, issue description, and a topic path:
 
 ## Example 1
  - Task/System Prompt: "Generate news article headlines from a summary of the article, avoiding clickbait."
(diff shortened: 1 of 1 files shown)


tawnymanticore:
It's definitely better, but still a bit flakey. still looks like some input vs output confusion for the input generation. the generator to make the input prompts seems to make outputs (summaries) about 25% of the time, and makes the correct wall of text as input about 75% of the time.

tawnymanticore:
I do get this warning in the logs, not sure if its related to these changes or the changes before. nothing crashes or anything:

```
INFO:     127.0.0.1:55909 - "POST /api/projects/189194447826/tasks/104775461614/generate_inputs HTTP/1.1" 200 OK
.../Kiln/Kiln/.venv/lib/python3.12/site-packages/pydantic/main.py:426: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue: Expected 9 fields but got 5 for type `Message` with value `Message(content='{"generated_samples":["Dog advent...ific_fields={'refusal': None, 'reasoning': None})` - serialized value may not be as expected.
  PydanticSerializationUnexpectedValue: Expected `StreamingChoices` but got `Choices` with value `Choices(finish_reason='st...finish_reason': 'stop'})` - serialized value may not be as expected
```

leonardmq:
I ran into this problem over the weekend on `main`, where instead of creating one sample with input text `car, plane, bicycle`, it would create 3 samples (one for `car`, one for `plane` and one for `bicycle`). I just tried this branch and it is better. 

It still happens; but less frequently. Whereas before, it would do it for every sample in my task, now it is doing the duplication for ~5-10% of samples it generates.

(The model also still mixes up output and input a bit when the prompt is complex; but that is not exactly the same problem.)

---

My Task is something like:
- prompt (something along these lines): Given a list of words, topics or sentences provided by the user. Generate a list of fill-in-the-blank exercises about the themes.
- input (unstructured): `like, daily life, activities`
- output (structured JSON):
```json
[
   {
      "question":"I like to eat <blank/>.",
      "options":[
         "food",
         "universe",
         "wood",
         "mood"
      ],
      "correct":"food",
      "explanation":"You eat food. The other things cannot be eaten"
   }
]
```

Issue #437, opened 2025-07-16
[Bug] Base Parser should strip trailing / leading whitespace


leonardmq:
**Describe the bug**
Some reasoning models work without R1 parser, but there is awkward whitespace around either the reasoning / parts of the reasoning, or the output / parts of the output. When using the R1 parser, that whitespace goes away, probably because we do strip the strings in the R1 parser.

Solution likely to be to add stripping for reasoning / output in the base parser, instead of returning the object unchanged in the base parser.

**Expected behavior**

**Screenshots**

<img width="1234" height="1008" alt="Image" src="https://github.com/user-attachments/assets/2bbaf8da-d5a9-4886-8577-1d7dc90a11c1" />

**System Information:**

- OS: MacOS Sequoia 15.5
- Browser: Chrome
- Kiln app Version: `main` branch (currently adding new models there)

**Additional context**
Happened with a few models from siliconflow.cn - such as `tencent/Hunyuan-A13B-Instruct`

leonardmq:
We decided against stripping the extra whitespace - to just return whatever the provider is returning. Whitespace stripping when models / providers are doing this can be done on the user side instead.

PR #438, opened 2025-07-16
fix: strip whitespace in base parser

leonardmq:
## What does this PR do?

Some reasoning models produce output with awkward whitespace around either the `reasoning` or the `output`. I noticed that using the R1 parser with those fixes the issue even when R1 parsing is not required - seems to be because we strip whitespace in the R1 parser, but don't in the base parser.

## Related Issues

https://github.com/Kiln-AI/Kiln/issues/437

## Checklists

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved output formatting by automatically removing extra spaces from results and intermediate outputs.

### DIFF
--- a/libs/core/kiln_ai/adapters/parsers/base_parser.py
+++ b/libs/core/kiln_ai/adapters/parsers/base_parser.py
@@ -6,4 +6,15 @@ class BaseParser:
         """
         Method for parsing the output of a model. Typically overridden by subclasses.
         """
+
+        if original_output.output is not None and isinstance(
+            original_output.output, str
+        ):
+            original_output.output = original_output.output.strip()
+
+        if original_output.intermediate_outputs is not None:
+            for key, value in original_output.intermediate_outputs.items():
+                if isinstance(value, str):
+                    original_output.intermediate_outputs[key] = value.strip()
+
         return original_output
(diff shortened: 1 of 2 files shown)


leonardmq:
@coderabbitai review

leonardmq:
May or may not merge - some concerns around possibly manipulating the output more than we need to

scosman (changes requested):
I suggest we don't do this. We should save what the model returns without parsing unless we have to (R1 style). More random magic like this, the more bugs people are going to hit when they try to take Kiln prompts and productionize.

leonardmq:
@scosman 

Agreed and will close the PR, but also thinking this problem applies even more to everything else we already do and not expose to the user.

For productionizing (without using the Kiln library), we should probably expose a code snippet with the following things prefilled:
- LiteLLM (it internally moves params around. and has some internal validation)
- the prompt
- the parameters prefilled:
  - LiteLLM provider name
  - LiteLLM model name; currently hard to know the official provider's model id / name from the UI as we show our own friendly name
  - provider-specific params that we fill in; like `thinking_enabled`, etc.
  - temperature, etc.
- our own parsing code; e.g. JSON parser, R1 parser - in the case of the JSON one we seem to even be extracting JSON within backticks in plain text responses

scosman:
Code snippet will be a bit hard. I'm doing tools now, and there's lots of logic. Probably can swing a code snippet for our library, with inlined run config.

leonardmq:
Closing this one as we won't be doing the whitespace stripping

PR #444, opened 2025-07-17
fix: gemini providers support json_schema

leonardmq:
## What does this PR do?

Update `structured_output_mode` to `json_schema` for Gemini models, as they work properly with it. 

I was unable to test Vertex AI for the `Gemini 1.5 *` models because these models are now deprecated on Vertex and cannot be used by new accounts (started on April 29th 2025: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions#legacy-stable).

Not sure if we can update `structured_output_mode` or if once the model has been set, changing that would risk interfering existing evals that may have been using these models?

### DIFF
--- a/libs/core/kiln_ai/adapters/ml_model_list.py
+++ b/libs/core/kiln_ai/adapters/ml_model_list.py
@@ -810,17 +810,17 @@ built_in_models: List[KilnModel] = [
             KilnModelProvider(
                 name=ModelProviderName.openrouter,
                 model_id="google/gemini-2.0-flash-001",
-                structured_output_mode=StructuredOutputMode.json_instruction_and_object,
+                structured_output_mode=StructuredOutputMode.json_schema,
             ),
             KilnModelProvider(
                 name=ModelProviderName.gemini_api,
                 model_id="gemini-2.0-flash",
-                structured_output_mode=StructuredOutputMode.json_instruction_and_object,
+                structured_output_mode=StructuredOutputMode.json_schema,
             ),
             KilnModelProvider(
                 name=ModelProviderName.vertex,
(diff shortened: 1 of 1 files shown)


libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -810,17 +810,17 @@ class KilnModel(BaseModel):
             KilnModelProvider(
```
leonardmq, line 814: All the changes were tested manually.

Double checked that the `Raw data` in each run had the following - assuming this data being present in the raw data rules out potential cache interference: 
```
output.source.properties.structured_output_mode: "json_schema"
```

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -750,26 +751,28 @@ class KilnModel(BaseModel):
...
+                gemini_reasoning_enabled=True,
+                thinking_level="medium",
             ),
             KilnModelProvider(
                 name=ModelProviderName.gemini_api,
                 model_id="gemini-2.5-pro",
                 structured_output_mode=StructuredOutputMode.json_schema,
                 suggested_for_data_gen=True,
                 suggested_for_evals=True,
-                # TODO: Gemini API doesn't return reasoning here, so we don't ask for it. Strange.
```
leonardmq, line 760: That was due to the models requiring this new gemini-specific param to be set:
```json
{ "reasoning": true }
```

and depending on the provider, also `thinking_level="some-level"`

leonardmq:
Tested all the models with `test_all_built_in_models_structured_output` and working ok

(CI is broken right now though)

PR #448, opened 2025-07-21
docs: update lib docs

leonardmq:
## What does this PR do?

Update lib documentation.

Changes:
- fix: `TaskOutputRating` expects `value=int` (not `score=int`) and an enum for `type="five_star"`)
- docs: add example about tagging Task runs programmatically

## Summary by CodeRabbit

* **Documentation**
  * Updated the example for creating a TaskOutputRating to use named parameters and enum constants.
  * Added a new section demonstrating how to programmatically tag Kiln Task runs based on output content.

### DIFF
--- a/libs/core/README.md
+++ b/libs/core/README.md
@@ -153,7 +153,10 @@ item = kiln_ai.datamodel.TaskRun(
             type=kiln_ai.datamodel.DataSourceType.human,
             properties={"created_by": "Jane Doe"},
         ),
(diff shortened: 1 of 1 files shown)


libs/core/README.md
```diff
@@ -237,6 +240,25 @@ for run in task.runs():
 
 ```
 
+## Tagging Task Runs programmatically
```
scosman: Add a link in the table of contents please!

PR #467, opened 2025-07-31
Fixed "Fine-Tune Prompt" link not working on "Compare Run Methods" page

aiyakitori:
## What does this PR do?

Previously in "Compare Run Methods" page tapping on "Fine-Tune Prompt" does not navigate to the corresponding Fine Tune run. Added logic to construct url for fine tune prompt ids. 

## Checklists

- [YES] New tests have been added to any work in /lib

https://github.com/user-attachments/assets/b45dc179-918b-43b4-ae5e-a5d0b785b632

## Summary by CodeRabbit

* **New Features**
  * Improved handling of prompt links for fine-tuned prompts, generating more accurate URLs for prompt IDs with the "fine_tune_prompt::" prefix.

* **Tests**
  * Added comprehensive tests to verify correct URL construction for various prompt ID formats and edge cases.

### DIFF
--- a/app/web_ui/src/lib/utils/link_builder.ts
+++ b/app/web_ui/src/lib/utils/link_builder.ts
@@ -6,6 +6,13 @@ export function prompt_link(
   if (!project_id || !task_id || !prompt_id) {
     return undefined
   }
+  // Special case for fine-tuned prompts
+  if (prompt_id.startsWith("fine_tune_prompt::")) {
+    const trimmed_prompt_id = prompt_id.replace("fine_tune_prompt::", "")
+    const fine_tune_id =
+      trimmed_prompt_id.split("::").pop() || trimmed_prompt_id
+    return `/fine_tune/${project_id}/${task_id}/fine_tune/${encodeURIComponent(fine_tune_id)}`
+  }
   if (!prompt_id.includes("::")) {
     // not an ID style prompt, link to static
     return `/prompts/${project_id}/${task_id}/generator_details/${encodeURIComponent(prompt_id)}`
--- a/app/web_ui/src/lib/utils/link_builder.ts
+++ b/app/web_ui/src/lib/utils/link_builder.ts
@@ -1,3 +1,5 @@
+const FINE_TUNE_PROMPT_PREFIX = "fine_tune_prompt::"
+
 export function prompt_link(
   project_id: string,
   task_id: string,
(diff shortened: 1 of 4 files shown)


app/web_ui/src/lib/utils/link_builder.ts
```diff
@@ -6,6 +8,14 @@ export function prompt_link(
   if (!project_id || !task_id || !prompt_id) {
     return undefined
   }
+  // Special case for fine-tuned prompts
+  if (prompt_id.startsWith(FINE_TUNE_PROMPT_PREFIX)) {
+    // Get the last component of the prompt ID. fine_tune_prompt::fine_tune_id
+    const trimmed_prompt_id = prompt_id.replace(FINE_TUNE_PROMPT_PREFIX, "")
+    const fine_tune_id =
+      trimmed_prompt_id.split("::").pop() || trimmed_prompt_id
+    return `/fine_tune/${project_id}/${task_id}/fine_tune/${encodeURIComponent(fine_tune_id)}`
```
scosman, line 17: nit: why encode one but not the others? I'm fine either way (I think our IDs only allow url compatible characters so fine without), but should be consistent.

app/web_ui/src/lib/utils/link_builder.ts
```diff
@@ -6,6 +8,14 @@ export function prompt_link(
   if (!project_id || !task_id || !prompt_id) {
     return undefined
   }
+  // Special case for fine-tuned prompts
+  if (prompt_id.startsWith(FINE_TUNE_PROMPT_PREFIX)) {
+    // Get the last component of the prompt ID. fine_tune_prompt::fine_tune_id
+    const trimmed_prompt_id = prompt_id.replace(FINE_TUNE_PROMPT_PREFIX, "")
+    const fine_tune_id =
```
scosman, line 15: why the split/pop? If the format is `fine_tune_prompt::fine_tune_id` then replace should do it.

app/web_ui/src/lib/utils/link_builder.ts
```diff
@@ -6,6 +8,14 @@ export function prompt_link(
   if (!project_id || !task_id || !prompt_id) {
     return undefined
   }
+  // Special case for fine-tuned prompts
```
scosman, line 11: what about task_run_config::? I guess we don't have a page for them?

aiyakitori: It seems like it, they are listed under "Compare Run Methods" page.

scosman: ah, please link it!

scosman: oh, but that needs and eval ID? We need to add a screen for these eventually. TBD when/where.

app/web_ui/src/lib/utils/link_builder.ts
aiyakitori: Turns out the format of the id is `fine_tune_prompt::[project_id]::[task_id]::fine_tune_id`

app/web_ui/src/lib/utils/link_builder.ts
aiyakitori: This is following the same pattern as other links. It only encodes the last component.

scosman (approved):

PR #469, opened 2025-08-01
Fix connect provider not centered in initial setup screen

aiyakitori:
## What does this PR do?

The connect provider interface was not properly centered during the initial setup flow, leading to poor visual alignment and user experience issues.

This PR fixes the centering issue by:
1. Added setup context awareness: Introduced an is_initial_setup_page prop to distinguish between the initial setup flow and other provider connection scenarios.
2. Applied conditional centering: Used Flexbox utilities (flex flex-col items-center) to center the provider connection interface when accessed during initial setup.
3. Maintained existing behavior: Other provider connection screens (outside of initial setup) remain unchanged

Before: 
image

After:
image

Connect Provider screen under `/settings/providers` remained unchanged. 
image

## Summary by CodeRabbit

* **New Features**
  * Improved layout styling for the provider connection page during initial setup, enhancing the visual alignment and user experience.

### DIFF
--- a/app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
+++ b/app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
@@ -301,6 +301,7 @@
     (provider) => provider.connected,
   )
   export let intermediate_step = false
+  export let is_initial_setup_page = false
   let api_key_provider: Provider | null = null
   $: {
     intermediate_step = api_key_provider != null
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
```diff
@@ -301,6 +301,7 @@
     (provider) => provider.connected,
   )
   export let intermediate_step = false
+  export let is_initial_setup_page = false
```
scosman: A component shouldn't be aware of its parent or parent's layout needs. Just name it `centered` and let the parent component set centred.

scosman (approved):

Issue #472, opened 2025-08-01
[Feature Request] Add Docker Model Runner Support


ilopezluna:
**Is your feature request related to a problem? Please describe.**
Support for running AI models locally using [Docker Model Runner](https://docs.docker.com/ai/model-runner/). This provides an alternative for local model execution with no API key required.

**Checks**

- [x] I've searched [the docs](https://docs.getkiln.ai) for a solution
- [x] I've searched for existing Github issues

**Describe the solution you'd like**
DMR as an available model provider

PR #473, opened 2025-08-01
Adds docker model runner

ilopezluna:
## What does this PR do?

Add support to Docker Model Runner. The implementation follows the same pattern as Ollama integration but uses OpenAI-compatible endpoints. 

## Related Issues

https://github.com/Kiln-AI/Kiln/issues/472

I, @ilopezluna, confirm that I have read and agree to the [Contributors License Agreement](https://github.com/Kiln-AI/Kiln/blob/main/CLA.md).

## Checklists

- [] New tests have been added to any work in /lib

## Summary by CodeRabbit

* **New Features**
  * Docker Model Runner added as a first-class local provider: connect flow, public connect API, model discovery, UI connect/status controls, icon, and optional custom URL.
* **Bug Fixes**
  * Improved connection validation and error handling; updated API-key/disconnect behavior for the new provider.
* **Documentation**
  * Provider listings and descriptions updated to include Docker Model Runner.
* **Tests**
  * Extensive unit and integration tests covering connection, model parsing, availability, and config persistence.


scosman:
@ilopezluna very cool!

Our CR bot code rabbit found some small issues (typos, etc). Want to tackle those (or just mark as resolved if it's wrong about something) and let me know when it's ready?

ilopezluna:
> @ilopezluna very cool!
> 
> Our CR bot code rabbit found some small issues (typos, etc). Want to tackle those (or just mark as resolved if it's wrong about something) and let me know when it's ready?

Thank you @scosman! CodeRabbit looks really cool, I hadn’t heard of it before 👏
I’ve just pushed the changes it suggested. I don’t have much experience with Python, so I’m trusting its judgment more than mine.

app/desktop/studio_server/provider_api.py
```diff
@@ -77,6 +82,70 @@ async def connect_ollama(custom_ollama_url: str | None = None) -> OllamaConnecti
...
+        if docker_connection is None:
+            raise HTTPException(
+                status_code=500,
+                detail="Failed to parse Docker Model Runner data - unsure which models are available.",
+            )
+
+    except openai.APIConnectionError:
+        raise HTTPException(
+            status_code=417,
+            detail="Failed to connect. Ensure Docker Model Runner is running.",
```
scosman: I had it running, but didn't have TCP connections enabled and got this message. 

I figured it out, but let's update the string to something more explicit: "Failed to connect. Ensure Docker Model Runner is running and you enabled TCP connections. See the Docker Model Runner docs for instructions." or something like that. Might be pushing the UI a bit, but they explicitly clicked connect so a good error is important.

app/desktop/studio_server/provider_api.py
```diff
@@ -77,6 +82,70 @@ async def connect_ollama(custom_ollama_url: str | None = None) -> OllamaConnecti
     return ollama_connection
 
 
+async def connect_docker_model_runner(
+    custom_docker_url: str | None = None,
+) -> DockerModelRunnerConnection:
+    # Similar to Ollama, but uses OpenAI-compatible endpoints
```
scosman: Nit: our Ollama integration is also using their OpenAI-compatible endpoint. They have their own API and OAI compatibility.

app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
```diff
@@ -369,9 +387,9 @@
 
   let custom_ollama_url: string | null = null
 
-  const connect_ollama = async (user_initated: boolean = true) => {
+  const connect_ollama = async (user_initiated: boolean = true) => {
```
scosman, line 427: heh, it was my typo all along 😀

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -181,6 +181,9 @@ async def response_format_options(self) -> dict[str, Any]:
                 if provider_name == ModelProviderName.ollama:
                     # Ollama added json_schema to all models: https://ollama.com/blog/structured-outputs
                     return self.json_schema_response_format()
+                elif provider_name == ModelProviderName.docker_model_runner:
+                    # Docker Model Runner uses OpenAI-compatible API with JSON schema support
```
scosman, line 185: JSON support is a minefield. Ollama is the only provider where we can just say "always use json schema" because they killed it on their constrained decoding implementation. 

Do you know if DMR is proper constrained decoding for and schema, or just supports the OpenAI API param, but your mileage may vary. There are ways to make Kiln work with either.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -1225,6 +1225,12 @@ class KilnModel(BaseModel):
                 supports_structured_output=False,
                 supports_data_gen=False,
             ),
+            KilnModelProvider(
+                name=ModelProviderName.docker_model_runner,
+                model_id="ai/llama3.2",
```
scosman: We should set the size explicitly, so if the default ever changes, our data is still correct. `llama3.2:3B-Q4_K_M`?

We should also add the rest of the models. I know the list is bit and ugly... but it's actually really good for usability. Details: https://getkiln.ai/blog/i_wrote_2000_llm_test_cases_so_you_dont_have_to

Adding the models docker supports in there will do a few nice things:
 - Nice names in the UI
 - Add them to the test suite
 - Once tested that they will appear higher in the dropdown as we'll trust them to perform certain tasks better (JSON mode, complex requests like synthetic data, etc).

scosman:
oh small nit: Finetune badge doesn't need a star. Stars are to differentiate "suggested" but we don't have that concept for fine tunes

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
ilopezluna: DMR just supports the OpenAI API param, I'll will remove the `elseif` and use the default case

libs/core/kiln_ai/adapters/ml_model_list.py
ilopezluna: I have update the model references to include a more explicit tag `<weight>-<quantization>`
I considered adding all supported models as well, but I thought it might be better to include them in a follow-up PR that focuses solely on listing the supported models.

That said, if you’d prefer to include them now, I’m happy to do so.

ilopezluna:
> One thing I would change: the 'latest' tag has been nothing but headaches for me with Ollama. Usability wise, it's a bit hard to figure out which model size it is. Over time they can release new models in same family, and then 'latest' is either not actually latest, or a different size than it used to be. When we store that a response was generated by a specific model 'latest' can burn us; we want the model name to be explicit and permanent. I'd suggest keeping model tags tied to the size.

Yeah, I remember we had this same discussion in the DMR team :) 
We were very close to _not_ including the `latest` tag, precisely for the reasons you mentioned.
We ended up adding it because we thought it might help new users. At the time, we were focused on driving adoption, so we decided to include it and updated the model overview on Hub to clearly indicate which version `latest` points to: https://hub.docker.com/r/ai/llama3.2
However I still not convinced of using this tag, because as you mention, it creates confusion.

ilopezluna:
> oh small nit: Finetune badge doesn't need a star. Stars are to differentiate "suggested" but we don't have that concept for fine tunes

Sorry @scosman I didn't quite get that. Could you please clarify what you mean?

scosman:
> Sorry @scosman I didn't quite get that. Could you please clarify what you mean?

That was a comment on another PR. No idea how I managed to drop it here. Please disregard!

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
scosman: Some servers support the parameter, but don't back it by constrained decoding. In those cases, the generated text can fail to comply to the schema. In those cases can be better to have model-by-model decision on how to get structured data out. For example, some models are great at tool calling, but poor at JSON, so we prefer tool calling. It really is a minefield 😀.

That said, it looks like DMR is doing proper constrained decoding with llama.cpp. I threw some complex schemas at Qwen 0.6b and it was 100%. So looks like this approach is good!

No need to remove the case. The default could change later. It's good to explicitly decide DMR uses json_schema if that's what is best.

ilopezluna: I'll update it later today

ilopezluna: Done! BTW did you manually create the requests to verify DMR with those complex schemas, or did you use a tool from the repo?

scosman: I used kiln to verify. I created a task with a complex input schema, then tried synthetic data gen on it. Data gen generates a nested array of possible inputs, so it's a reasonably complex schema.

app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
```diff
@@ -452,6 +470,82 @@
       custom_url_str
   }
 
+  let custom_docker_url: string | null = null
```
scosman: You don't have any way of setting the custom URL in the UI. For Ollama, if they click connect and it fails we show a link+modal. Can probably copy that here?

Image

libs/core/kiln_ai/adapters/docker_model_runner_tools.py
```diff
@@ -0,0 +1,119 @@
...
+        base_url = docker_model_runner_base_url()
+        # Use OpenAI client to get models list
+        client = openai.OpenAI(
+            api_key="dummy",  # Docker Model Runner doesn't require API key
+            base_url=f"{base_url}/v1",
+            max_retries=0,
+        )
+        models_response = client.models.list()
+        # Convert to dict format for parsing
+        models_dict = {"data": [{"id": model.id} for model in models_response]}
```
scosman: It's strange to go from a nice typed object into a dict, then to have a bunch of type safety checks in parse_docker_model_runner_models.

You can make parse_docker_model_runner_models take a `List[Model]` instead of Any and it will be a lot cleaner.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -1225,6 +1225,12 @@ class KilnModel(BaseModel):
                 supports_structured_output=False,
                 supports_data_gen=False,
             ),
+            KilnModelProvider(
+                name=ModelProviderName.docker_model_runner,
+                model_id="ai/llama3.2:3B-Q4_K_M",
+                structured_output_mode=StructuredOutputMode.json_schema,
+                supports_data_gen=False,
+            ),
```
scosman, line 1411: It would be good to add the others in this PR. The usability of DMR goes way up with them.

app/desktop/studio_server/provider_api.py
```diff
@@ -77,6 +82,69 @@ async def connect_ollama(custom_ollama_url: str | None = None) -> OllamaConnecti
...
+        and not custom_docker_url.startswith("http://")
+        and not custom_docker_url.startswith("https://")
+    ):
+        raise HTTPException(
+            status_code=400,
+            detail="Invalid Docker Model Runner URL. It must start with http:// or https://",
+        )
+
+    try:
+        from kiln_ai.adapters.docker_model_runner_tools import (
```
scosman: nit: import should be at the top of the file unless there reasons it needs to have it's loading delayed

app/desktop/studio_server/provider_api.py
```diff
@@ -77,6 +82,69 @@ async def connect_ollama(custom_ollama_url: str | None = None) -> OllamaConnecti
...
+            docker_model_runner_base_url,
+        )
+
+        base_url = custom_docker_url or docker_model_runner_base_url()
+
+        # Use OpenAI client to test connection
+        client = openai.OpenAI(
```
scosman: Duplicate logic. Just use get_docker_model_runner_connection?

app/desktop/studio_server/provider_api.py
```diff
@@ -170,6 +238,11 @@ async def get_available_models() -> List[AvailableModels]:
         if ollama_models:
             models.insert(0, ollama_models)
 
+        # Docker Model Runner is special: check which models are installed
+        docker_models = await available_docker_model_runner_models()
+        if docker_models:
+            models.insert(-1 if ollama_models else 0, docker_models)
```
scosman: nit: -1 is a bit strange (before last item). Did you mean 1?

Maybe just add docker at 0 first, and Ollama at 0 after. That way Ollama keeps top spot if both are present?

app/desktop/studio_server/provider_api.py
```diff
@@ -894,6 +975,88 @@ async def available_ollama_models() -> AvailableModels | None:
...
+                    docker_models.models.append(
+                        ModelDetails(
+                            id=model.name,
+                            name=model.friendly_name,
+                            supports_structured_output=docker_provider.supports_structured_output,
+                            supports_data_gen=docker_provider.supports_data_gen,
+                            supports_logprobs=docker_provider.supports_logprobs,
+                            suggested_for_data_gen=docker_provider.suggested_for_data_gen,
+                            suggested_for_evals=docker_provider.suggested_for_evals,
+                            uncensored=False,
```
scosman: should be `docker_provider.uncensored` and `docker_provider.suggested_for_uncensored_data_gen`, not hardcoded. We can OTA update the list of models, so if docker adds a Mistral model we can properly mark as uncensored over the air.

app/desktop/studio_server/provider_api.py
```diff
@@ -894,6 +975,88 @@ async def available_ollama_models() -> AvailableModels | None:
...
+                            # Docker Model Runner uses OpenAI-compatible API with JSON schema support
+                            structured_output_mode=StructuredOutputMode.json_schema,
+                        )
+                    )
+        for docker_model in docker_connection.untested_models:
+            docker_models.models.append(
+                ModelDetails(
+                    id=docker_model,
+                    name=docker_model,
+                    supports_structured_output=False,
```
scosman: I think we can hardcode supports_structured_output to true since DMR supports json_schema constrained decoding natively. That way you won't get big ugly warnings using docker for JSON tasks.

leonardmq:
Nice! I see DMR also has some (or [one](https://hub.docker.com/r/ai/mxbai-embed-large)?) embedding model that we might be able to add support for our upcoming RAG tooling after this is merged.

This is the first external PR since we added a test coverage CI, I tried to run it (results are [here](https://github.com/Kiln-AI/Kiln/actions/runs/16723256659)), but looks like it is not posting the nicely formatted report back as a comment into the PR if it is from a fork 😄

.cursor/rules/project.mdc
```diff
@@ -8,7 +8,7 @@ alwaysApply: true
 - The project supports Python 3.10 and above
 - When writing tests:
   1.  Always use pytest for tests in python code
-  2.  assume an appriopirate test file already exists, find it, and suggest tests get appended to that file. If no such file exists, ask me before assuming a new test file is the correct route.
+  2.  assume an appropriate test file already exists, find it, and suggest tests get appended to that file. If no such file exists, ask me before assuming a new test file is the correct route.
```
scosman, line 11: 😬 

Thanks

ilopezluna:
> Nice! I see DMR also has some (or [one](https://hub.docker.com/r/ai/mxbai-embed-large)?) embedding model that we might be able to add support for our upcoming RAG tooling after this is merged.
> 
We recently added: https://hub.docker.com/r/ai/granite-embedding-multilingual and https://hub.docker.com/r/ai/nomic-embed-text-v1.5 as well 👍 

> This is the first external PR since we added a test coverage CI, I tried to run it (results are [here](https://github.com/Kiln-AI/Kiln/actions/runs/16723256659)), but looks like it is not posting the nicely formatted report back as a comment into the PR if it is from a fork 😄
>
Thanks for pointing it out @leonardmq ! I'm not familiar with the coverage conventions of the project, could you advise me on whether I should add more tests?

leonardmq:
> Thanks for pointing it out @leonardmq ! I'm not familiar with the coverage conventions of the project, could you advise me on whether I should add more tests?

@ilopezluna - we do not have a hard number for coverage requirements, current overall backend coverage is at 92%. We try to go for 100% coverage, but some modules are harder to test.

Latest diff coverage: https://github.com/Kiln-AI/Kiln/actions/runs/16907251293/artifacts/3746939147

My advice would be to look at the files with low coverage in this diff coverage and see if the untested lines are worth testing, and see if we already have similar tests for `ollama` or some other provider. For example, for `connect_docker_model_runner` (in `provider_api.py`), we might have some `ollama` tests that might be copiable (in `test_provider_api.py`). Disclaimer that I am not too familiar with our Ollama code, so am not sure.

Precision: this is only about the backend; we have few tests on the frontend at the moment so no need to worry about that part

libs/core/kiln_ai/adapters/ml_model_list.py
ilopezluna: Added in https://github.com/Kiln-AI/Kiln/pull/473/commits/6fc2711bdfe0c5665e52b3ea44699614b5f897da

ilopezluna: BTW, I've also added the models with the latest and default tags. Most DMR users will use those, although, as we discussed, it can still be confusing.
Because these tags point to the same model I was wondering that would be nice to have a way to allow multiple ids or aliases per models.
Anyway, let me know if you prefer to remove these model references from this PR.

scosman: Ah, so we can only have 1 entry per provider in there. I'll add a test to check that. If we add multiple "Llama 8B on DMR" becomes ambiguous which isn't good for reproducibility.

Let's just use the Q4_K_M versions (or similar)? The user can still use the others; they will automatically appear in the dropdown, but with their own more specific IDs.

app/desktop/studio_server/provider_api.py
```diff
@@ -975,6 +1040,88 @@ async def available_ollama_models() -> AvailableModels | None:
...
+                            # Docker Model Runner uses OpenAI-compatible API with JSON schema support
+                            structured_output_mode=StructuredOutputMode.json_schema,
+                        )
+                    )
+        for docker_model in docker_connection.untested_models:
+            docker_models.models.append(
+                ModelDetails(
+                    id=docker_model,
+                    name=docker_model,
+                    supports_structured_output=True,
```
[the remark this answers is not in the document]
scosman: this was intentional

libs/core/kiln_ai/adapters/test_docker_model_runner_tools.py
```diff
@@ -0,0 +1,309 @@
...
+import httpx
+import openai
+import pytest
+
+from kiln_ai.adapters.docker_model_runner_tools import (
+    DockerModelRunnerConnection,
+    docker_model_runner_base_url,
+    parse_docker_model_runner_models,
+)
+from libs.core.kiln_ai.datamodel.datamodel_enums import ModelProviderName
```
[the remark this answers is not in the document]
scosman: @ilopezluna we should take this suggestion. The libs/core path is repo-specific, won't work for folks using the library or running tests from other directories (90% sure, but still best to use package targeting imports).

scosman (changes requested):
Looking good! Few tiny changes then good to go:

 - We only support one model provider per provider in ml_model_list (see comment). I should have had a test.
 - one small import fix in the tests

libs/core/kiln_ai/adapters/test_docker_model_runner_tools.py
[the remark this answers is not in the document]
ilopezluna: Addressed [here](https://github.com/Kiln-AI/Kiln/pull/473/commits/aacc0b91b5fb5995b51a677676dc5d6ad3c32c21)

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -1606,6 +1639,20 @@ class KilnModel(BaseModel):
             ),
         ],
     ),
+    # Gemma 3 270M
+    KilnModel(
+        family=ModelFamily.gemma,
+        name=ModelName.gemma_3_0p27b,
+        friendly_name="Gemma 3 270M",
```
scosman, line 1646: That was fast!

libs/core/kiln_ai/adapters/test_ml_model_list.py
```diff
@@ -186,3 +188,39 @@ def test_no_reasoning_for_structured_output():
                 assert provider.reasoning_capable, (
                     f"{model.name} {provider.name} has reasoning_optional_for_structured_output but is not reasoning capable. This field should only be defined for models that are reasoning capable."
                 )
+
+
+def test_unique_providers_per_model():
```
scosman, line 193: Thanks!

PR #474, opened 2025-08-01
Skip create task on import

aiyakitori:
## What does this PR do?

When importing an existing Kiln project that already contains tasks, users were being redirected through the "Create a Task" setup flow. 

This PR adds routing logic that checks whether an imported project already has existing tasks. If the project has existing tasks it would now skip the task creation step and go to Select Task. If the project has no tasks, the flow remained the same as before. 

## Summary by CodeRabbit

* **New Features**
  * Improved post-import navigation: After importing a project, users are redirected based on whether the project contains tasks. If tasks exist, users are taken to task selection; otherwise, they are guided to the project page to create tasks.

* **Bug Fixes**
  * Enhanced error handling during project import to ensure users are redirected appropriately even if there is an issue fetching project tasks.

### DIFF
--- a/app/web_ui/src/routes/(fullscreen)/setup/(setup)/create_project/edit_project.svelte
+++ b/app/web_ui/src/routes/(fullscreen)/setup/(setup)/create_project/edit_project.svelte
@@ -165,11 +165,11 @@
 
       if (redirect_on_created && data?.id) {
         // Check if the imported project has tasks to decide where to redirect
-        const should_skip_task_creation = await check_project_has_tasks(data.id)
+        const should_skip_task_creation = await project_has_tasks(data.id)
 
         if (should_skip_task_creation) {
           // Project has tasks, go directly to main app
-          goto("/")
+          goto("/run")
         } else {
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(fullscreen)/setup/(setup)/create_project/edit_project.svelte
```diff
@@ -162,8 +162,18 @@
...
       await tick()
+
       if (redirect_on_created && data?.id) {
-        redirect_to_project(data.id)
+        // Check if the imported project has tasks to decide where to redirect
+        const should_skip_task_creation = await project_has_tasks(data.id)
+
+        if (should_skip_task_creation) {
+          // Project has tasks, go directly to main app
+          goto("/run")
```
scosman: Try UX of `/setup/select_task` (private window will remove saved data). Let's them pick a task, which they won't have done yet right after import.

scosman (approved):
Lgtm!

PR #476, opened 2025-08-02
Add Cerebras [WIP]

scosman:
Qwen-Coder attempt at adding a  provider. Zero human edits on first commit. Edited after be me.

3.1M input tokens, 11k output tokens. Used Cline.

Prompt:

```
I want to add Cerebras as a provider to Kiln.

 - Add it in ML model list as provider (will cause type errors where we need to write code). Run the type checker (if you need to use terminal and can't see them automatically run `uv run pyright .` to ) to find issues.
 - It should be modeled like Ollama: a custom OpenAI compatible endpoing based router. See adapter_registry.py. Use it's open AI compatible endpoint `https://api.cerebras.ai/v1 `
 - run generate_schema.sh to update our frontend APIs to add the new provider. This will cause type errors you need to respove on the front end.
 - add necessary UI in connect_provider and other places
```


app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
```diff
@@ -181,6 +180,18 @@
         "Bedrock is quite difficult to setup.\nFor beginners we suggest other providers, like OpenRouter, as they easier to setup and have more models.",
       api_key_fields: ["Access Key", "Secret Key"],
     },
+    {
+      name: "Cerebras",
+      id: "cerebras",
+      description: "Exceptionally fast inference on custom hardware.",
+      featured: false,
+      api_key_steps: [
+        "Go to https://cerebras.ai/settings/keys",
```
leonardmq: This URL is a `404`. The actual URL takes in the org ID in the path, so we cannot link to it directly.

We can link to https://cloud.cerebras.ai/ and tell them to navigate to the API Keys section from there. This is what their official docs do here: https://inference-docs.cerebras.ai/quickstart

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -2422,6 +2447,14 @@ class KilnModel(BaseModel):
                 formatter=ModelFormatterID.qwen3_style_no_think,
                 supports_data_gen=True,
             ),
+            KilnModelProvider(
```
leonardmq, line 2450: That one has the same weird formatting problem as some of the SiliconFlow model, where we get weird leading whitespace / empty line (only when thinking is disabled; if thinking is enabled on this same model, we don't get the extra whitespace)

PR #478, opened 2025-08-05
Update Run Properties UI with drop down

aiyakitori:
## What does this PR do?

Update the Run properties to use a drop down to show additional fields. 
* Added temperature and top p. 
* Move some of the properties to the See All list. 
* Use friendly name for model provider instead of their id. 

image

## Checklists

- [ YES] New tests have been added to any work in /lib

### DIFF
--- a/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
+++ b/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
@@ -154,23 +154,13 @@
     }
 
     if (run?.output?.source?.properties?.structured_output_mode) {
-      const mode = run.output.source.properties.structured_output_mode
-      let displayValue = mode
-      // Convert technical names to more user-friendly display names
-      if (mode === "json_schema") {
-        displayValue = "JSON Schema"
-      } else if (mode === "json_instructions") {
-        displayValue = "JSON Instructions"
-      } else if (mode === "native_json") {
-        displayValue = "Native JSON"
-      } else if (mode === "unknown") {
-        displayValue = "Not Set"
+      let mode = run.output.source.properties.structured_output_mode
+      if (typeof mode === "string" && mode.startsWith("json_")) {
+        properties.push({
+          name: "JSON Mode",
+          value: mode,
+        })
       }
-
-      properties.push({
-        name: "JSON Mode",
-        value: displayValue,
-      })
     }
 
     if (run?.input_source?.properties?.created_by) {
--- a/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
+++ b/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
@@ -135,6 +135,20 @@
       })
     }
 
+    if (run?.output?.source?.properties?.temperature !== undefined) {
+      properties.push({
+        name: "Temperature",
+        value: run.output.source.properties.temperature,
+      })
+    }
+
+    if (run?.output?.source?.properties?.top_p !== undefined) {
+      properties.push({
+        name: "Top P",
+        value: run.output.source.properties.top_p,
+      })
+    }
+
     if (run?.output?.source?.properties?.structured_output_mode) {
       const mode = run.output.source.properties.structured_output_mode
       let displayValue = mode
@@ -159,13 +173,6 @@
       properties.push({
         name: "Created By",
         value: run.input_source.properties.created_by,
-      })
-    }
-
-    if (run?.output?.source?.properties?.top_p !== undefined) {
-      properties.push({
-        name: "Top P",
-        value: run.output.source.properties.top_p,
       })
     }
 
--- a/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
+++ b/app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
@@ -20,6 +20,7 @@
   import { goto } from "$app/navigation"
   import DeleteDialog from "$lib/ui/delete_dialog.svelte"
   import PropertyList from "$lib/ui/property_list.svelte"
+  import Collapse from "$lib/ui/collapse.svelte"
   import { prompt_link } from "$lib/utils/link_builder"
   import type { ProviderModels, PromptResponse } from "$lib/types"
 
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -104,34 +139,34 @@
...
-      topic_path = run?.input_source?.properties?.topic_path?.replaceAll(
-        ">>>>>",
-        " > ",
-      )
+    if (run?.output?.source?.properties?.structured_output_mode) {
+      let mode = run.output.source.properties.structured_output_mode
+      if (typeof mode === "string" && mode.startsWith("json_")) {
+        properties.push({
+          name: "JSON Mode",
+          value: mode,
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: Should we add a friendly name for structured_output_mode?

Since there isn't an existing schema for "Available Structured Output Mode", this will be a more extensive change if we want to follow the pattern for `provider_name_from_id `?

scosman:
visually it's a bit heavy. The stuff we want to hide is taking up more visual weight/space.

Can we make it just a "See all" that's underlined, text-grey-500, and "text-xs". When you click that, we set a var "show_all=true", which triggers a recalculate of properties list to be the full list, and hides the "See all".

Collapse is heavy here and makes it longer.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: Yes but keep it simple. No need for API (backend doesn't have friendly names, this is a pure UI concept anyways).

You can write a separate `.ts` that takes a `StructuredOutputMode` (defined in types.ts) and returns a friendly name. Trick is to make it automatically catch missing cases in the future if we add a new value. This is from sonnet using typechecking to catch missing values:

```
/**
 * Converts StructuredOutputMode to a human-readable string.
 * This function uses exhaustive case checking - if you add a new case to StructuredOutputMode,
 * TypeScript will force you to handle it here.
 */
export function structuredOutputModeToString(
  mode: StructuredOutputMode,
): string {
  switch (mode) {
    case "default":
      return "Default (Legacy)"
    case "json_schema":
      return "JSON Schema"
    case "function_calling_weak":
      return "Weak Function Calling"
    case "function_calling":
      return "Function Calling"
    case "json_mode":
      return "JSON Mode"
    case "json_instructions":
      return "JSON Instructions"
    case "json_instruction_and_object":
      return "JSON Instructions + Mode"
    case "json_custom_instructions":
      return "None"
    case "unknown":
      return "Unknown"
    default: {
      // This ensures exhaustive checking - if you add a new case to StructuredOutputMode
      // and don't handle it above, TypeScript will error here
      const exhaustiveCheck: never = mode
      throw new Error(`Unhandled StructuredOutputMode: ${exhaustiveCheck}`)
    }
  }
}
```

Then maybe update structured_output_options to use the function as well, so we don't repeat the names in 2 place.

Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: Updated `structured_output_options `

image

app/web_ui/src/lib/stores.ts
```diff
@@ -454,3 +455,39 @@ export function rating_options_for_sample(
     })
     .map((option) => option.requirement)
 }
+
+/**
+ * Converts StructuredOutputMode to a human-readable string.
```
scosman: This should be in formatters.ts, it's not a store. Stores are a svelte concept.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -261,11 +298,20 @@
...
+              ...get_properties(run, $current_task_prompts, $model_info),
+              ...(see_all_properties ? get_advanced_properties(run) : []),
+            ]}
             title="Properties"
           />
+          <button
+            class="text-xs text-gray-500 underline cursor-pointer bg-transparent border-none p-0"
+            on:click={() => (see_all_properties = !see_all_properties)}
+          >
+            {see_all_properties ? "See Less" : "See All"}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte:320]: add `text-left` as everything else is left aligned.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -90,6 +87,45 @@
...
+    }
+
+    let topic_path: string | undefined = undefined
+    if (
+      run?.input_source?.properties?.topic_path &&
+      typeof run?.input_source?.properties?.topic_path === "string"
+    ) {
+      topic_path = run?.input_source?.properties?.topic_path?.replaceAll(
+        ">>>>>",
+        " > ",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte:106]: 👍

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -104,34 +140,34 @@
...
-      typeof run?.input_source?.properties?.topic_path === "string"
-    ) {
-      topic_path = run?.input_source?.properties?.topic_path?.replaceAll(
-        ">>>>>",
-        " > ",
-      )
+    if (run?.output?.source?.properties?.structured_output_mode) {
+      let mode = run.output.source.properties.structured_output_mode
+      if (typeof mode === "string" && mode.startsWith("json_")) {
+        properties.push({
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: Frontend shouldn't make assumption about what  valid values are (startsWith).

I think you can update the helper to `return undefined` instead of `throw new Error(Unhandled`. Keep the type check for catching missing values, might need to `console.warn("missing type "+ exhaustiveCheck)` to get type checker to shut up about unused variable (without getting it to shut up when we actually want the error).

I believe the property list hides undefined values (but check)

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -261,11 +298,20 @@
           <div class="text-xl font-bold mb-4">Input</div>
           <Output raw_output={run.input} />
         </div>
-        <div class="w-72 2xl:w-96 flex-none flex flex-col">
+        <div class="w-72 2xl:w-96 flex-none flex flex-col gap-4">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte]: The gap also added space between header and property list so it's too large now. just add some `mt-4` to the button to only add space there.

scosman (approved):

PR #481, opened 2025-08-05
Don't require built in model names to be in the enum, since we now push new models OTA

scosman:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when selecting unavailable models or providers, ensuring the system returns no result instead of raising errors.

* **Tests**
  * Added and updated tests to verify that missing or misconfigured models and providers now result in a safe return value rather than an exception.

### DIFF
--- a/libs/core/kiln_ai/adapters/provider_tools.py
+++ b/libs/core/kiln_ai/adapters/provider_tools.py
@@ -74,21 +74,21 @@ def builtin_model_from(
     name: str, provider_name: str | None = None
 ) -> KilnModelProvider | None:
     """
-    Gets a model and provider from the built-in list of models.
+    Gets a model provider from the built-in list of models.
 
     Args:
         name: The name of the model to get
         provider_name: Optional specific provider to use (defaults to first available)
 
     Returns:
-        A tuple of (provider, model), if the model is found and has a provider
+        A KilnModelProvider, or None if not found
     """
     # Select the model from built_in_models using the name
     model = next(filter(lambda m: m.name == name, built_in_models), None)
     if model is None:
         return None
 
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/adapters/provider_tools.py
```diff
@@ -82,23 +81,17 @@ def builtin_model_from(
         provider_name: Optional specific provider to use (defaults to first available)
 
     Returns:
-        A tuple of (provider, model)
-
-    Raises:
-        ValueError: If the model or provider is not found, or if the provider is misconfigured
+        A tuple of (provider, model), if the model is found and has a provider
```
leonardmq: This method is returning `KilnModelProvider | None`, not a tuple. The entire docstring does not seem to match what it is doing now. 

Could also possibly rename the method as its name suggests it returns a model, but it returns a model provider instead

libs/core/kiln_ai/adapters/provider_tools.py
```diff
@@ -82,23 +81,17 @@ def builtin_model_from(
...
-        return None
-
     # Select the model from built_in_models using the name
-    model = next(filter(lambda m: m.name == name, built_in_models))
+    model = next(filter(lambda m: m.name == name, built_in_models), None)
     if model is None:
-        raise ValueError(f"Model {name} not found")
+        return None
 
     # If a provider is provided, select the provider from the model's provider_config
```
leonardmq: The comment refers to a `provider_config`, but not sure what it is - from searching `provider_config` in the code, it seems it might have been a historical / deprecated concept for OpenAI compatible providers. Not sure if the comment is still valid

PR #482, opened 2025-08-06
Add ExternalTool model

aiyakitori:
## What does this PR do?

Add a new `ExternalTool` data model to model configurations to make an external MCP server call. Starts with remote_mcp server for now. 

* Add new model for remote_mcp
* Add necessary fields to perform basic remote mcp call. 

## Checklists

- [Yes] Tests have been run locally and passed
- [Yes] New tests have been added to any work in /lib

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring external tools, allowing users to define connections to external servers for tool calls within projects.
  * Projects can now include and manage external tools alongside existing tasks.

* **Bug Fixes**
  * Improved validation for external tool configuration, ensuring required fields are present and correctly formatted.

* **Tests**
  * Introduced comprehensive tests to verify the creation, validation, and behavior of external tool configurations.

### DIFF
--- a/libs/core/kiln_ai/datamodel/external_tool.py
+++ b/libs/core/kiln_ai/datamodel/external_tool.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from typing import Any, Dict
 
 from pydantic import Field, model_validator
 
@@ -33,13 +34,9 @@ class ExternalTool(KilnParentedModel):
         default=None,
         description="A description of the external tool for you and your team. Will not be used in prompts/training/validation.",
     )
-    server_url: str = Field(
-        description="The URL of the remote MCP server.",
-        min_length=1,
-    )
-    headers: dict[str, str] | None = Field(
-        default=None,
-        description="HTTP headers to use when calling the server_url.",
+    properties: Dict[str, Any] = Field(
+        default={},
+        description="Configuration properties specific to the tool type.",
     )
 
     @model_validator(mode="after")
@@ -47,10 +44,23 @@ class ExternalTool(KilnParentedModel):
         """Validate that each tool type has the required configuration."""
         match self.type:
             case ToolType.remote_mcp:
-                if not self.server_url:
-                    raise ValueError("server_url must be set when type is 'remote_mcp'")
-                if not self.headers:
+                server_url = self.properties.get("server_url", None)
+                if not isinstance(server_url, str):
+                    raise ValueError(
+                        "server_url must be a string for external tools of type remote_mcp"
+                    )
+                if not server_url:
+                    raise ValueError(
+                        "server_url is required for external tools of type remote_mcp"
+                    )
+
+                headers = self.properties.get("headers", None)
+                if not headers:
                     raise ValueError("headers must be set when type is 'remote_mcp'")
+                if not isinstance(headers, dict):
+                    raise ValueError(
+                        "headers must be a dictionary for external tools of type remote_mcp"
+                    )
             case _:
                 # Type checking will catch missing cases
                 raise_exhaustive_enum_error(self.type)
--- a/libs/core/kiln_ai/datamodel/external_tool.py
+++ b/libs/core/kiln_ai/datamodel/external_tool.py
@@ -1,4 +1,4 @@
-from typing import Literal
+from enum import Enum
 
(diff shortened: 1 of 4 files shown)


libs/core/kiln_ai/datamodel/external_tool.py
```diff
@@ -0,0 +1,29 @@
...
+
+from kiln_ai.datamodel.basemodel import (
+    FilenameString,
+    KilnParentedModel,
+)
+
+
+class ExternalTool(KilnParentedModel):
+    """
+    Configuration for communicating with a remote MCP (Model Context Protocol) Server for LLM tool calls.
```
scosman: nit: remote != external. We'll have external but local too.

libs/core/kiln_ai/datamodel/external_tool.py
```diff
@@ -0,0 +1,29 @@
...
+    name: FilenameString = Field(description="The name of the external tool.")
+    description: str | None = Field(
+        default=None,
+        description="A description of the external tool for you and your team. Will not be used in prompts/training/validation.",
+    )
+    server_url: str = Field(
+        description="The URL of the remote MCP server.",
+        min_length=1,
+    )
+    api_key: str = Field(
```
scosman: nit: look at remote MCP protocol def. I don't think there's a standard concept of API KEY (Firecrawl jams it into URL). I think it's URL + headers? Just start with URL as a property, then iterate as you learn the protocol?

libs/core/kiln_ai/datamodel/external_tool.py
```diff
@@ -0,0 +1,46 @@
...
+
+    name: FilenameString = Field(description="The name of the external tool.")
+    type: Literal["remote_mcp"] = Field(
+        description="The type of external tool. Remote tools are hosted on a remote server",
+    )
+    description: str | None = Field(
+        default=None,
+        description="A description of the external tool for you and your team. Will not be used in prompts/training/validation.",
+    )
+    server_url: str = Field(
```
scosman: oh, I would make the properties all in a `properties: Dict[string, Any]`. Individual property fields will break when we add a second type and it doesn't have a server URL. They will also make backwards compatibility and customization a pain.

The validator can check something like

```
match self.type:
   case ExternalToolType.remote_mcp:
       server_url = self.properties.get("server_url", None)
       if not isinstance(server_url, str):
          raise ValueError("server_url must be a string for external tools of type remote_mcp")
       if not server_url:
          raise ValueError("server_url is required for external tools of type remote_mcp")
       if not server_url.startsWith("https://") and not server_url.startsWith("http://")
          raise ValueError(f"remote_mcp server_url must start with http:// or https://. Got: {server_url}")
```

libs/core/kiln_ai/datamodel/external_tool.py
```diff
@@ -0,0 +1,46 @@
...
+class ExternalTool(KilnParentedModel):
+    """
+    Configuration for communicating with a external MCP (Model Context Protocol) Server for LLM tool calls. External tools can be remote or local.
+
+    This model stores the necessary configuration to connect to and authenticate with
+    external MCP servers that provide tools for LLM interactions.
+    """
+
+    name: FilenameString = Field(description="The name of the external tool.")
+    type: Literal["remote_mcp"] = Field(
```
scosman: Use an enum, then you can also setup the validator with a proper exhaustive enum check, which forces you to consider each value when you add one.

search `raise_exhaustive_enum_error` for examples.

libs/core/kiln_ai/datamodel/external_tool.py
```diff
@@ -0,0 +1,46 @@
...
+    )
+
+    @model_validator(mode="after")
+    def validate_remote_mcp_config(self) -> "ExternalTool":
+        """Validate that remote_mcp type has server_url and headers configured."""
+        if self.type == "remote_mcp":
+            if not self.server_url:
+                raise ValueError("server_url must be set when type is 'remote_mcp'")
+            if self.headers is None or self.headers == {}:
+                raise ValueError(
```
scosman: empty should be fine. Most won't require headers.

scosman (approved):

PR #413, opened 2025-07-10
feat: RAG pipeline datamodel

leonardmq:
## What does this PR do?

Changes:
- add `RAGPipeline` datamodel to store the combination of `ExtractorConfig`, `ChunkerConfig`, `EmbeddingConfig`
- add endpoints to create / list `ChunkerConfig` and `EmbeddingConfig`
- add `RAGPipeline` orchestrator class to run a full RAG pipeline; to replace the `run_extractor_runner_with_status` which is too light for a whole pipeline run, client can subscribe to progress updates (and convert to SSE in the API layer)

The idea is to then use these in the UI and recenter the `Extractor` page around a higher level concept of `RAG Pipeline`, if we do E2E eval, users do not necessarily need separate views of each part of the pipeline. The main UI then would be a form that lets the user create a full pipeline:
- extractor: show the existing ones in a dropdown; or create one if you want a new one
- chunker config: show the existing ones in a dropdown (currently, the only difference between chunker configs is the `chunk_overlap` and `chunk_size` properties; as we only support one type (`fixed_window`))
- embedding config: can abstract this away and simply let the users pick an embedding model in the dropdown (under the hood, those will map to different embedding configs, but no need to expose that immediately)

## Checklists

- [x] Tests have been run locally and passed
- [x] New tests have been added to any work in /lib

PR #487, opened 2025-08-07
Leonard/rag vector indexing

leonardmq:
## What does this PR do?

Changes:
- add `VectorStoreConfig` model; it is project-scoped (looking ahead, probably will have different projects linking out to different (remote) stores)
- ref to `vector_store_config_id` in `RagConfig`
- add adapter for `VectorStore`, with an abstract base class, and concrete `LanceDBAdapter` that supports basic operations we need: `create_index`, `upsert_chunks`, `search` (WIP: refine interface)
- although we support adapters for vector storage, we will only have LanceDB for now, so the vector store config is created on the backend and not exposed to the user (WIP: not sure where to instantiate that default adapter or document its specific ID scheme)

## Checklists


leonardmq:
Current issues:
- lancedb requires `pyarrow`which is very large (110MB)
- some things to fix wrt embedding dimensions; we need to persist that into the data model because we need it down the line to create the index / table schema

leonardmq:
Removed weaviate, added qdrant - supports both vector and FTS with BM25, and we can add hybrid

leonardmq:
Currently, support for:
- lancedb -> problem is pyarrow is big, so build is big
- chromadb -> FTS is only exact substring match filtering (no BM25), hybrid hard to do
- qdrant -> good vector search, FTS (BM25 and BM42 variant), hybrid baked in (with RRF), but maybe local qdrant-client version too much of a toy version (different from the real db)

Removed HNSW support; for now we will be doing bruteforce KNN, which is fine so long as dataset is not in the tens of thousands. Also saves some UI and default values we need to figure out.

Next steps:
- pause this branch
- review feedback and merge https://github.com/Kiln-AI/Kiln/pull/413
- maybe use llama-index wrapper (unclear if suitable with index already shaped and filled out without llama-index upstream pipelines, apparently ok and only need to point llama index to the right property that contains our chunk text)
- integrate this into UI (need to add some progress tracking in RagWorkflowRunner for indexing)
- integrate into tool calling

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
```diff
@@ -0,0 +1,58 @@
...
+class SearchResult(BaseModel):
+    document_id: str = Field(description="The id of the Kiln document.")
+    chunk_idx: int = Field(description="The index of the chunk.")
+    chunk_text: str = Field(description="The text of the chunk.")
+    similarity: float | None = Field(
+        description="The score of the chunk, which depends on the similarity metric used."
+    )
+
+
+class KilnVectorStoreQuery(BaseModel):
```
scosman: nit: avoid "Kiln" in class names, it's all kiln 😆

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class LanceDBAdapter(BaseVectorStoreAdapter):
+    def __init__(
+        self,
+        rag_config: RagConfig,
+        vector_store_config: VectorStoreConfig,
+        lancedb_vector_store: LanceDBVectorStore,
```
scosman: I feel like the constructor should take care of building lancedb_vector_store. Move build_lancedb_vector_store into here, and call it inside __init__ (don't pass it in). Keep the registry thin, and platform specific logic isolated.

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                        value=node_hash,
+                        operator=FilterOperator.EQ,
+                    ),
+                ],
+                condition=FilterCondition.AND,
+            )
+        )
+        return len(nodes) > chunk_idx
+
+    def hash_node(
```
scosman: there are ways to speed this up if we find out we need it in benchmarking. Can stream in the values and use binary if JSON is slowing it down.

If we build indexes often might be worthwhile. If we find a way to only build once, probably not important.

Standard perf comment: don't optimize unless we've seen it be an issue in a benchmark. 

Pref approach would be:
1) benchmark "add_chunks_with_embeddings" with something like 100k embeddings. Can just generate random strings + embeddings. If it's fast enough, we're done.
2) break it down: benchmark hash_node, see how much time is from this.

I'd say worth adding the benchmark for #1 now.

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                }
+            ).encode()
+        ).hexdigest()
+
+    async def add_chunks_with_embeddings(
+        self,
+        records: list[Tuple[str, ChunkedDocument, ChunkEmbeddings]],
+    ) -> None:
+        nodes: List[TextNode] = []
+        for document_id, chunked_document, chunk_embeddings in records:
```
scosman: this is a big loop that could process 10M items, and hold the main thread hostage, causing slow perf of async things.

ChatGPT says `await asyncio.sleep(0)` this is idiomatic way to release control occasionally, allowing other asynchronous things to keep moving. Maybe call that after each doc (but not each chunk)?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                        text=chunk_text,
+                        embedding=embedding.vector,
+                        metadata={
+                            "kiln_doc_id": document_id,
+                            "kiln_chunk_idx": chunk_idx,
+                            "kiln_node_hash": node_hash,
+                        },
+                    )
+                )
+        await self.lancedb_vector_store.async_add(nodes)
```
scosman: what's the perf like if we call add each loop, instead of building all nodes in memory? Maybe per node, maybe per doc, maybe flush every 1000? But cap memory usage.

libs/core/kiln_ai/datamodel/vector_store.py
```diff
@@ -0,0 +1,111 @@
...
+    from kiln_ai.datamodel.project import Project
+
+
+class VectorStoreType(str, Enum):
+    LANCE_DB_FTS = "lancedb_fts"
+    LANCE_DB_HYBRID = "lancedb_hybrid"
+    LANCE_DB_VECTOR = "lancedb_vector"
+
+
+class LanceDBQueryType(str, Enum):
```
scosman: should be internal to lanced_adapter. Let it's constructor handle the mapping.

scosman: Can we just kill this separate enum?

Make the LanceDBAdapter take `vector_store_type: VectorStoreType.LANCE_DB_FTS | VectorStoreType.LANCE_DB_HYBRID | VectorStoreType.LANCE_DB_VECTOR` and store that value. Type checking will do the rest. Can check the store type. Makes it easier to add more too (QdrantVectorStore)

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+        self,
+        rag_config: RagConfig,
+        vector_store_config: VectorStoreConfig,
+        lancedb_vector_store: LanceDBVectorStore,
+    ):
+        super().__init__(rag_config, vector_store_config)
+        self.lancedb_vector_store = lancedb_vector_store
+        self.config_properties = self.vector_store_config.lancedb_properties
+
+    async def is_chunk_indexed(
```
scosman: not used and can remove? My search isn't finding any calls out side of tests.

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
```diff
@@ -0,0 +1,58 @@
...
+        records: list[Tuple[str, ChunkedDocument, ChunkEmbeddings]],
+    ) -> None:
+        pass
+
+    @abstractmethod
+    async def search(self, query: KilnVectorStoreQuery) -> List[SearchResult]:
+        pass
+
+    @abstractmethod
+    async def get_all_chunks(self) -> List[SearchResult]:
```
scosman: Perf wise this will never scale. Can we just remove it and force people to use proper search/indexes?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+    async def search(self, query: KilnVectorStoreQuery) -> List[SearchResult]:
+        query_result = await self.lancedb_vector_store.aquery(
+            VectorStoreQuery(
+                **self.build_kwargs_for_query(query),
+            ),
+            query_type=self.query_type,
+        )
+        return self.format_query_result(query_result)
+
+    async def get_all_chunks(self) -> List[SearchResult]:
```
scosman: per above, can we just delete?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                raise_exhaustive_enum_error(vector_store_config.store_type)
+
+    @staticmethod
+    def lancedb_path_for_config(rag_config: RagConfig) -> str:
+        data_dir = Config.shared().local_data_dir()
+        if isinstance(data_dir, str):
+            data_dir = Path(data_dir)
+        if rag_config.id is None:
+            raise ValueError("Vector store config ID is required")
+        return str(data_dir / "lancedb" / rag_config.id)
```
scosman: add a "rag_indexes" folder above lanceDB for future flexibility.

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                return LanceDBQueryType.HYBRID
+            case VectorStoreType.LANCE_DB_VECTOR:
+                return LanceDBQueryType.VECTOR
+            case _:
+                raise_exhaustive_enum_error(vector_store_config.store_type)
+
+    @staticmethod
+    def lancedb_path_for_config(rag_config: RagConfig) -> str:
+        data_dir = Config.shared().local_data_dir()
+        if isinstance(data_dir, str):
```
scosman: remove this. local_data_dir is typed. Might need to fix tests to pass Path, but don't change code to match tests, other way around.

libs/core/kiln_ai/utils/config.py
```diff
@@ -261,6 +261,14 @@ def update_settings(self, new_settings: Dict[str, Any]):
                 yaml.dump(current_settings, f)
             self._settings = current_settings
 
+    def local_data_dir(self) -> Path:
```
scosman: I think we can just remove this and use settings_dir. It's not a great name, but we don't need a "data" nested under there, the folder is already called `.kiln_ai` and has subfolders like logs (which is data). Just add the `rag_indexes` under it, like `.kiln_ai/rag_indexes/lancedb`

if anything, we rename the settings_dir accessor, or keep local_data_dir as an alias (just call settings_dir).

scosman (changes requested):
mostly perf comments/questions

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,243 @@
...
+                )
+
+                nodes.append(
+                    TextNode(
+                        text=chunk_text,
+                        embedding=embedding.vector,
+                        metadata={
+                            "kiln_doc_id": document_id,
+                            "kiln_chunk_idx": chunk_idx,
+                            "kiln_node_hash": node_hash,
```
scosman: Better to chat in person, but why use this hash? We have all the fields in the hash already stored in dedicated fields (text, embedding, doc_id and chunk index). Look ups will be faster if we query based on those fields (use index) instead of metadata (probably unindexed).

Can we kill the whole hash thing?

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
leonardmq: `llama_index` has a class (that we use) called `VectorStoreQuery` so this was to avoid naming collision / confusion as we use both:

https://github.com/Kiln-AI/Kiln/blob/dd7453dd0da485e9379a862405ce0c5ca87c806e/libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py#L184

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: The idea here was to make it easier to mock it and pass in a mock, but seems like we can `patch` this relatively easily. 

Before refactoring it, I need to triple check the `create | overwrite` param (https://github.com/Kiln-AI/Kiln/blob/dd7453dd0da485e9379a862405ce0c5ca87c806e/libs/core/kiln_ai/adapters/vector_store/vector_store_registry.py#L36) to ensure that we don't risk nuking the store on every initialization if we skip the factory - I thought it was behaving one way, but ended up wondering if it actually does anything

scosman: I think we can use similar factory, just inside init. But not sure I follow the comment.

leonardmq: Moved initialization of the LanceDB dependency into the `LanceDBAdapter` constructor in: https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

The `mode = create | overwrite` does not seem to have an impact as far as our use case is concerned (so removed it; by default it is `overwrite`)

libs/core/kiln_ai/datamodel/vector_store.py
leonardmq: Removed (other comment asking to remove it below)

leonardmq: These are the values that LanceDB adapter expects on creation (whereas the `VectorStoreType` is our own store type). Removed this enum, and made a switch that returns string literals instead inside the adapter.

https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Yes, that was one of the upsert attempts that I then dropped. Removed in https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
leonardmq: This was not meant to scale, this was just used in tests as some of them check the whole table to see what is going on. Moved the method into the test file instead. https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: This was not meant to scale, this was just used in tests as some of them check the whole table to see what is going on. Moved the method into the test file instead. https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Added `rag_indexes` folder in https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Now using `settings_dir` (as per other comment), which is not a `Path` anymore, but so this is removed.  https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/utils/config.py
leonardmq: Removed and using `settings_dir` instead (in the `LanceDBAdapter`). https://github.com/Kiln-AI/Kiln/pull/487/commits/93c2c7372fb56321e2cb19cf56f4f53638c44864

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
leonardmq: Have not renamed this; do you still want to rename it given the naming conflict mentioned above?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: We do batching at the caller; if we were getting 10M documents and their chunks here, they would be in memory upstream, so that would be no good before reaching this. 

Here we do the `chunks_text = await chunked_document.load_chunks_text()` for every document, so presumably that would already release the event loop on every document.

However, the `async_add` from `llama_index` is not actually `async` (it is an async wrapper around a synchronous blocking operation). LanceDB has an `async` API but the `llama_index` driver does not use it. So the `async_add` itself should be blocking the event loop.

`await asyncio.sleep()` at the document level would probably do the same that we already do with `chunks_text = await chunked_document.load_chunks_text()` at the same level.

Smaller batch size would probably alleviate; currently we batch documents at the caller, and internally here we insert all the chunks for a given document at once. If a document is very long and has say 5000 chunks, that is a big blocking write to LanceDB.

Made some changes in https://github.com/Kiln-AI/Kiln/pull/487/commits/90c81e14c0271c534232ac960952d81de156943b:
- added chunk batching inside the documents, to avoid outlier individual documents that have many thousands of chunks from causing huge writes
- commented out `table.optimize()`, maybe we don't need it, and it is not `async`, might be extra slow
- added `asyncio.sleep(0)` but suspect that is likely redundant given the above

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte
```diff
@@ -66,19 +66,29 @@
     "embedding",
     config_progress,
   )
+  $: indexing_progress_value = get_step_progress_value(
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte:69]: This whole progress tracking is clunky but don't think it is worth refactoring just yet. When refactoring time comes for this though, probably the idea will be to have the server send through each step, with their name and expected total, and status, so we can remove a lot of this logic.

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte
```diff
@@ -272,9 +295,15 @@
             <div class="font-medium text-sm text-base-content">
               {step.label}
             </div>
-            <div class="text-xs text-gray-500">
-              {step.progress} / {progress_max} documents
-            </div>
+            {#if step.name === "indexing"}
+              <div class="text-xs text-gray-500">
+                {step.progress} / {chunk_progress_max} chunks
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte:300]: On the backend, when we check how many records we have (for the initial progress setting), we only know the number of chunks that are indexed (we query the db for `count_records()`), so we need to express the indexing progress in terms of chunks.

libs/core/kiln_ai/adapters/rag/rag_runners.py
```diff
@@ -229,7 +243,7 @@ async def collect_jobs(self) -> list[ExtractorJob]:
         return jobs
 
     async def run(self) -> AsyncGenerator[RagStepRunnerProgress, None]:
-        async with shared_async_lock_manager.acquire(self.lock_key, timeout=0.5):
+        async with shared_async_lock_manager.acquire(self.lock_key, timeout=60):
```
leonardmq, line 263: I increased the lock timeout mostly because of this scenario:
- you have many RAG Configs that share ancestry (they have the same extractor for example)
- then you create a new document
- we trigger all the RAG Configs to run at the same time
- the ones that share ancestry will compete for the lock

By giving it some slack here, if you have even like 10 RAG Configs that share the same ancestry, probably going to be able to complete without any of the locks timing out.

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,292 @@
...
+        # it exposes higher level operations (that rely on internal
+        # fields like ref_doc_id); make sure implementation mirrors
+        # the upstream llama_index logic that we do not use
+        self.index = VectorStoreIndex(
+            [],
+            storage_context=storage_context,
+            embed_model=None,
+        )
+
+    async def delete_nodes_by_document_id(self, document_id: str) -> None:
```
leonardmq, line 104: We are not currently using this one, but I would want to keep it in here (it is not exposed in the base adapter), because it is not obvious that we support that or how to do that or what it requires (`ref_doc_id` be set upstream)

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,292 @@
...
+            vector_store=self.lancedb_vector_store
+        )
+
+        # FIXME:
+        # embed_model=None should be valid and result in llama_index initializing it
+        # to a mock (like it does elsewhere) but there is a fallback in the
+        # VectorStoreIndex constructor that overrides None with "default"
+        # and tries to load OpenAI and throws if the OPENAI_API_KEY is not set
+        # maybe should open an issue on their repo
+        os.environ["OPENAI_API_KEY"] = "dummy"
```
leonardmq: Not sure what to do here - I would want to avoid setting the real key since `llama_index` should not be querying any embedding itself. But also overriding the env var like this could interfere pretty badly with unrelated parts of the codebase. Looked around and did not find a way to pass it in via the constructor of `VectorStoreIndex`.

Should I just set the real key?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Now with the extra idempotence check (`get_nodes_by_ids` to see if the records already exist), it is a lot slower.

I added another strategic `await asyncio.sleep(0)` to avoid a case where all nodes already exist and we keep checking without ever doing any (real) async at all

libs/server/kiln_server/document_api.py
```diff
@@ -1061,13 +1096,30 @@ async def create_rag_config(
                 detail=f"Embedding config {request.embedding_config_id} not found",
             )
 
+        # TODO: get vector store config params from the request
```
leonardmq, line 1136: Separate PR to add UI for these parameters - hardcoded to FTS config for now

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Currently in terms of batching, it is set to be:
- 20 documents in memory in the `RagIndexingStepRunner` - kept it low because each document could be super long and have many chunks
- we pass the list of the 20 documents and their parts as `[(doc_id, embedding[], chunk[]), ...]` to the `LanceDBAdapter.add_chunks_with_embeddings(tuples)`, and in there we insert batches of `100` nodes (chunks)

These numbers are currently relatively arbitrary - could tune a bit more, but probably best to keep batches small as memory seems to build up quickly

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -268,7 +269,8 @@ function sortRagConfigs(
 
 function calculateStatus(progress: RagProgress): RagConfigurationStatus {
   if (
-    progress.total_document_completed_count === progress.total_document_count
+    progress.total_document_completed_count === progress.total_document_count &&
+    progress.total_chunk_completed_count === progress.total_chunk_count
   ) {
```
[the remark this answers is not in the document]
leonardmq: `total_chunk_completed_count` includes substeps pertaining to chunk (that includes indexing), so `rogress.total_chunk_completed_count === progress.total_chunk_count` is true when indexing is completed

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte
```diff
@@ -87,6 +97,15 @@
...
 
+  $: total_chunks = config_progress?.total_chunk_count || 0
+  $: chunks_completed_pct =
+    total_chunks > 0
+      ? Math.round(
+          ((config_progress?.total_chunk_completed_count || 0) / total_chunks) *
+            100,
+        )
+      : 0
+
```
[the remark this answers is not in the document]
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/run_rag_dialog.svelte:108]: There are document-specific steps (extract, chunk. embed) and chunk-specific ones (indexing). `chunks_completed_pct` is the percentage of completion of everything related to chunks, including indexing; we have the same thing for `docs_completed_pct` as well

app/web_ui/src/lib/api_schema.d.ts
```diff
@@ -3335,6 +3312,11 @@ export interface components {
              * @description The ID of the embedding config used to embed the documents.
              */
             embedding_config_id: string | null;
+            /**
+             * Vector Store Config Id
+             * @description The ID of the vector store config that was used to store the documents.
+             */
+            vector_store_config_id: string | null;
             /** Model Type */
```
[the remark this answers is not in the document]
leonardmq: This is codegen

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
```diff
@@ -0,0 +1,54 @@
...
+class KilnVectorStoreQuery(BaseModel):
+    query_string: Optional[str] = Field(
+        description="The query string to search for.",
+        default=None,
+    )
+    query_embedding: Optional[List[float]] = Field(
+        description="The embedding of the query.",
+        default=None,
+    )
+
```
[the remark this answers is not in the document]
leonardmq: What is a valid query depends on the vector store being used, would rather keep validation elsewhere at the store level (as it currently is; in `build_kwargs_for_query`)

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,311 @@
...
+                # free up event loop to avoid risk of looping for a long time
+                # without any real async ops releasing the event loop at all
+                # (get_nodes_by_ids implementation in llama_index is actually sync
+                # and it is slow)
+                await asyncio.sleep(0)
+                continue
+            else:
+                # otherwise, we just insert all the chunks for this document again
+                pass
+
```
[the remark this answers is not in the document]
leonardmq: A little bruteforce, but each document would only have a few dozens chunks so overwriting is relatively cheap. The thinking here is that if a doc has only been partially inserted (say, 10 chunks, when the incoming doc has 15 chunks), we assume something went wrong - could also be the result of a conflicting document processing incoming via a foreign branch or something like that. That would be very obscure if it were to happen, so bruteforce helps remove much of that concern

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,311 @@
...
+    async def count_records(self) -> int:
+        try:
+            table = self.lancedb_vector_store.table
+            if table is None:
+                raise ValueError("Table is not initialized")
+            count = table.count_rows()
+            return count
+        except TableNotFoundError:
+            return 0
+
```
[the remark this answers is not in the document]
leonardmq: This check is mostly for the typechecker. If it were to actually fire, that would likely be an actual problem. If the table does not exist, the `TableNotFoundError` exception should kick in (and we catch that one because we know why it is happening)

libs/core/kiln_ai/adapters/vector_store/test_vector_store_registry.py
```diff
@@ -0,0 +1,191 @@
...
+    """Create a temporary database path for testing."""
+    db_path = tempfile.mkdtemp(suffix=".lancedb")
+    yield db_path
+    # Cleanup
+    if os.path.exists(db_path):
+        import shutil
+
+        shutil.rmtree(db_path)
+
+
```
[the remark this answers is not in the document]
leonardmq: How does this work? Does `autouse=True` inject the fixture into every test (and every other fixtures) in this file?

libs/core/kiln_ai/datamodel/rag.py
```diff
@@ -24,3 +29,13 @@ class RagConfig(KilnParentedModel):
     embedding_config_id: ID_TYPE = Field(
         description="The ID of the embedding config used to embed the documents.",
     )
+
+    vector_store_config_id: ID_TYPE = Field(
+        description="The ID of the vector store config used to store the documents.",
+    )
```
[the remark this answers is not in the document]
leonardmq: `vector_store_config_id` is required in every RagConfig; similar to `extractor_config_id`, `chunker_config_id`, etc. If some dependent code breaks because of this not being nullable, that code must be fixed, not this made optional

libs/core/pyproject.toml
```diff
@@ -23,16 +23,22 @@
...
     "jsonschema>=4.23.0",
+    "llama-index-core>=0.12.42",
+    "lancedb>=0.24.2",
     "litellm>=1.72.6",
     "openai>=1.53.0",
     "pdoc>=15.0.0",
+    "pyarrow>=19.0.1",
     "pydantic>=2.9.2",
+    "pylance>=0.32.1",
     "pytest-benchmark>=5.1.0",
```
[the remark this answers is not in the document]
leonardmq: `pylance` is not the VS Code extension of the same name, it is a LanceDB package: https://pypi.org/project/pylance/

If it is already getting pulled by the lance package, maybe we can remove it here

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+
+        return self._index
+
+    async def delete_nodes_by_document_id(self, document_id: str) -> None:
+        # higher level operation that requires ref_doc_id to be set on the nodes
+        # which is set through the source node relationship
+        self.index.delete_ref_doc(document_id)
+
+    async def get_nodes_by_ids(self, node_ids: List[str]) -> List[BaseNode]:
```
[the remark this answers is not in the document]
leonardmq: Would rather let that one bubble up as that might be a sign that operations are not done in the correct order on the caller side

libs/core/kiln_ai/datamodel/vector_store.py
```diff
@@ -0,0 +1,105 @@
...
+class VectorStoreConfig(KilnParentedModel):
+    name: FilenameString = Field(
+        description="A name for your own reference to identify the vector store config.",
+    )
+    store_type: VectorStoreType = Field(
+        description="The type of vector store to use.",
+    )
+    properties: dict[str, str | int | float] = Field(
+        description="The properties of the vector store config, specific to the selected store_type.",
+    )
```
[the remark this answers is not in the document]
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/487/commits/9415a5879363efee55a0767e28d6d4e588d59a2f

libs/core/kiln_ai/adapters/vector_store/test_vector_store_registry.py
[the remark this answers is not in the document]
leonardmq: Nice, addressed in https://github.com/Kiln-AI/Kiln/pull/487/commits/58454cff326aa1960d00f32133f2c770b65e6168 - applied same logic to `test_lancedb_adapter.py` too

libs/core/kiln_ai/adapters/vector_store/test_base_vector_store.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class TestBaseVectorStoreAdapter:
+    """Test the base vector store adapter abstract class."""
+
+    def test_cannot_instantiate_abstract_class(self):
+        """Test that the abstract base class cannot be instantiated."""
+        with pytest.raises(TypeError):
+            BaseVectorStoreAdapter(MagicMock(spec=VectorStoreConfig))
+
```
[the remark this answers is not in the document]
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/487/commits/9415a5879363efee55a0767e28d6d4e588d59a2f

libs/core/kiln_ai/adapters/rag/rag_runners.py
```diff
@@ -519,27 +690,60 @@ def update_workflow_progress(
...
+                    self.current_progress.total_chunks_indexed_count = (
+                        self.current_progress.total_chunks_indexed_count
+                        + step_progress.success_count
+                    )
+                if step_progress.error_count is not None:
+                    self.current_progress.total_chunks_indexed_error_count = max(
+                        self.current_progress.total_chunks_indexed_error_count,
+                        step_progress.error_count
+                        + self.initial_progress.total_chunks_indexed_error_count,
+                    )
```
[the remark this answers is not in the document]
leonardmq: This is aligned with the other error count calculations

libs/core/pyproject.toml
```diff
@@ -23,16 +23,21 @@ dependencies = [
     "google-cloud-aiplatform>=1.84.0",
     "google-genai>=1.21.1",
     "jsonschema>=4.23.0",
+    "lancedb>=0.24.2",
     "litellm>=1.72.6",
     "openai>=1.53.0",
     "pdoc>=15.0.0",
+    "pyarrow>=19.0.1",
```
scosman: do we need an explicit dependency? I assume lancedb depends on it and we don't need to list it here?

libs/core/kiln_ai/utils/uuid.py
```diff
@@ -0,0 +1,8 @@
+import uuid
+
+KILN_UUID_NAMESPACE = uuid.NAMESPACE_DNS
```
scosman: Nit: technically the strings aren't domains, so should use a custom UUID here. Not that important.

libs/core/kiln_ai/datamodel/test_vector_store.py
```diff
@@ -0,0 +1,320 @@
...
+        self, name, mock_vector_store_fts_config_properties
+    ):
+        """Test VectorStoreConfig accepts valid names."""
+        config = VectorStoreConfig(
+            name=name,
+            store_type=VectorStoreType.LANCE_DB_FTS,
+            properties=mock_vector_store_fts_config_properties,
+        )
+        assert config.name == name
+
```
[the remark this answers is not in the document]
leonardmq: This does not seem accurate. We allow spaces now

libs/server/kiln_server/test_document_api.py
```diff
@@ -1755,3 +1794,146 @@ async def mock_run():
...
+            if missing_sub_config_type == "vector_store_config"
+            else mock_vector_store_config
+        )
+
+        with pytest.raises(HTTPException) as exc_info:
+            await build_rag_workflow_runner(mock_project, str(rag_config.id))
+
+        assert exc_info.value.status_code == 404
+        assert error_message in exc_info.value.detail
+
```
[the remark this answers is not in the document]
leonardmq: `@pytest.mark.asyncio` is optional (check out `pytest.ini` where we set it to auto)

libs/core/kiln_ai/datamodel/test_vector_store.py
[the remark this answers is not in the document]
leonardmq: For example, check out this test: https://github.com/Kiln-AI/Kiln/blob/68e3b7983b24f20834f059cc2ba0d69118f6e7f5/libs/core/kiln_ai/datamodel/test_basemodel.py#L337

libs/core/kiln_ai/utils/validation.py
```diff
@@ -0,0 +1,48 @@
+from typing import Any, TypeVar, Union
+
+T = TypeVar("T")
+
+
+def validate_return_dict_prop(
+    dict: dict[str, Any], key: str, type: type[T], optional: bool = False
```
scosman: optional parameter not checked, can remove.

Did code rabbit run? This is his job!

libs/core/kiln_ai/utils/validation.py
```diff
@@ -0,0 +1,48 @@
...
+    Raises:
+        ValueError: If the property is not found or is not of the specified type.
+
+    Example:
+        >>> validate_return_dict_prop({"key": "value"}, "key", str)
+        "value"
+    """
+    if key not in dict or not isinstance(dict[key], type):
+        raise ValueError(
+            f"{key} is a required property for LanceDB vector store configs and must be of type {type}"
```
scosman: string is way to specific for a general utils/validator.py. Move into the class where they are used, or just generalize the error string.

libs/core/kiln_ai/adapters/vector_store/vector_store_registry.py
```diff
@@ -0,0 +1,39 @@
...
+async def vector_store_adapter_for_config(
+    rag_config: RagConfig,
+    vector_store_config: VectorStoreConfig,
+) -> BaseVectorStoreAdapter:
+    vector_store_config_id = vector_store_config.id
+    if vector_store_config_id is None:
+        raise ValueError("Vector store config ID is required")
+
+    match vector_store_config.store_type:
+        case VectorStoreType.LANCE_DB_FTS:
```
scosman: nit: can use `VectorStoreType.LANCE_DB_FTS | VectorStoreType.LANCE_DB_HYBRID |VectorStoreType.LANCE_DB_VECTOR` here I think

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
scosman: I'd say we import theirs like `import .....VectorStoreQuery as LlamaIndexVectorStoreQuery` and keep our names clean. Namespaces should solve this, not names. If we use both in the same class we can import `as KilnVectorStoreQuery` for clarity

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,331 @@
...
+            vector_store=self.lancedb_vector_store
+        )
+
+        # FIXME:
+        # embed_model=None should be valid and result in llama_index initializing it
+        # to a mock (like it does elsewhere) but there is a fallback in the
+        # VectorStoreIndex constructor that overrides None with "default"
+        # and tries to load OpenAI and throws if the OPENAI_API_KEY is not set
+        # maybe should open an issue on their repo
+        if "OPENAI_API_KEY" not in os.environ:
```
scosman: Oh, yeah we need to fix this

1) we should never give it a valid key. We don't want it calling API, so don't let if have a real key. If we do need to hack around a bug, give them a fake key always.
2) We don't want to set os.environ permanently. Config uses that key. Other might too (we're a library). We don't get to muck with os.environ.
3) Should file a bug too and link it here

Can we just set it during the constructor? if so can use a context manager:

```
@contextmanager
def temporary_env(var, value):
    old_value = os.environ.get(var)
    os.environ[var] = value
    try:
        yield
    finally:
        if old_value is None:
            os.environ.pop(var, None)  # remove if it didn’t exist before
        else:
            os.environ[var] = old_value

...

with temporary_env("OPENAI_API_KEY", "dummy"):
  self._index = VectorStoreIndex( ...
```

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
scosman: maybe just put it on the first line of the loop. Don't need to worry about early returns/continue. Someone editing this later won't think about this.

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,331 @@
...
+                if len(node_batch) >= nodes_batch_size:
+                    # async_add is currently not async, LanceDB has an async API but
+                    # llama_index does not use it, so it is synchronous and blocking
+                    # avoid calling with too many nodes at once
+                    await self.lancedb_vector_store.async_add(node_batch)
+                    node_batch.clear()
+
+            await asyncio.sleep(0)
+
+        if node_batch:
```
scosman, line 225: this trailing one needed?

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
```diff
@@ -0,0 +1,331 @@
...
+                SearchResult(
+                    document_id=document_id,
+                    chunk_idx=chunk_idx,
+                    chunk_text=node.get_content(),
+                    similarity=similarity,
+                )
+            )
+        return results
+
+    def build_kwargs_for_query(self, query: KilnVectorStoreQuery) -> Dict[str, Any]:
```
scosman: Nit: no any. even if union, typing is good.

libs/core/kiln_ai/adapters/rag/rag_runners.py
```diff
@@ -428,16 +481,165 @@ async def run(self) -> AsyncGenerator[RagStepRunnerProgress, None]:
...
+            ):
+                chunk_count = 0
+                for record in records:
+                    chunk_count += len(record[1].chunks)
+
+                try:
+                    await vector_store.add_chunks_with_embeddings(records)
+                    indexed_count += chunk_count
+                    yield RagStepRunnerProgress(
+                        success_count=chunk_count,
```
scosman, line 617: shouldn't success be indexed_count and error count be tracked?

libs/core/kiln_ai/adapters/rag/rag_runners.py
```diff
@@ -428,16 +481,165 @@ async def run(self) -> AsyncGenerator[RagStepRunnerProgress, None]:
...
+                self.vector_store_config,
+            )
+
+            yield RagStepRunnerProgress(
+                success_count=0,
+                error_count=0,
+            )
+
+            indexed_count = 0
+            async for records in self.collect_records(
```
scosman: can you use `value, doc, embedding` here instead of `record[1]` tuple indexes? Would make the code much more readable. Even if it has to go on new lines (but I think you can use inline assignment)

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/create_eval_config/+page.svelte
```diff
@@ -165,9 +165,7 @@
             // @ts-expect-error provider is not typed, but server will validate
             provider: provider_name,
             properties: {
-              // @ts-expect-error properties are not typed, but server will validate
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/create_eval_config/+page.svelte:168]: All these type changes are strange. What makes them required on your branch but not main?

Comment (leonardmq) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/create_eval_config/+page.svelte:168]: This is the svelte-check thing likely due to dependency changes that we discussed before, and I still don't know the cause - I tried to narrow it down but still don't know what caused the change. I suspect the change was due to dependencies on the backend.

To clarify - I removed these `@ts-expect-error` because svelte check now says the types are already correct and the compiler bypass comments are no longer needed. Other branches (including `main`) needed these bypasses because types were not agreeing and now apparently they do.

In the API schema codegen, the changes are this:
```ts
// some types used to be like this:
Record<string, never>

// now they are like this:
{ [key: string]: unknown }
```

libs/core/kiln_ai/adapters/rag/rag_runners.py
leonardmq: `records` here is `list[Tuple[str, ChunkedDocument, ChunkEmbeddings]]` (a whole batch of documents and their associated chunks / vectors), not just a one document tuple so we cannot unpack it like that.

Alternatively, we could make a pydantic model / dataclass out of it (instead of tuple), would make it more obvious and I would imagine not much of a performance cost here

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Yes, if `total % nodes_batch_size != 0` then at the end of the loop there will be an incomplete batch that has not been flushed

libs/core/kiln_ai/adapters/rag/rag_runners.py
```diff
@@ -428,16 +481,163 @@
...
+            yield jobs
+            jobs.clear()
+
+    async def count_total_chunks(self) -> int:
+        total_chunk_count = 0
+        async for records in self.collect_records(batch_size=1):
+            _, chunked_doc, _ = records[0]
+            total_chunk_count += len(chunked_doc.chunks)
+        return total_chunk_count
+
```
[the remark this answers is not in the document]
leonardmq: We are counting the total number of existing chunks, that is unrelated to the document_ids filtering which controls whether we should only process the subset of documents specified in that array

libs/core/kiln_ai/adapters/rag/rag_runners.py
leonardmq: Removed the `indexed_count` - the events coming out of the step are incremental (and accumulated in the orchestrator).

https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Added context manager in https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/adapters/vector_store/base_vector_store_adapter.py
leonardmq: Renamed in https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Moved to first line inside the loop (and removed the other ones): https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/utils/validation.py
leonardmq: Whoops, forgot to remove that one - I wrote a separate method for the optional, but initially tried to make it a flag but that was too convoluted with the generics

Removed the flag arg in: https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/utils/uuid.py
leonardmq: Refactored in https://github.com/Kiln-AI/Kiln/pull/487/commits/1d7000cbc115ed40e6846d1d3793616531a8d29b

Looks like this now:
```py
import uuid

# do not change this, or this will break backwards compatibility with existing UUIDs
KILN_UUID_NAMESPACE = uuid.uuid5(uuid.NAMESPACE_DNS, "kiln.tech")


def string_to_uuid(s: str) -> uuid.UUID:
    """Return a deterministic UUIDv5 for the input string."""
    return uuid.uuid5(KILN_UUID_NAMESPACE, s)
```

libs/core/pyproject.toml
leonardmq: It does not seem to be needed - I removed and then ran `uv sync` again and it works fine. I suspect I might have needed this before when we still had the dependency override preventing it from being installed in the root `pyproject.toml`.

Removed in https://github.com/Kiln-AI/Kiln/commit/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/utils/validation.py
leonardmq: Indeed 😐  - Added an arg to pass in a message prefix in https://github.com/Kiln-AI/Kiln/commit/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/adapters/vector_store/vector_store_registry.py
leonardmq: Updated in https://github.com/Kiln-AI/Kiln/commit/1d7000cbc115ed40e6846d1d3793616531a8d29b

libs/core/kiln_ai/adapters/vector_store/lancedb_adapter.py
leonardmq: Union does not work. 

Doing `Dict[str, List[float] | str]` throws a typing error where it is getting unpacked because it does not know which key is which type:
```py
query_result = await self.lancedb_vector_store.aquery(
            LlamaIndexVectorStoreQuery(
                **self.build_kwargs_for_query(query), # here
            ),
            query_type=self.query_type,
        )
```

I refactored to use a `TypedDict` in https://github.com/Kiln-AI/Kiln/pull/487/commits/0a772beb4aa3b1a993f22da37ce99e0cb46c34e9 - and that seems to work

libs/core/kiln_ai/adapters/rag/rag_runners.py
leonardmq: Refactored to bundle them into a `dataclass` instead of a tuple: https://github.com/Kiln-AI/Kiln/pull/487/commits/fb8253517f5332dd70c3a43668c72e2de0a66423

PR #490, opened 2025-08-08
Add basic UI to allow external tool management

aiyakitori:
## What does this PR do?

Add basic UI to allow user to connect external tools under settings. 
* Add "Manage Tools" page under settings
* Add "Detail Tool" page based on tool id
* Add "Add Tool" page. 
* Update connect_remote_MCP API to allow empty headers.
* Add new API `get_tool` which takes a tool id and return the corresponding tool. 

## Checklists

- [YES] New tests have been added to any work in /lib


app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
+<script lang="ts">
+  import AppPage from "../../../../app_page.svelte"
+  import FormContainer from "$lib/utils/form_container.svelte"
+  import FormElement from "$lib/utils/form_element.svelte"
+  import { client } from "$lib/api_client"
+  import { page } from "$app/stores"
+  import { goto } from "$app/navigation"
+  import { KilnError, createKilnError } from "$lib/utils/error_handlers"
+
+  $: project_id = $page.params.project_id
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: nit: this works fine but `$:` is special svelte syntax that makes something reactive (automatically updates, and can update UI automatically). You don't need that here. Since you're using it in functions, you can just use `$page.params.project_id` where you need the value, or `const project_id = $page.params.project_id` inside the method if you want a short form. There's a (tiny) cost to reactive vars.

If you have svelte UI rendering it, or a `{#if` using it, you'd need the reactive version.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+  let server_url = ""
+  let description = ""
+
+  // Headers as array of key/value pairs
+  interface HeaderPair {
+    key: string
+    value: string
+  }
+
+  let headers: HeaderPair[] = [{ key: "", value: "" }]
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: start empty? Most server won't need it, and will probably just cause validation errors?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        <button
+          type="button"
+          class="btn btn-outline btn-sm"
+          on:click={addHeader}
+        >
+          + Add Header
+        </button>
+      </div>
+
+      <FormElement
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: put description under name? Feels like we should keep metadata, and server data together?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        label="Description"
+        inputType="textarea"
+        id="mcp_description"
+        description="Optional description of what this tool does"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Usually do `A description of this MCP server for your reference.`

"Optional" is already added by the FormElement so no need to repeat it.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+      }
+    } catch (e) {
+      error = createKilnError(e)
+    } finally {
+      submitting = false
+    }
+  }
+</script>
+
+<AppPage title="Add Tools">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Maybe: "Add Remote MCP Tools" to make this more specific?

Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: And note for you later: have a `Read the Docs` link once we have docs. We always try to link to the docs in app where we can.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+  <div class="max-w-2xl">
+    <div class="mb-6">
+      <h2 class="text-xl font-semibold mb-2">Connect Remote MCP Server</h2>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: just use this as title

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+      <p class="text-sm text-gray-600">
+        Connect to a remote Model Context Protocol (MCP) server to add external
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Move this string to AppPage.subtitle for standard styling and can remove here.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        tools to your project.
+      </p>
+    </div>
+
+    <FormContainer
+      submit_label="Connect Remote MCP"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Nit: just "Connect" since title has info.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        {
+          params: {
+            path: {
+              project_id,
+            },
+          },
+          body: {
+            name: name.trim(),
+            server_url: server_url.trim(),
+            headers: headersObj as unknown as Record<string, never>, // Type assertion to work around schema issue
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: We should do it the right way. We shouldn't have any "as" or "unknown" anywhere. 

But your function returns the right type? is this even needed?

If type workarounds are absolutely necessary (very rarel) use comments to ignore not "as".

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+      submitting = false
+    }
+  }
+</script>
+
+<AppPage title="Add Tools">
+  <div class="max-w-2xl">
+    <div class="mb-6">
+      <h2 class="text-xl font-semibold mb-2">Connect Remote MCP Server</h2>
+      <p class="text-sm text-gray-600">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: our official grey is `text-grey-500`. Small thing to be consistent one. I should name it so we have a firm name. I'll add a `design_guide.md` for agents too.

And oddly: I like 600 more, but still prefer consistent to mismatch.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        optional={true}
+        bind:value={headers}
+      />
+
+      <div class="space-y-3">
+        {#each headers as header, index}
+          <div class="flex gap-2 items-center">
+            <input
+              type="text"
+              placeholder="Header name (e.g., Authorization)"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: design nit: repeated placeholders will get ugly. Stealing the design from create task should fix it.

For future reference: Look at `<InfoTooltip>` for a nice way to put a tooltip with more info. Column headers can be short "Header Name" and. "Value" but you have have the more wordy details/example in the tooltip.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+        placeholder="https://example.com/mcp"
+        bind:value={server_url}
+      />
+
+      <!-- Headers section -->
+      <FormElement
+        inputType="header_only"
+        label="Headers"
+        id="headers"
+        description="Add authentication headers or other required headers for the MCP server"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Design for a very basic audience. Don't assume they will know what headers are, and assume a field that confuses them will trip them up can cause them to churn. I try to hide all "advanced" options in a "Advanced" collapse although I think stings are enough here. 

Maybe: "Optional. If the documentation for the server you're adding requires custom headers, enter them here."
Tooltip (info_text in the FormElement control): "These are usually not needed. Some MCP servers require custom headers, such as the 'Authorization' headers. Refer to the documentation for the server you're adding to see if they require headers."

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+    if (headers.length > 1) {
+      headers = headers.filter((_, i) => i !== index)
+    }
+  }
+
+  function buildHeadersObject(): Record<string, string> {
+    const headersObj: Record<string, string> = {}
+
+    for (const header of headers) {
+      if (header.key.trim() && header.value.trim()) {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:32]: logic wise: they could have a Header name filled out, and it will just get dropped, which kinda feels like a bug.

I'd suggest:

 - start with no headers, not an empty pair. Fewer text boxes is good for novice users. Most MCP servers don't need them
 - Try replacing the <inputs> with <FormElement> inputs leaving the default optional=false and setting hide_label=true. You should get nicer inline error UI if they leave something blank (red outline) forcing them to remove it or fix it. Not 100% sure hide_label will work here, but worth a shot.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+            />
+            <input
+              type="text"
+              placeholder="Header value (e.g., Bearer token-here)"
+              class="input input-bordered flex-1"
+              bind:value={header.value}
+            />
+            <button
+              type="button"
+              class="btn btn-error btn-sm"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: save btn-error fore scary destructive actions. Not normal UI elements. Can do "hover:btn-error" effects for red when hovering, but it's too loud all the time.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+      <FormElement
+        inputType="header_only"
+        label="Headers"
+        id="headers"
+        description="Add authentication headers or other required headers for the MCP server"
+        optional={true}
+        bind:value={headers}
+      />
+
+      <div class="space-y-3">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Always copy an existing design if you can. Keeps the whole app feeling more uniform.

We have this UI style defined in "Edit Task" "Part 2: Requirements". Go steal that layout.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,181 @@
...
+
+      <!-- Headers section -->
+      <FormElement
+        inputType="header_only"
+        label="Headers"
+        id="headers"
+        description="Add authentication headers or other required headers for the MCP server"
+        optional={true}
+        bind:value={headers}
+      />
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:145]: you're getting an ugly space after the header before the content. It's because the FormContainer puts an even amount of space between each child (usually nice), but the header and content are split into 2 children here. Just put both in a shared div, and then there's only 1 child of FormContainer and things will be better.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+                <td class="text-sm">
+                  {tool.description || "No description available"}
+                </td>
+              </tr>
+            {/each}
+          </tbody>
+        </table>
+      </div>
+    {:else}
+      <div class="font-light text-gray-500 text-sm">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: We have a nice control for this. `<EmptyIntro>`

Always look around at existing designs and use those before writing a new one.

Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: once you add it, we'll have 2 "Add Tool" buttons on the same screen (no good). You can make the other one not appear if this does (we do it elsewhere, copy that pattern)

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+<div class="max-w-[1400px]">
+  <AppPage
+    title="Tools"
+    sub_subtitle="Connect your projects to tools like remote MCP servers, Kiln built-in tools, and more"
+    action_buttons={[
+      {
+        label: "Add Tool",
+        href: `/settings/manage_tools/${project_id}/add_tools`,
+        primary: true,
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Convention: You're only allowed one blue "primary" button per screen, and it should always be main function of the screen. This is a button up in the top right, not the primary action. Keep it default grey.

I probably break my own rule sometimes, fix if you see it! 😆

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+    if (tool.id) {
+      goto(`/settings/manage_tools/${project_id}/tools/${tool.id}`)
+    }
+  }
+</script>
+
+<div class="max-w-[1400px]">
+  <AppPage
+    title="Tools"
+    sub_subtitle="Connect your projects to tools like remote MCP servers, Kiln built-in tools, and more"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: just use subtitle. sub_subtitle is only I you already used subtle 😆 .

Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Write the strings for "now". There's no Kiln built in tools or "and more". This is the kinda bug we'll never catch. "Connect your project to tools with MCP servers".

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+    action_buttons={[
+      {
+        label: "Add Tool",
+        href: `/settings/manage_tools/${project_id}/add_tools`,
+        primary: true,
+      },
+    ]}
+  >
+    <div class="font-medium mt-8">Available Tools</div>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: The title of the page is already tools. No need to repeat yourself

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+          {error.getMessage() || "An unknown error occurred"}
+        </div>
+      </div>
+    {:else if tools && tools.length > 0}
+      <div class="overflow-x-auto rounded-lg border mt-4">
+        <table class="table">
+          <thead>
+            <tr>
+              <th>Name</th>
+              <th>ID</th>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: ID is too much detail for the screen. They never need to know the ID.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,126 @@
...
+              <tr
+                class="hover:bg-base-200 cursor-pointer"
+                on:click={() => navigateToTool(tool)}
+                on:keydown={(e) => e.key === "Enter" && navigateToTool(tool)}
+                role="button"
+                tabindex="0"
+              >
+                <td class="font-medium">{tool.name}</td>
+                <td class="text-sm text-gray-500">{tool.id}</td>
+                <td class="text-sm">{tool.type}</td>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Add a friendly name lookup in formatters.ts, using the same trick for enum exhaustive

app/web_ui/src/routes/(app)/settings/+page.svelte
```diff
@@ -58,6 +58,18 @@
         },
       ],
     },
+    {
+      category: "Tools",
+      items: [
+        {
+          name: "Manage Tools",
+          description:
+            "Connect your projects to tools like remote MCP servers, Kiln built-in tools, and more",
```
scosman: ditto.

app/web_ui/src/routes/(app)/settings/+page.svelte
```diff
@@ -71,7 +83,7 @@
       ],
     },
     {
-      category: "Tools & Support",
+      category: "Support",
```
scosman: nothing in this section is "Support".... not sure what to call it though...

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte
```diff
@@ -0,0 +1,178 @@
...
+  }
+</script>
+
+<div class="max-w-[1400px]">
+  <AppPage
+    title={tool?.name || "Tool Details"}
+    subtitle="Tool Information"
+    action_buttons={[
+      {
+        label: "Back to Tools",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]: Browsers have back buttons, apps don't need to.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  function goBack() {
+    goto(`/settings/manage_tools/${project_id}`)
+  }
+</script>
+
+<div class="max-w-[1400px]">
+  <AppPage
+    title={tool?.name || "Tool Details"}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]: Always make page titles what the page is, not the entity name. They could name their tool anything, which could be wildly confusing. People tend to write 💩 names.

Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]: Also: this will be "flashy". It will say one thing for a few milliseconds, then another when it loads. Avoid flashes. Hide UI until loaded behind a spinner.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte
```diff
@@ -0,0 +1,178 @@
...
+        primary: false,
+      },
+    ]}
+  >
+    {#if loading}
+      <div class="w-full min-h-[50vh] flex justify-center items-center">
+        <div class="loading loading-spinner loading-lg"></div>
+      </div>
+    {:else if error}
+      <div
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte:123]: So for this screen: take another pass using an existing good screen in the app as reference like you did for the other two. Usually there are controls you can reuse. This one's style feels out of place in the app. `<PropertyList>` mostly.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.ts
```diff
@@ -0,0 +1,5 @@
+export const prerender = false
+
+export async function load() {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.ts]: Shouldn't need load, and might mess up static builds.

scosman (commented):
Lots of little comments. Code is mostly good, just lots of pointers, controls, style ideas and strings.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]:
```
      </div>
```
There is a loading spinner for fetch_tool()

app/web_ui/src/routes/(app)/settings/+page.svelte
aiyakitori: I was struggling to come up with a name as well...

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Updated schema to be `Dict[str: str]` which matches the header spec and would get rid of the "as" workaround.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Added empty_tools component.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/empty_tools.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/empty_tools.svelte:1]: Oh there's a new component for this `<Intro>` in intro.ts. Use that. Not all older ones are migrated to it, but we should use it going forward.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/empty_tools.svelte
```diff
@@ -0,0 +1,32 @@
...
+          stroke="currentColor"
+          stroke-width="1.5"
+        />
+      </svg>
+    </div>
+    <div class="font-medium text-lg">No tools found for this project.</div>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/empty_tools.svelte]: Don't treat empty like an error, treat it like onboarding a fresh user. 100% of users will get this message, it should be encouraging/hopeful.

"Add Tools to Your Project"
"Tools can add powerful capabilities to your AI tasks." "Add MCP servers to give you system new skills."

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -0,0 +1,121 @@
...
+                <td class="text-sm">
+                  {tool.description || "No description available"}
+                </td>
+              </tr>
+            {/each}
+          </tbody>
+        </table>
+      </div>
+    {:else if is_empty}
+      <EmptyTools {project_id} />
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: vertical position needs work. Look at what we do for the other ones.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+        tools to your project."
+>
+  <div class="max-w-2xl">
+    <FormContainer
+      submit_label="Connect"
+      on:submit={connect_remote_mcp}
+      bind:error
+      bind:submitting
+    >
+      <FormElement label="Name" id="name" bind:value={name} max_length={120} />
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: needs a description. "A name to identify this MCP server."

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+        id="mcp_description"
+        description="A description of this MCP server for your reference."
+        optional={true}
+        bind:value={description}
+      />
+
+      <FormElement
+        label="Server URL"
+        id="mcp_server_url"
+        description="The URL of the remote MCP server"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: missing period

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+        placeholder="https://example.com/mcp"
+        bind:value={server_url}
+      />
+
+      <!-- Headers section -->
+      <FormElement
+        inputType="header_only"
+        label="Headers"
+        id="headers_section"
+        description="If the documentation for the server you're adding requires custom headers, enter them here."
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:142]: feels like we should have a info_msg here with more details. I think I typed on in last PR feedback?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+        optional={true}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: optional=false here? Optional label doesn't quite apply to a list UI that allows zero items already.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+        value=""
+      />
+
+      <FormList
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:147]: Feels a little weird when empty. Maybe add an optional empty_msg to FormList, and when populated it shows a string message when empty. For this "No Headers"? Just an idea

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte
```diff
@@ -0,0 +1,153 @@
...
+        value: tool.created_by,
+      })
+    }
+
+    return properties
+  }
+</script>
+
+<div class="max-w-[1400px]">
+  <AppPage title={"Tool Details"}>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]: This page still needs a lot of work to be consistent. Duplicative title "Tool Details" + "Tool Information", properties shouldn't render as JSON (that's an internal API detail, not a user facing thing), we don't put our app pages inside rounded boxes.

Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tools/[tool_id]/+page.svelte]: Generally:
 - Need to match the style of app more closely. Always Look at other pages for examples and follow those. The closest to this one is "Fine Tune" screen - it's a screen just describing an entity. Look at all the similar pages first, and use the design elements matching them. Some are newer/better, so use the best ones 😀 
 - Use standard controls 
 - Really think about what information should be user facing vs what is internal code/details we don't want to expose. Every bit of confusing UI is a chance to confuse users and have them churn. For this one "version" and "model type" are confusing and unnecessary internal details. 
 - Nail the strings. Look at every UI with "new user who doesn't understand the concepts" POV, and also "expert user wanting to get shit done" POV. Should work for both so 1) use short, descriptive good names, 2) use tooltips to explain extra details whenever it's not entirely clear what something is to a noob, 3) no unnecessary UI (datamodel name, data model version).
 - Think though first user experience in detail. What do they want to accomplish? What do they know? What do we want them to know? This note is more for the empty message above than for this page.
 - Never print an enum value or json or dict in the UI. Parse them and map them to english. Exception is JSON model output.
 - Always see how it looks on mobile. 

For this page:
 - Kill the rounded border, we don't use that anywhere. Use our standard margins.
 - put name as subtitle (see fine-tune page)
 - make it 2 columns of property lists
 - First property list is "Properties" and has standard props (id, name, description). Look at other pages for standard order (ID on top).
 - Kill version and model_type, we don't show that anywhere. That's internal logic, not for users. It will confuse them. If we wanted them we'd need tool tips, explaining what they are, but we don't need them.
 - Second is list is properties specific to the "type" in the properties field (oddly we won't title this "Properties" in UI). Probably need custom title and parsing logic per type, for for Remote MCP "Connection Details" with server and headers. URL and headers should be rendered nicely as `<PropertyList>` not json.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -0,0 +1,176 @@
...
+              id="header_value_{item_index}"
+              info_description="The header value, such as 'Bearer your-token-here'"
+              placeholder="Value"
+              light_label={true}
+              bind:value={headers[item_index].value}
+            />
+          </div>
+        </div>
+      </FormList>
+    </FormContainer>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:181]: Can we a separate PR but we should think through editing.

If we just want to allow editing name/descripition, we have a `<EditDialog>` control you can add to the tool details page.

If we want to allow editing everyone, we usually make the this entire page into a control, then use on 2 pages, an add_tool page, and and edit_tool page. Edit tool page will fetch existing values and set them via `export name` style vars in the controls

scosman (changes requested):

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: used same settings as Eval.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:142]: right, missed it earlier!

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:181]: Totally, i was thinking the remove/edit features were all missing.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:147]: image

scosman (approved):
LGTM. Maybe squash merge this one?

PR #495, opened 2025-08-12
Improve the efficenty of our dev server. It's eating my battery!

scosman:
- use watchfiles instead of polling
 - only reload if python files change
 - add 100ms delay to debounce if changing many files


aiyakitori:
The UV lock file has tons of diff due to new format. Should we upgrade to new uv on main before this change?

scosman:
@aiyakitori yeah, I can do that separately, which should cleanup the merge here.

scosman:
@aiyakitori done. Main already had the new format from another merge, but this branch didn't have it. Merging main in resolved the conflict, now change is real change (adding watchfiles)

aiyakitori (approved):

app/desktop/dev_server.py
```diff
@@ -22,4 +22,6 @@
         host="127.0.0.1",
         port=8757,
         reload=True,
+        # Debounce when changing many files (changing branch)
+        reload_delay=0.1,
```
leonardmq, line 26: Is `uvicorn` automatically using `watchfiles` if it is installed (and whatever else it was using until now if it is not installed)?

leonardmq: Seems like it: https://www.uvicorn.org/settings/#reloading-without-watchfiles

PR #500, opened 2025-08-14
Add MCPServer to connect to real server

aiyakitori:
## What does this PR do?

* Add MCPServer to connect to real MCP Server.
* Implement MCP server's `list_tool()` API and update tools page to include available tools. 

image

## Checklists


app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -103,14 +103,22 @@
...
         name: key,
         value: String(value || "N/A"),
       }),
     )
   }
+
+  function getAvailableTools(tool: ExternalToolServerApiDescription) {
+    const properties = tool.available_tools.map((tool: Tool) => ({
+      name: tool.name,
+      value: tool.description || "No description available",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: Just "NA" for not available, especially in tables/list. This text repeated 6 times is ugly.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -138,6 +146,14 @@
             properties={getDetailsProperties(tool_server)}
             title="Properties"
           />
+          {#if getAvailableTools(tool_server).length > 0}
+            <div class="mt-8">
+              <PropertyList
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: Use a table here? 3 columns

 - Tool Name
 - Tool description (see comment on NA)
 - Arguments: format the parameter list. Nothing complex, maybe just a `<ul>`

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: Same goes for parent page.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -138,6 +146,14 @@
             properties={getDetailsProperties(tool_server)}
             title="Properties"
           />
+          {#if getAvailableTools(tool_server).length > 0}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: We should say "This server does not expose any tools" or something like that when zero.

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -21,6 +23,10 @@ class ExternalToolServerCreationRequest(BaseModel):
     description: str | None = None
 
 
+class ExternalToolServerApiDescription(ExternalToolServer):
+    available_tools: list[Tool]
```
scosman: Don't use the mcp package object to define our API. We don't want a 3rd party controlling our API. Plus it needs to be a Pydantic model so I'm surprised it works at all. And we're trying to allow non-mcp externalToolServers in the future.

Make a new `class ToolApiDecription(BaseClass)` with the fields you want. Can add a class method that takes a MCP server and returns a ToolApiDecription.

```
  @classmethod
  def tool_from_mcp_server(cls, mcpTool Tool):
    return ToolApiDecription(
       name=mcpTool.name,
       description=mcpTool.description,
       ...
  )
```

Then call it like
```
ToolApiDecription.tool_from_mcp_server(mcpTool)
```

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -41,20 +47,34 @@ async def get_available_tool_servers(
...
+            mcp_server = MCPServer(tool_server)
+            tools_result = await mcp_server.list_tools()
+            available_tools = tools_result.tools
+
+        return ExternalToolServerApiDescription(
+            name=tool_server.name,
+            type=tool_server.type,
+            description=tool_server.description,
+            properties=tool_server.properties,
+            parent=tool_server.parent,
```
scosman: This is strange. Caller is calling with the parent ID to get this, they should already have the parent. An API returning the parent is generally backwards. Plus if we do want it for some reason, we can beat "parent" for a name. Use the actual type/name (project?). Just remove this.

Edit: you probably just had to do this because extending the disk backed data model. Fix about should help address this.

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -21,6 +23,10 @@ class ExternalToolServerCreationRequest(BaseModel):
     description: str | None = None
 
 
+class ExternalToolServerApiDescription(ExternalToolServer):
```
scosman: I'd remove the inheritance. You're getting fields you don't want, like parent. Just define your API here. It might be similar, but it's not really inheritance.

libs/core/kiln_ai/tools/mcp_server.py
```diff
@@ -0,0 +1,53 @@
...
+    from kiln_ai.tools.mcp_server_tool import MCPServerTool
+
+
+class MCPServer:
+    """
+    A class for interacting with MCP servers.
+    """
+
+    def __init__(self, tool_server: ExternalToolServer):
+        self._tool_server = tool_server
```
scosman: nit: _server_model? This is a tool server, so `tool_server.tool_server` is going to be complicated.

libs/core/kiln_ai/tools/mcp_server.py
```diff
@@ -0,0 +1,53 @@
...
+            write_stream,
+            _,
+        ):
+            # Create a session using the client streams
+            async with ClientSession(read_stream, write_stream) as session:
+                # Initialize the connection
+                await session.initialize()
+                # List available tools
+                self._tools = await session.list_tools()
+                print(f"Available tools: {[tool.name for tool in self._tools.tools]}")
```
scosman: no prints

libs/core/kiln_ai/tools/mcp_server.py
```diff
@@ -0,0 +1,53 @@
...
+                return self._tools
+
+    async def get_tool(self, tool_name: str) -> MCPServerTool:
+        if tool_name not in [tool.name for tool in self._tools.tools]:
+            raise ValueError(f"Tool {tool_name} not found")
+
+        from kiln_ai.tools.mcp_server_tool import MCPServerTool
```
scosman: yeah, need to cleanup the imports/circular, per discussion.

libs/core/kiln_ai/tools/mcp_server.py
```diff
@@ -0,0 +1,53 @@
...
+from mcp import ClientSession, ListToolsResult
+from mcp.client.streamable_http import streamablehttp_client
+
+from kiln_ai.datamodel.external_tool import ExternalToolServer, ToolServerType
+
+if TYPE_CHECKING:
+    from kiln_ai.tools.mcp_server_tool import MCPServerTool
+
+
+class MCPServer:
```
scosman: Looking at this, I think we don't need a wrapper class of the MCP library. Trying to make one is causing all the issues with circular deps, and in the end it is just a 1-line call to the mcp session.

Idea:

 - Kill this class: MCPServer
 - Write MCPSessionMgr (see below)
 - Wherever we were using MCPServer, just do `async with MCPSessionMgr.shared().mcp_client(external_tool_server)  as session:  return await session.list_tools()`. Not enough logic to wrap in a class.
 - Update MCPServerTool constructor to take a ExternalToolModel and tool name. It can use the MCPSessionMgr to do everything internally. No need to create a server to get a tool, just make the MCPServerTool directly.

If we had more logic here it might be difference, but it's pretty much exposing existing mcp calls (list_tools).

```
from mcp import ClientSession

class MCPSessionMgr:
  _shared_instance = None

  @classmethod
  def shared(cls):
    if cls._shared_instance is None:
      cls._shared_instance = cls()
    return cls._shared_instance

  // similar signature to 
  @asynccontextmanager
  async def mcp_client(tool_server ExternalToolServer) -> AsyncGenerator[
    ClientSession,
    None,
  ]:
    // can add a switch on tool_server.type here later
    if tool_server.type != "remote_mcp":
       raise RuntimeError("invalid external tool server type")
    async with streamablehttp_client(tool_server.properties.get("url")) as (
        read_stream,
        write_stream,
        _,
    ):
        # Create a session using the client streams
        async with ClientSession(read_stream, write_stream) as session:
          yield session
```

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,38 @@
...
+class MCPServerTool(KilnTool):
+    def __init__(self, server: MCPServer, tool_name: str):
+        self._server = server
+        self._tool_name = tool_name
+
+    def run(self, **kwargs) -> str:
+        return "Hello, world!"
+
+    async def call_tool(self, **kwargs) -> CallToolResult:
+        server_url = self._server._tool_server.properties.get("server_url")
```
scosman: yeah, all duplicate logic should move into MCPSessionMgr per plan above

app/desktop/studio_server/tool_servers_api.py
aiyakitori: Yep, removed the inheritance removed the confusion.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -157,6 +177,46 @@
...
+                </tr>
+              </thead>
+              <tbody>
+                {#each tool_server.available_tools as tool}
+                  <tr>
+                    <td class="font-medium">{tool.name}</td>
+                    <td>{tool.description || "N/A"}</td>
+                    <td>
+                      {#if formatToolArguments(tool.inputSchema)}
+                        <ul class="list-disc list-inside">
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: The list is a bit ugly, will try refining it.

libs/core/kiln_ai/tools/tool_id.py
```diff
@@ -23,13 +23,28 @@ class KilnBuiltInToolId(str, Enum):
...
 
+    # MCP remote tools must have format: mcp::remote::<server_id>::<tool_name>
+    if id.startswith(MCP_REMOTE_TOOL_ID_PREFIX):
+        # Check server_id and tool_name are not empty and there are exactly 4 parts
+        parts = id.split("::")
+        if len(parts) != 4 or not parts[2] or not parts[3]:
+            raise ValueError(
+                f"Invalid MCP remote tool ID: {id}. Expected format: 'mcp::remote::<server_id>::<tool_name>'."
+            )
+        return id
```
aiyakitori, line 47: Update tool_id to include remote MCP style ids.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -157,6 +177,46 @@
           {/if}
         </div>
       </div>
+      <!-- Row 2: Available Tools full width -->
+      <div class="mb-10">
+        <h3 class="text-xl font-bold mb-4">Available Tools</h3>
+        {#if tool_server.available_tools.length > 0}
+          <div class="overflow-x-auto rounded-lg border">
+            <table class="table">
+              <thead>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte:218]: Header row doesn't match our usual style.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: yeah, one note since I saw it. Make it "None" not "N/A" since that's more accurate. And be consistent on style. NA is grey in some places, black in others.

libs/core/kiln_ai/tools/test_mcp_session_manager.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from kiln_ai.datamodel.external_tool import ExternalToolServer, ToolServerType
+from kiln_ai.tools.mcp_session_manager import MCPSessionManager
+
+
```
scosman, line 8: Add a fixture with a mock which mocks whatever makes the web-request (initialize probably) so we still are testing the code in CI. Sonnet is good at mocks. 

Should have tests for your errors, and success case.

Keep the skipped one for manual work, but add above for CI. Will catch if ExternalTools changes field names and stuff like that.

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -1,12 +1,19 @@
...
 from kiln_ai.datamodel.external_tool import ExternalToolServer, ToolServerType
-from kiln_ai.tools.tool_id import KilnBuiltInToolId, ToolId
+from kiln_ai.tools.mcp_session_manager import MCPSessionManager
+from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, ToolId
 from kiln_server.project_api import project_from_id
+from mcp import Tool
 from pydantic import BaseModel, Field, ValidationError
 
+"""
+This class is used to describe the external tool server under Settings -> Manage Tools UI.
```
scosman: Doc string needs to be inside the class, first line after `class Name...:`. Here it is describing the file, not the class.

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -22,6 +29,43 @@ class ExternalToolServerCreationRequest(BaseModel):
     description: str | None = None
 
 
+"""
+This class is a wrapper of MCP's Tool object to be displayed in the UI under tool_server/[tool_server_id].
```
scosman: Ditto

app/desktop/studio_server/tool_servers_api.py
scosman: @coderabbit confirm the above comment

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -33,29 +77,27 @@ def connect_tool_servers_api(app: FastAPI):
     async def get_available_tools(
         project_id: str,
     ) -> List[ToolApiDescription]:
-        # TODO: add a real implementation of this
```
scosman, line 36: 🫡

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -1,12 +1,19 @@
...
+from typing import Any, Dict, List
 
 from fastapi import FastAPI, HTTPException
 from kiln_ai.datamodel.basemodel import ID_TYPE
 from kiln_ai.datamodel.external_tool import ExternalToolServer, ToolServerType
-from kiln_ai.tools.tool_id import KilnBuiltInToolId, ToolId
+from kiln_ai.tools.mcp_session_manager import MCPSessionManager
+from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, ToolId
 from kiln_server.project_api import project_from_id
+from mcp import Tool
```
scosman: `as McpTool` probably good here. It isn't a general `Tool` in this context, and name might cause bugs

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -1,12 +1,19 @@
...
+from typing import Any, Dict, List
 
 from fastapi import FastAPI, HTTPException
 from kiln_ai.datamodel.basemodel import ID_TYPE
 from kiln_ai.datamodel.external_tool import ExternalToolServer, ToolServerType
-from kiln_ai.tools.tool_id import KilnBuiltInToolId, ToolId
+from kiln_ai.tools.mcp_session_manager import MCPSessionManager
+from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, ToolId
 from kiln_server.project_api import project_from_id
+from mcp.types import Tool
```
scosman: add `as McpTool` to make the class name clearer in this context. Ir just `import mcp` and use `mcp.Tool` inline. `Tool` is a bad name here as it's not really a tool in the tool server context

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -33,29 +77,27 @@ def connect_tool_servers_api(app: FastAPI):
...
-                description="Divide two numbers",
-            ),
-        ]
+        project = project_from_id(project_id)
+
+        # Get available tools from MCP servers
+        available_tools = []
+        for server in project.external_tool_servers(readonly=True):
+            # Get available tools from remote MCP servers only
+            if server.type == ToolServerType.remote_mcp:
```
scosman: use a `match` whenever we are checking enum values, and exhaustive errors for `default` pretty much everywhere. It protects against a ton of bugs, by making the type checker find missing cases when we add new types.

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -76,20 +118,40 @@ async def get_available_tool_servers(
...
             None,
         )
-        if not tool:
+        if not tool_server:
             raise HTTPException(status_code=404, detail="Tool not found")
 
-        return tool
+        # Get available tools based on server type
+        available_tools = []
+        if tool_server.type == ToolServerType.remote_mcp:
```
scosman: Ditto: match+exhaustive not if

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -0,0 +1,50 @@
...
+    @asynccontextmanager
+    async def mcp_client(
+        self,
+        tool_server: ExternalToolServer,
+    ) -> AsyncGenerator[
+        ClientSession,
+        None,
+    ]:
+        # Only support remote MCP servers for now
+        if tool_server.type != "remote_mcp":
```
scosman: Ditto: use match+exhaustive, not if.

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
+import asyncio
+
+from mcp.types import CallToolResult, TextContent, Tool
```
scosman: Ditto on name. Tool -> McpTool or `mcp.Tool`

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
...
+            name=name,
+            description="Not Loaded",
+            parameters_schema={
+                "type": "object",
+                "properties": {},
+            },  # empty object for now, our JSON schema validation will fail if properties are missing
+        )
+
+    def run(self, **kwargs) -> str:
+        result = asyncio.run(self._call_tool(**kwargs))
```
scosman: `asyncio.run` is not ideal. My fault because I made KilnToolInterface sync.

Can we make a separate PR making KilnToolInterface all asynch, and just using await?

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
...
+        if not result.content:
+            return ""
```
scosman: Seems like an error? Tool call didn't return any content isn't normal. I'd raise.

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # raise error if the first block is not a text block
```
scosman, line 52: Don't suppress any errors. If we only support text blocks, and only support exactly 1 then it's a Kiln Error if we get anything else. Always raise for any unsupported case. Entire big projects could fail if we are dropping/ignoring `content[1]` and it contains important info.

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,67 @@
...
+        if not isinstance(result.content[0], TextContent):
+            raise ValueError("First block must be a text block")
+
+        content: TextContent = result.content[0]
+        return str(content.text)
```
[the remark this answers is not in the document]
scosman: Agree 2x:

 - handle the MCP errors, and pass those up. It will probably have a useful error message which will help debugging. Not sure of CR's comment is right, but check.
 - pretty much never use `str`. Check type is string and return it. Raise error if it isn't string. Using str will hide major issues. We had a big one was that str converted a Python object into a string in a field instead of JSON string of the same object.

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class MCPServerTool(KilnTool):
+    def __init__(self, model: ExternalToolServer, name: str):
+        self._tool_server_model = model
+        self._name = name
+        self._tool: Tool | None = None
+
+        #  Some properties are not available until the tool is loaded asynchronously
+        super().__init__(
```
scosman: Just extend KilnToolInterface instead of KilnTool. KilnTool is a base class with some helpers on top of KilnToolInterface, but you're fighting them and not using them. 

That way you can implement name() and description(), and in those load the name with no hacks or invalid state.

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,70 @@
...
+            self._tool_server_model
+        ) as session:
+            result = await session.call_tool(
+                name=self.name(),
+                arguments=kwargs,
+            )
+            return result
+
+    async def _load_tool_properties(self):
+        tool = await self._get_tool(self._name)
```
scosman, line 77: make this a lazy load so it's only called once?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte:218]: talked offline, this is the same table from fine tune.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: UI updated

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -81,18 +81,23 @@ async def get_available_tools(
...
+                                ToolApiDescription(
+                                    id=f"{MCP_REMOTE_TOOL_ID_PREFIX}{server.id}::{tool.name}",
+                                    name=tool.name,
+                                    description=tool.description,
+                                )
+                                for tool in tools_result.tools
+                            ]
+                        )
+                case _:
+                    raise RuntimeError(
```
scosman: this isn't an exhaustive check, just a runtime error. The type checker won't catch it, we'll still miss it if we add a value. Look at `raise_exhaustive_enum_error`

app/desktop/studio_server/tool_servers_api.py
```diff
@@ -131,14 +136,21 @@ async def get_tool_server(
...
+                    tool_server
+                ) as session:
+                    tools_result = await session.list_tools()
+
+                    available_tools = [
+                        ExternalToolApiDescription.tool_from_mcp_tool(tool)
+                        for tool in tools_result.tools
+                    ]
+            case _:
+                raise RuntimeError(
```
scosman: ditto

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -29,8 +29,11 @@ async def mcp_client(
         None,
     ]:
         # Only support remote MCP servers for now
-        if tool_server.type != "remote_mcp":
-            raise RuntimeError("invalid external tool server type")
+        match tool_server.type:
+            case ToolServerType.remote_mcp:
+                pass
+            case _:
+                raise RuntimeError("invalid external tool server type")
```
scosman: ditto

scosman (changes requested):

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -112,23 +112,33 @@
...
-        const isRequired = required.includes(name)
-        const type = (schemaObj.type as string) || "unknown"
-        const description = (schemaObj.description as string) || ""
-        return `${name}${isRequired ? "*" : ""} (${type})${description ? `: ${description}` : ""}`
-      })
-      .join("|||") // Use a delimiter that we can split on later
+    const args: Argument[] = []
+
+    for (const [name, schema] of Object.entries(properties)) {
+      const schemaObj = schema as Record<string, unknown>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: pending chat on type checking

libs/core/kiln_ai/tools/test_mcp_server_tool.py
```diff
@@ -50,7 +50,9 @@ def test_run_success(self, mock_session_manager):
         mock_session_manager.shared.return_value.mcp_client.return_value.__aenter__.return_value = mock_session
 
         result_content = [TextContent(type="text", text="Success result")]
-        call_result = CallToolResult(content=cast(list[ContentBlock], result_content))
+        call_result = CallToolResult(
+            content=cast(list[ContentBlock], result_content), isError=False
```
scosman: no casts. In tests you can use `# type: ignore` comments

libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -22,7 +22,7 @@ async def toolcall_definition(self) -> Dict[str, Any]:
         pass
 
     @abstractmethod
-    async def id(self) -> KilnBuiltInToolId:
+    async def id(self) -> ToolId:
```
scosman, line 25: thank you!

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -1,36 +1,50 @@
...
+from kiln_ai.tools.base_tool import KilnToolInterface
 from kiln_ai.tools.mcp_session_manager import MCPSessionManager
-from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX
+from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, ToolId
 
 
-class MCPServerTool(KilnTool):
+class MCPServerTool(KilnToolInterface):
     def __init__(self, model: ExternalToolServer, name: str):
+        self._tool_id = f"{MCP_REMOTE_TOOL_ID_PREFIX}{model.id}::{name}"
```
scosman: nit: but I try to use data_model since `model` is so common in AI. Doesn't really matter here but generally helps

scosman (changes requested):

libs/core/kiln_ai/tools/mcp_server_tool.py
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+    async def run(self, **kwargs) -> Any:
+        result = await self._call_tool(**kwargs)
+
+        if result.isError:
+            raise ValueError(
+                f"Tool {await self.name()} returned an error: {result.content}"
+            )
+
```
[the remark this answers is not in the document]
aiyakitori: If the function call_tool() or list_tool() already raises, this shouldn't matter?

scosman (approved):

PR #517, opened 2025-08-21
fix: wrong filename in sdk codegen command in cursor rule

leonardmq:
## What does this PR do?

Fix the API client codegen script filename in Cursor rule.

## Summary by CodeRabbit

* **Documentation**
  * Updated contributor instructions for regenerating bindings after Python API changes to reference the correct schema generation script.
  * Clarifies the regeneration process, replacing outdated guidance to reduce confusion and build hiccups during API updates.
  * No impact on application behavior; end users are unaffected.
  * Release artifacts and interfaces remain unchanged.


leonardmq:
Not needed anymore - uses MCP instead

PR #518, opened 2025-08-21
feat: add RAG tool

leonardmq:
## What does this PR do?

Add `RagTool` implementation to allow using RAG Search Tools in UI.

This branch used to be a giant merge between `rag` and `tool` going into `main`, this is what the massive history in the PR is about. But now it is up-to-date with both `main` and `rag` and going into `rag`.

## Checklists

## Summary by CodeRabbit

- **New Features**
  - New "Search Tools (RAG)" support: RAG-based search tools shown in Available Tools and Manage Tools, plus a UI flow to add/configure search tools (templates, create/customize).
  - New project API to list search tools and corresponding frontend types/components.

- **UI**
  - New carousel + pages for suggesting/adding search tools; form fields for tool name/description and supported file types; several layout and label refinements.

- **Validation**
  - Added snake_case tool-name validation for RAG tool metadata.

- **Tests**
  - Expanded test coverage for RAG behavior, ID handling, registry paths, and tool/endpoint integration.


libs/core/kiln_ai/tools/rag_tools.py
```diff
@@ -0,0 +1,153 @@
...
+                ChunkContext(
+                    metadata={
+                        "document_id": knn_result.document_id,
+                        "chunk_idx": knn_result.chunk_idx,
+                    },
+                    text=knn_result.chunk_text,
+                )
+            )
+
+        result = "\n=========\n".join(
```
leonardmq: Need to have another look at llama_index see what format they use by default

leonardmq:
@coderabbitai summary

libs/core/kiln_ai/tools/rag_tools.py
```diff
@@ -0,0 +1,152 @@
...
+from kiln_ai.datamodel.tool_id import ToolId
+from kiln_ai.datamodel.vector_store import VectorStoreConfig, VectorStoreType
+from kiln_ai.tools.base_tool import KilnToolInterface
+
+
+class ChunkContext(BaseModel):
+    metadata: dict
+    text: str
+
+    def serialize(self) -> str:
```
leonardmq, line 29: This is similar to `llama_index` formatting. We could offer picking different formatters down the line and have that as part of a synthesizer config (we don't have that right now).

Also could allow adding metadata to files (right now we only have doc id and chunk idx; we should add page number as well, but also let user set arbitrary per-document ones)

leonardmq:
@coderabbitai review

libs/core/kiln_ai/datamodel/tool_id.py
```diff
@@ -81,3 +91,15 @@ def mcp_server_and_tool_name_from_id(id: str) -> tuple[str, str]:
                 f"Invalid MCP tool ID: {id}. Expected format: 'mcp::(remote|local)::<server_id>::<tool_name>'."
             )
     return parts[2], parts[3]  # server_id, tool_name
+
+
+def rag_config_id_from_id(id: str) -> str:
+    """
+    Get the RAG config ID from the ID.
+    """
+    error_message = f"Invalid RAG tool ID: {id}. Expected format: 'kiln_tool::rag::<rag_config_id>'."
```
scosman: nit, weird to allocate this before it's needed.

libs/core/kiln_ai/tools/rag_tools.py
```diff
@@ -0,0 +1,152 @@
...
+            str(self._rag_config.vector_store_config_id), self.project.path
+        )
+        if vector_store_config is None:
+            raise ValueError(
+                f"Vector store config not found: {self._rag_config.vector_store_config_id}"
+            )
+        self._vector_store_config = vector_store_config
+        self._vector_store_adapter: BaseVectorStoreAdapter | None = None
+
+    @cached_property
```
scosman, line 79: handy!

libs/core/kiln_ai/tools/rag_tools.py
```diff
@@ -0,0 +1,152 @@
...
+
+    async def run(self, query: str) -> str:
+        _, embedding_adapter = self.embedding
+
+        vector_store_adapter = await self.vector_store()
+        store_query = VectorStoreQuery(
+            query_embedding=None,
+            query_string=query,
+        )
+        if self._vector_store_config.store_type in [
```
scosman: nit: defensive coding. Make a switch with an exhaustive check. Forces us to update this whenever we add enum values.

libs/core/kiln_ai/datamodel/tool_id.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/518/commits/213bd49a356545db9e9e7acf8d3698edc4cce518 - I suspect I copied this from elsewhere but it is not there anymore 😄

libs/core/kiln_ai/tools/rag_tools.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/518/commits/213bd49a356545db9e9e7acf8d3698edc4cce518

Also helped find out there was a bug. Fixed bug where we were generating vectors for `hybrid` and `fts`, but not for `vector-only` type search. We should not generate embeddings for `fts`, and should generate them for the `vector` query type.

PR #520, opened 2025-08-21
Bring back the demo tools, if a config value is set

scosman:
Also: organize the available_tool list into sets. One set for demo tools, one for each MCP server, and ready for RAG.

Screenshot 2025-08-21 at 5 42 30 PM

## Summary by CodeRabbit

- New Features
  - Available tools are now grouped into named sets (e.g., per server) instead of a single flat list; the UI shows grouped tool options.
  - Optional demo tools set (Addition, Subtraction, Multiplication, Division) can be enabled via an environment flag.
  - Expanded public API schema: adds audio, image, file, and function/tool-call message types.
  - Task run payloads can include rich, chat-style traces.
- Tests
  - Updated to validate grouped tool sets, demo tools flag behavior, and multi-server scenarios.

### DIFF
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -5,7 +5,8 @@ from fastapi import FastAPI, HTTPException
 from kiln_ai.datamodel.basemodel import ID_TYPE
 from kiln_ai.datamodel.external_tool_server import ExternalToolServer, ToolServerType
 from kiln_ai.tools.mcp_session_manager import MCPSessionManager
-from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, ToolId
+from kiln_ai.tools.tool_id import MCP_REMOTE_TOOL_ID_PREFIX, KilnBuiltInToolId, ToolId
+from kiln_ai.utils.config import Config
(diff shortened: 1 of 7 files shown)


app/web_ui/src/lib/api_schema.d.ts
```diff
@@ -1249,6 +1254,171 @@ export interface components {
             /** Imported Count */
             imported_count: number;
         };
+        /**
```
aiyakitori, line 1257: Was this missed by other changes? Should we setup some hook to run ./generate_schema on pull requests and make sure there is no diff?

app/desktop/studio_server/tool_api.py
```diff
@@ -99,7 +107,45 @@ async def get_available_tools(
...
                     raise_exhaustive_enum_error(server.type)
 
-        return available_tools
+            if available_mcp_tools:
+                tool_sets.append(
+                    ToolSetApiDescription(
+                        set_name="MCP Server: " + server.name,
+                        tools=available_mcp_tools,
+                    )
+                )
```
aiyakitori, line 116: I think at some point we would need to make the ordering of these tool "sets" consistent throughout the app. 

i.e. 
1. remote MCP
2. local MCP
3. RAG
...

PR #522, opened 2025-08-22
Aiya/add server screen

aiyakitori:
## What does this PR do?

* Update UI flow to include sample tools for users to add.
* Add support to add_tool page to allow pre-fill content. Instead of href, use onClick. 
* Some small UI tweaks on table column width to make things look better.  

image
image
image

## Checklists

### DIFF
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
@@ -8,7 +8,7 @@
     name: string
     description: string
     server_url: string
-    headers: { key: string; value: string }[]
+    headers: { key: string; value: string; placeholder: string | null }[]
     button_text: string
   }
 
@@ -32,14 +32,20 @@
       name: "Firecrawl",
       description: "Add Firecrawl to your project to search the web.",
       server_url: "https://firecrawl.com",
-      headers: [{ key: "Authorization", value: "" }],
+      headers: [], // Firecrawl uses https://mcp.firecrawl.dev/{FIRECRAWL_API_KEY}/sse
       button_text: "Connect",
     },
     {
       name: "Perplexity",
       description: "Add Perplexity to your project to search the web.",
       server_url: "https://api.perplexity.ai",
-      headers: [{ key: "Authorization", value: "" }],
+      headers: [
+        {
+          key: "Authorization",
+          value: "",
+          placeholder: null,
+        },
+      ],
       button_text: "Connect",
     },
     {
@@ -47,7 +53,13 @@
       description:
         "Adds ability to read repositories and code files, manage issues and PRs, analyze code, and automate workflow.",
       server_url: "https://api.githubcopilot.com/mcp/",
-      headers: [{ key: "Authorization", value: "" }],
+      headers: [
+        {
+          key: "Authorization",
+          value: "",
+          placeholder: "Bearer <Github Personal Access Token (PAT)>",
+        },
+      ],
       button_text: "Connect",
     },
     {
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
@@ -43,6 +43,25 @@
       button_text: "Connect",
     },
     {
+      name: "Twelve Data",
+      description:
+        "Integrates with Twelve Data API to provide real-time quotes, historical OHLCV price data, and instrument metadata for stocks, forex pairs, and cryptocurrencies across global markets.",
+      server_url: "https://mcp.twelvedata.com/mcp/",
+      headers: [
+        {
+          key: "Authorization",
+          value: "",
+          placeholder: "apikey {TWELVE_DATA_API_KEY}",
+        },
+        {
+          key: "X-OpenAPI-Key",
+          value: "",
+          placeholder: "YOUR_OPENAI_API_KEY",
+        },
+      ],
+      button_text: "Connect",
+    },
+    {
       name: "Postman Echo",
       description: "Simple MCP Server to test MCP tool connections.",
       server_url: "https://postman-echo-mcp.fly.dev/",
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
@@ -4,22 +4,21 @@
 
   let sections = [
     {
-      category: "Current Workspace",
+      category: "Sample Tools",
       items: [
         // TODO: Add more custom tool servers, pre-filled in.
         {
-          name: "Edit Current Task",
-          description:
-            "Modify your current task's prompt, requirements, and configuration settings.",
-          button_text: "Edit Task",
-          href: `/settings/edit_task/${$ui_state?.current_project_id}/${$ui_state?.current_task_id}`,
+          name: "Firecrawl",
+          description: "Add Firecrawl to your project to search the web.",
+          button_text: "Connect",
+          href: `/settings/manage_tools/${$ui_state?.current_project_id}/add_tools/firecrawl`,
         },
         {
-          name: "Edit Current Project",
+          name: "Demo Math Tool",
           description:
-            "Update your current project's name, description, and settings.",
-          button_text: "Edit Project",
-          href: "/settings/edit_project/" + $ui_state.current_project_id,
+            "Add built-in math tool to add/subtract/multiply/divide numbers.",
+          button_text: "Connect",
+          href: `/settings/manage_tools/${$ui_state?.current_project_id}/add_tools/math`,
         },
       ],
     },
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
@@ -29,26 +29,6 @@
 
   const sampleMcpServers: McpServer[] = [
     {
-      name: "Firecrawl",
-      description: "Add Firecrawl to your project to search the web.",
-      server_url: "https://firecrawl.com",
-      headers: [], // Firecrawl uses https://mcp.firecrawl.dev/{FIRECRAWL_API_KEY}/sse
-      button_text: "Connect",
-    },
-    {
-      name: "Perplexity",
-      description: "Add Perplexity to your project to search the web.",
-      server_url: "https://api.perplexity.ai",
-      headers: [
-        {
-          key: "Authorization",
-          value: "",
-          placeholder: null,
-        },
-      ],
-      button_text: "Connect",
-    },
-    {
       name: "GitHub",
       description:
         "Adds ability to read repositories and code files, manage issues and PRs, analyze code, and automate workflow.",
@@ -57,7 +37,7 @@
         {
           key: "Authorization",
           value: "",
-          placeholder: "Bearer <Github Personal Access Token (PAT)>",
+          placeholder: "Bearer {GITHUB_PERSONAL_ACCESS_TOKEN}",
         },
       ],
       button_text: "Connect",
@@ -80,15 +60,6 @@
           ...tool,
           on_click: () => connectRemoteMcp(tool),
         })),
-        {
-          name: "Demo Math Tool",
-          description:
-            "Add built-in math tool to add/subtract/multiply/divide numbers.",
-          button_text: "Connect",
-          on_click: () => {
-            // TODO: Register math tool
-          },
-        },
       ],
     },
     {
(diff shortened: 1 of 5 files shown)


app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
+
+                <div class="flex-shrink-0 ml-6">
+                  <div
+                    class="btn btn-mid group-hover:btn-primary transition-colors duration-200"
+                    style="min-width: 12rem;"
+                  >
+                    {item.button_text}
+                  </div>
+                </div>
+              </button>
```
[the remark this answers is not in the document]
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:134]: This is following /settings page. With this change the UI looked different and we lose consistency.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+              "Header names/values must not contain invalid characters",
+            )
+          }
+        }
+      }
+
+      const headersObj = buildHeadersObject()
+
+      //  Make a request to the server and make sure it's valid
+      try {
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: @scosman I'm not sure if this is the right place to validate the server. Should this be done under the post API `@app.post("/api/projects/{project_id}/connect_remote_mcp")` or this should be done on the UI side like the current implementation? 

The MCPServer does have a sendPing() API available. 

```
            # Validate the server URL by sending a ping request
            async with MCPSessionManager.shared().mcp_client(tool) as session:
                await session.send_ping()
```

app/desktop/studio_server/tool_api.py
```diff
@@ -77,6 +77,28 @@ class ToolSetApiDescription(BaseModel):
...
+                ToolApiDescription(
+                    id=f"{MCP_REMOTE_TOOL_ID_PREFIX}{server.id}::{tool.name}",
+                    name=tool.name,
+                    description=tool.description,
+                )
+                for tool in tools_result.tools
+            ]
+    except Exception:
+        # Skip the tool when we can't connect to the server
+        return []
```
aiyakitori: @leonardmq here we return an empty list if any of the server raised an error.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
+          value: "",
+          placeholder: "YOUR_OPENAI_API_KEY",
+        },
+      ],
+      button_text: "Connect",
+    },
+    {
+      name: "Postman Echo",
+      description: "Simple MCP Server to test MCP tool connections.",
+      server_url: "https://postman-echo-mcp.fly.dev/",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:66]: I get an error with this one: Unexpected error: Load failed. Unable to connect to the server. Please check the URL, headers and ensure the server is accessible

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
-            },
-          },
-          body: {
-            name: name.trim(),
-            server_url: server_url.trim(),
-            headers: headersObj,
-            description: description.trim() || null,
+          key: "Authorization",
+          value: "",
+          placeholder: "Bearer {GITHUB_PERSONAL_ACCESS_TOKEN}",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Maybe set value to "Bearer REPLACE_WITH_GITHUB_PERSONAL_ACCESS_TOKEN"

That way they don't have to type "Bearer "

P2.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
+    },
+    {
+      name: "Twelve Data",
+      description:
+        "Integrates with Twelve Data API to provide real-time quotes, historical OHLCV price data, and instrument metadata for stocks, forex pairs, and cryptocurrencies across global markets.",
+      server_url: "https://mcp.twelvedata.com/mcp/",
+      headers: [
+        {
+          key: "Authorization",
+          value: "",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: ditto

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
+      items: [
+        {
+          name: "Remote MCP Servers",
+          description:
+            "Connect to remote MCP servers to add tools to your project.",
+          href: `/settings/manage_tools/${$ui_state?.current_project_id}/add_tools/remote_mcp`,
+          button_text: "Connect",
+          on_click: () => {
+            goto(
+              `/settings/manage_tools/${$ui_state?.current_project_id}/add_tools/remote_mcp`,
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Try to never use current_project_id, we should kill it eventually. It might not be the project_id in the URL, in which case you have a bug. It's mostly for loading the /run page or fresh page loads of root url ("/").

You should use $page.params.project_id, which is the actual project ID from the URL. setting this at the top is a common pattern: `$: project_id = $page.params.project_id`

There's a few other places of this too, grab em all.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -1,183 +1,145 @@
...
-        id="headers_section"
-        description="If the documentation for the server you're adding requires custom headers, enter them here."
-        info_description="These are usually not needed. Some MCP servers require custom headers, such as the 'Authorization' headers. Refer to the documentation for the server you're adding to see if they require headers."
-        value=""
-      />
+<AppPage title="Add Tools">
+  <div class="max-w-4xl mt-12 space-y-12">
+    {#each sections as section}
+      <div class="space-y-6">
+        <!-- Category Header -->
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:104]: I think I said to copy/paste, but going forward let's be better about code sharing. Should create a "settings_header.svelte" and "settings_item.svelte" UI components in `/lib/ui`, and use them in both places. P2 if you want to do in this PR or not, this isn't awful or anything.

It has potential of making code cleaner. Instead of fancy maps and dynamic building of lists, you can just put a bunch of components in order. Easier to read/write/customize.

 - Header, title = "Sample tools"
 - SettingsItem, title="Github", description="...", onclick={openTool("https..."}
 - SettingsItem, title="PostHog", description="...", onclick={openTool("https..."}
 - etc

You don't want to do that today because it would be a ton of repeated code, but with a component that's gone.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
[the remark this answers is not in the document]
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:134]: @aiyakitori click "resolve" when done with comments, especially coderabbits. Helps keep PRs clean and let review know it's addressed (or you reviewed and chose to ignore).

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  let headers: HeaderPair[] = []
+
+  // Form state
+  let error: KilnError | null = null
+  let submitting = false
+
+  // Populate fields from parent page state if provided (only if fields are empty)
+  $: if ($page.state) {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: so a `$` creates a reactive block that will re-run when svelte decides it should (here for page state changing, but should be able to re-run at any time). You wouldn't really want this to re-run reactively, as it could replace user edits with the old static content.

use onMount instead, which runs once after the component mounts.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+    const state = $page.state as {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: no casting / never use as. Really: never allowed. It leads to ugly breaking runtime errors.

Check if the fields exist and use them if they do. More

```
const state = $page.state || {}
if ("name" in state && typeof state["name"] === "string") {
    name = state.name
}
```

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  async function connect_remote_mcp() {
+    try {
+      error = null
+      submitting = true
+
+      // Validate required fields
+      if (!name.trim()) {
+        throw new Error("Name is required")
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: Always do validation logic to the server. The client can always be bypassed. The server owns checking data is valid.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+            )
+          }
+        }
+      }
+
+      const headersObj = buildHeadersObject()
+
+      //  Make a request to the server and make sure it's valid
+      try {
+        const response = await fetch(server_url.trim(), {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: move this to the server as well.

Making the request from inside the browser is going to add lots of headers and limitations that our server doesn't have (lots of defaults header, CORS, etc).

We don't want to check "can the web app connect to MCP" (we never want to do that). Even it if can't (CORS error), the server might be able to. We want to test "can the MCP python library we integrate connect from our backend" so just check that directly.

MCP protocol wise, I'm not sure a GET to the base URL has to do anything or has to return an error if headers are incorrect. That might not be part of the standard? A "list_tools" call would be a better way to really validate it is working?

Also: same issue of validation - validation should be on the server. Never trust the client to validate. They could just skip this and break the server.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+      } catch (e) {
+        // Log custom error message
+        const error = e as Error
+        throw new Error(
+          `${error.message}. Unable to connect to the server. Please check the URL, headers and ensure the server is accessible`,
+        )
+      }
+
+      const { data, error: api_error } = await client.POST(
+        "/api/projects/{project_id}/connect_remote_mcp",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte:69]: put your validation (including server check) inside this API. Make it return really helpful errors like you do here and those will bubble up the same way in the catch block below.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+      />
+
+      <!-- Headers section -->
+      <FormElement
+        inputType="header_only"
+        label="Headers"
+        id="headers_section"
+        description="If the documentation for the server you're adding requires custom headers, enter them here."
+        info_description="These are usually not needed. Some MCP servers require custom headers, such as the 'Authorization' headers. Refer to the documentation for the server you're adding to see if they require headers."
+        value=""
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte:147]: value needed for header?

scosman (changes requested):

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: Moving to server actually makes validation a lot easier!

Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: Alternatively we can use the send_ping() API from MCP.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:66]: This was caused by weird browser request, once this was moved the server it worked. URL was correct. 
I validated the logic using GitHub and missed the empty header case.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte:147]: Only the ones that needs API_Key, will update the wording.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:104]: We would need a different kind of Local/MCP as well. Would keep this as it and clean up in another PR.

app/desktop/studio_server/tool_api.py
```diff
@@ -77,6 +131,52 @@ class ToolSetApiDescription(BaseModel):
...
+            tools_result = await session.list_tools()
+            return [
+                ToolApiDescription(
+                    id=f"{MCP_REMOTE_TOOL_ID_PREFIX}{server.id}::{tool.name}",
+                    name=tool.name,
+                    description=tool.description,
+                )
+                for tool in tools_result.tools
+            ]
+    except Exception:
```
scosman: generally don't ignore "Exception". If you want to ignore a specific exception that's fine, but keep it specific.

app/desktop/studio_server/tool_api.py
```diff
@@ -212,27 +300,24 @@ async def connect_remote_mcp(
...
-                properties=properties,
-                parent=project,
-            )
+        # Create the ExternalToolServer with required fields
+        properties = {
+            "server_url": tool_data.server_url,
+            "headers": tool_data.headers,
+        }
+
+        tool = ExternalToolServer(
```
scosman: nit, this is a tool server not a tool

scosman (approved):
optional nits, but also can merge

app/desktop/studio_server/tool_api.py
```diff
@@ -30,6 +32,58 @@ class ExternalToolServerCreationRequest(BaseModel):
     headers: Dict[str, str] = Field(default_factory=dict)
     description: str | None = None
 
+    @model_validator(mode="after")
+    def validate_server_details(self):
+        """Validate server URL and headers format."""
+        # Validate server URL
+        server_url = self.server_url
```
leonardmq, line 39: Optional thing: Pydantic seems to have some types for URL validation; maybe could save some of the manual URL validation work: https://docs.pydantic.dev/2.0/usage/types/urls/

aiyakitori: Will address these in next PR.

PR #530, opened 2025-08-25
Updated judge default score type value

sfierro:
## Summary by CodeRabbit

* **Improvements**
  * Automatically picks an appropriate score metric when pass/fail-style outputs are present (defaults to MSE); otherwise uses Kendall Tau.
  * Defers sorting and score display until a score type is determined to ensure consistent ordering.

* **Bug Fixes**
  * Prevents premature rendering and sorting of score headers/cells while scoring data isn’t ready.

* **UX**
  * Shows "None" with a tooltip for missing scores to clarify unavailable data.

### DIFF
--- a/app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
+++ b/app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
@@ -75,12 +75,7 @@
         )))
   )
 
-  // Update sorting when score_type or score_summary changes
-  $: if (eval_configs && score_summary && evaluator) {
-    sortEvalConfigs()
-  }
-
-  // Sort eval_configs whenever score_type changes
+  // Sort eval_configs whenever score_type or score_summary changes
   $: if (score_type && eval_configs && score_summary && evaluator) {
     sortEvalConfigs()
   }
@@ -212,6 +207,15 @@
         throw error
       }
       evaluator = data
+      // Override default score_type to "mse" when eval rating type is pass_fail or pass_fail_critical
+      // We have different scores to compare human preference to LLM-as-Judge in evals. Kendall Tau (the current default) is a great score for comparing ranked data, but kinda lousy for pass/fail data
+      const hasPassFailRating = evaluator.output_scores.some(
+        (score) =>
+          score.type === "pass_fail" || score.type === "pass_fail_critical",
+      )
+      if (hasPassFailRating) {
+        score_type = "mse"
+      }
     } catch (error) {
       eval_error = createKilnError(error)
     } finally {


app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
```diff
@@ -560,18 +569,20 @@
...
-                    <InfoTooltip
-                      tooltip_text={info_tooltip_text(
-                        output_score.type,
-                        score_type,
-                      )}
-                      no_pad={true}
-                    />
-                  </th>
-                {/each}
+                {#if score_type}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: `if` should be inside. Pretty sure tables will complain or do strange things if missing cells in a row.

Alternative to allowing null is to update it when it loads. Update will be atomic to the data coming in so should be fine visually, rending in the same frame. Would minimize code change.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
```diff
@@ -50,7 +50,20 @@
     | "pearson"
     | "kendalltau"
 
-  let score_type: ScoreType = "kendalltau"
+  let score_type: ScoreType | null = null
+
+  // Set default score_type to "mse" when eval rating type is pass_fail or pass_fail_critical, otherwise use "kendalltau"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: I'd comment on the "why" here more than the what. The code says what it does, but it's a bit of a weird thing to do, so a comment on why we are doing this is probably helpful.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
```diff
@@ -50,7 +50,20 @@
...
+  $: if (evaluator?.output_scores && !score_type) {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: You could also put this inside get_eval after fetching the eval for simplicity. 

This should work fine, but this syntax is a reactive block, which will re-run anytime evaluator, output_scores or score_type change. Like I said, should be fine (especially since you have the `!score_type` check), but since we're setting something for initial state, and don't actually want it to update automatically when other things change, I'd lean to putting in in get_eval.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
```diff
@@ -75,12 +88,7 @@
         )))
   )
 
-  // Update sorting when score_type or score_summary changes
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte:78]: merci

scosman (changes requested):
some nits

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: Love this idea!

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
```diff
@@ -652,58 +668,60 @@
...
-                        <InfoTooltip
-                          tooltip_text="No scores were found for this judge. Click 'Run All Eval' to generate scores and ensure your golden dataset has human ratings."
-                          no_pad={true}
-                        />
-                      {/if}
-                    </td>
-                  {/each}
+                      </td>
+                    {/each}
+                  {/if}
```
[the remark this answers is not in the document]
Comment (scosman) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: @sfierro the comment about 0.0 spearman seems valid (but separate). Mind checking that out in another PR?

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/eval_configs/+page.svelte]: Ah yeah with your other comment I can definitely update when it loads in the get_eval, with the way I had it as a reactive statement I needed the null to know when it was the first time

PR #534, opened 2025-08-26
Use str instead of enum when loading remote config provider models

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Improved compatibility with remote/unknown model names across providers.
  * Provider model listings now include remote models with correct identifiers.
  * Applies a sensible default structured output mode per provider when a model is unrecognized.

* **Bug Fixes**
  * Prevents errors when encountering model names not in the built-in list.

* **Tests**
  * Added coverage to verify graceful handling and correct exposure of remote models in provider listings.

### DIFF
--- a/app/desktop/studio_server/provider_api.py
+++ b/app/desktop/studio_server/provider_api.py
@@ -16,7 +16,6 @@ from kiln_ai.adapters.docker_model_runner_tools import (
 from kiln_ai.adapters.ml_model_list import (
     KilnModel,
     KilnModelProvider,
-    ModelName,
     ModelProviderName,
     StructuredOutputMode,
     built_in_models,
@@ -157,7 +156,7 @@ class ProviderModel(BaseModel):
 
 
 class ProviderModels(BaseModel):
-    models: Dict[ModelName, ProviderModel]
+    models: Dict[str, ProviderModel]
 
 
 def connect_provider_api(app: FastAPI):
--- a/app/desktop/studio_server/test_provider_api.py
+++ b/app/desktop/studio_server/test_provider_api.py
@@ -16,6 +16,7 @@ from kiln_ai.adapters.ml_model_list import (
     ModelProviderName,
     StructuredOutputMode,
     built_in_models,
+    default_structured_output_mode_for_model_provider,
 )
 from kiln_ai.utils.config import Config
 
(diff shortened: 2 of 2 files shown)


app/desktop/studio_server/test_provider_api.py
```diff
@@ -2893,3 +2894,48 @@ async def test_connect_wandb_request_exception(mock_config_shared, mock_requests
         not hasattr(mock_config, "wandb_api_key")
         or mock_config.wandb_api_key != "test-api-key"
     )
+
+
+def test_remote_model_name_not_in_enum_handled_gracefully():
```
sfierro, line 2899: This was already passing without my change but it is kind of related code we should be robust around too I think (and sonnet thought)

app/desktop/studio_server/test_provider_api.py
```diff
@@ -2893,3 +2894,48 @@ async def test_connect_wandb_request_exception(mock_config_shared, mock_requests
...
+        model_name=remote_model_name,
+        provider=ModelProviderName.openai,
+        default=StructuredOutputMode.json_schema,
+    )
+
+    # Should return the default mode since the model is not found
+    assert result == StructuredOutputMode.json_schema
+
+
+def test_get_providers_models_handles_remote_models(client):
```
sfierro, line 2914: this is the test that failed before and passes now!

app/desktop/studio_server/provider_api.py
```diff
@@ -157,7 +156,7 @@ class ProviderModel(BaseModel):
 
 
 class ProviderModels(BaseModel):
-    models: Dict[ModelName, ProviderModel]
+    models: Dict[str, ProviderModel]
```
sfierro, line 159: the fix!

app/desktop/studio_server/test_provider_api.py
leonardmq: So impact is the model selection dropdown (and similar pieces) is unusable if remote config serves any model that the client did not already know about. 

Did you run into something like this while using the app or how did you find out?

scosman: @leonardmq a user running the prod version noticed the error in console. App was pretty much functional without it, as available_models still work. We don't use this endpoint for as much (worth asking if both are needed)

All users are basically living with this broken right now, and no complains except for the console log which is strange.

scosman (approved):
This change is good.

The chat brought up an interesting question: what is this API used for and why does no one complain when it's broken. We also have `/available_models` which is what we use for most things related to model info. Our pages load both right on startup. 

Can we implement the things this API is used for (model_name_from_id/get_model_info) using available_models in stores.ts and avoid having 2 calls to fetch model info? Not urgent

scosman:
replying to my own comment: this is indeed a subset. Very different structure. We could definitely just derive the results of this API on the client from available_models and avoid a fetch. That said, it's a <<1ms local fetch so this isn't urgent.

PR #536, opened 2025-08-27
Support going straight from fine tune page to proper data gen page

sfierro:


leonardmq (commented):
Looks ok to me, tested and works correctly (but letting Steve approve)

scosman (approved):

PR #538, opened 2025-08-28
Add support for local MCP server

aiyakitori:
## What does this PR do?

* Add support for local MCP Server by introducing a new ToolServerType of `local_mcp`.
* Add new API `connect_local_mcp`, related UI screens
* Add new svelte component for settings header and item to share the code. 

## Checklists

## Summary by CodeRabbit

- **New Features**
  - Connect Local MCP servers via a new settings flow and API; local and remote MCPs are discoverable through a unified tooling list.
  - Tool details now show run configuration, arguments and environment variables.
  - Curated sample remote and local tools for quick setup.

- **Refactor**
  - Settings UI refactored into reusable header/section/item components.

- **Bug Fixes**
  - Clearer connectivity/error propagation and stricter input validation for local tool setup.

- **Tests**
  - Expanded coverage for local MCP workflows, ID handling, and connectivity scenarios.

### DIFF
--- a/app/desktop/studio_server/test_tool_api.py
+++ b/app/desktop/studio_server/test_tool_api.py
@@ -2246,7 +2246,7 @@ def test_local_tool_server_creation_request_missing_command():
 
 
 def test_local_tool_server_creation_request_empty_args():
-    """Test LocalToolServerCreationRequest accepts empty args list (arguments no longer required)"""
+    """Test LocalToolServerCreationRequest accepts empty args list"""
 
     request = LocalToolServerCreationRequest(
         name="Empty Args Server",
@@ -2716,7 +2716,7 @@ def test_create_local_tool_server_empty_command(client, test_project):
 
 
 async def test_create_local_tool_server_empty_args(client, test_project):
-    """Test local tool server creation succeeds when args are empty (arguments no longer required)"""
+    """Test local tool server creation succeeds when args are empty"""
     tool_data = {
         "name": "empty_args_tool",
         "command": "python",
--- a/libs/core/kiln_ai/tools/test_mcp_session_manager.py
+++ b/libs/core/kiln_ai/tools/test_mcp_session_manager.py
@@ -254,7 +254,7 @@ class TestMCPSessionManager:
             )
 
     async def test_local_mcp_empty_args_allowed(self):
-        """Test that empty args list is now allowed for local MCP (arguments no longer required)."""
+        """Test that empty args list is now allowed for local MCP."""
         # Should not raise any exception - empty args are now allowed
         tool_server = ExternalToolServer(
             name="empty_args_server",


app/desktop/studio_server/tool_api.py
```diff
@@ -85,6 +90,26 @@ def validate_server_details(self):
...
+
+    @model_validator(mode="after")
+    def validate_command(self):
+        """Validate command format."""
+        if not self.command:
+            raise ValueError("Command is required")
+
+        # Validate args
+        if not self.args:
+            raise ValueError("Args are required")
```
scosman: Maybe not? A MCP server could just be a command without args: golang, rust, binaries

app/desktop/studio_server/tool_api.py
```diff
@@ -174,6 +208,19 @@ async def validate_tool_server_connectivity(tool_server: ExternalToolServer):
                     status_code=422,
                     detail=f"Failed to connect to the server: {str(e)}",
                 )
+        case ToolServerType.local_mcp:
```
scosman: don't duplicate code. Can just have `case ToolServerType.remote_mcp | ToolServerType.local_mcp`

Edit: like you do below

app/web_ui/src/routes/(app)/settings/+page.svelte
```diff
@@ -123,77 +125,18 @@
...
-                  <div
-                    class="btn btn-mid group-hover:btn-primary transition-colors duration-200"
-                    style="min-width: 12rem;"
-                  >
-                    {item.button_text}
-                  </div>
-                </div>
-              </button>
-            {/if}
+            <SettingsItem
```
scosman: nice

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -69,15 +91,49 @@
...
+      name: "Firecrawl",
+      description:
+        "Firecrawl is a tool that allows you to crawl websites and extract data.",
+      command: "npx",
+      args: ["-y", "firecrawl-mcp"],
+      env_vars: [
+        {
+          key: "FIRECRAWL_API_KEY",
+          value: "",
+          placeholder: "REPLACE_WITH_FIRECRAWL_API_KEY",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Don't use the "REPLACE_WITH" for placeholder. Can just use plain English. Only need that for cases where we set the value and they need to replace a string.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -69,15 +91,49 @@
...
+    },
+    {
+      name: "Filesystem",
+      description:
+        "Read, write, and manipulate local files through a controlled API.",
+      command: "npx",
+      args: [
+        "-y",
+        "@modelcontextprotocol/server-filesystem",
+        "REPLACE_WITH_OTHER_ALLOWED_DIRECTORIES",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Ditto, and we need a better strong. What's "other", I don't see any others? Give examples.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -69,15 +91,49 @@
...
     {
       category: "Sample Tools",
       items: [
-        // TODO: Add more custom tool servers, pre-filled in.
-        ...sampleMcpServers.map((tool) => ({
+        ...sampleRemoteMcpServers.map((tool) => ({
           ...tool,
           on_click: () => connectRemoteMcp(tool),
         })),
+        ...sampleLocalMcpServers.map((tool) => ({
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: maybe another PR, but I'l just make this inline in the HTML now that you have a control. Will be way cleaner than this concatenation / building / mapping code. Easier to read and write.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+</script>
+
+<AppPage
+  title="Connect Local MCP Server"
+  subtitle="Connect to a local Model Context Protocol (MCP) server to add external
+        tools to your project."
+>
+  <div class="max-w-2xl">
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:125]: Add a "installation instructions box" for the demo ones that tell you how to install. 

Something like:

```
{#if $page.state.installation_instruction}
  <Warning warning_color="warning" warning_icon="info" large_icon={true} warning_message={$page.state.installation_instruction} />
{/if}
```

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -0,0 +1,194 @@
...
+        label="Description"
+        inputType="textarea"
+        id="mcp_description"
+        description="A description of this MCP server for your reference."
+        optional={true}
+        bind:value={description}
+      />
+
+      <FormElement
+        label="Command"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:150]: command and args need info_descriptions. They are non obvious to new users.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -0,0 +1,194 @@
...
+        id="command"
+        description="The command to run the MCP server."
+        placeholder="uv, python, npx etc."
+        bind:value={command}
+      />
+
+      <FormElement
+        label="Arguments"
+        id="args"
+        description="The arguments to pass to the MCP server. Each argument should be space separated."
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: "The arguments to pass to the MCP server." -> ""The arguments to pass to the command above". These are technically passed to the command not the server

"Each argument should be space separated" seems weird. It's just an argument list. Sometimes it has `--asdf=1234`. It's a weird detail to include. Just remove it.

Add info_descriptions for noobs.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -0,0 +1,194 @@
...
+        empty_content={{
+          key: "",
+          value: "",
+        }}
+        let:item_index
+      >
+        <div class="flex gap-2">
+          <div class="flex-1 max-w-[200px]">
+            <FormElement
+              label="Environment Variable Name"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: Just "Name", it says "Environment Variable" twice above this.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -0,0 +1,194 @@
...
+              id="env_var_name_{item_index}"
+              info_description="The name of the environment variable, such as 'API_KEY'"
+              placeholder="Name"
+              light_label={true}
+              bind:value={env_vars[item_index].key}
+            />
+          </div>
+          <div class="flex-1">
+            <FormElement
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:202]: Screenshot 2025-08-28 at 3 29 30 PM

Errors are ugly. Might not be specific to this PR though.

app/desktop/studio_server/tool_api.py
```diff
@@ -174,6 +208,19 @@ async def validate_tool_server_connectivity(tool_server: ExternalToolServer):
...
+            try:
+                async with MCPSessionManager.shared().mcp_client(
+                    tool_server
+                ) as session:
+                    # Use list tools to validate the server is reachable
+                    await session.list_tools()
+            except Exception as e:
```
scosman: same issue as other PR:

 - never catch all exceptions, should always be targeted at specific failure. Let "Exception" bubble up.
 - Don't assume 422 (input error) when that's no what you caught.

app/desktop/studio_server/tool_api.py
```diff
@@ -174,6 +208,19 @@ async def validate_tool_server_connectivity(tool_server: ExternalToolServer):
                     status_code=422,
                     detail=f"Failed to connect to the server: {str(e)}",
                 )
+        case ToolServerType.local_mcp:
+            try:
+                async with MCPSessionManager.shared().mcp_client(
+                    tool_server
+                ) as session:
+                    # Use list tools to validate the server is reachable
+                    await session.list_tools()
```
scosman: Not sure if we can address it, but fire crawl doesn't check API key on list_tools. Any ideas?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -111,6 +138,14 @@
       }),
     )
   }
+  function getEnvVarsProperties(tool: ExternalToolServerApiDescription) {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: nit: I'd generalize all the way or not at all. You're checking tool_server.type both in the script tag and down in HTML, pick one.

Generalize all the way: Make a method that gets all of the data for the property lists you want
```
[
  {name: "Properties", content: {prop1:val1,...},
  {name: "Environment Variables", content: {....}},
  ...
]
```

Don't generalize:
 - don't make change to properties above and do it in HTML
```
{#if tool_server.type === "local_mcp"}
   all the PropertyList you want here
{:elseif tool_server.type === "remote_mcp"}
   all the PropertyList you want here
{/end}
```

I prefer option 2. Way more readable. Keeps the code linear and in 1 place. If you end up wanting something more custom (spacing, subtitles) you can just add them.

Don't be afraid to inline things, you don't need to define methods for one-liners:
```
<PropertyList
                properties={Object.entries(tool.properties["env_vars"] || {}).map(
      ([key, value]) => ({
        name: key,
        value: String(value || "N/A"),
      }),
    )}
                title="Environment Variables"
               />
```

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -192,20 +227,25 @@
...
-              <div class="mt-8">
-                <PropertyList
-                  properties={getHeadersProperties(tool_server)}
-                  title="Headers"
-                />
-              </div>
-            {/if}
+          <PropertyList
+            properties={getConnectionProperties(tool_server)}
+            title="Connection Details"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: "Connection" seems remote MCP specific, better title for local?

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -47,20 +48,50 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+                        "command is required for external tools of type 'local_mcp'"
+                    )
+
+                args = self.properties.get("args", None)
+                if not isinstance(args, list):
+                    raise ValueError(
+                        "args must be a list for external tools of type 'local_mcp'"
+                    )
+                if not args:
+                    raise ValueError(
```
scosman: per above, empty is valid

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -47,20 +48,50 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+                        "args is required for external tools of type 'local_mcp'"
+                    )
+
+                env_vars = self.properties.get("env_vars", {})
+                if not isinstance(env_vars, dict):
+                    raise ValueError(
+                        "env_vars must be a dictionary for external tools of type 'local_mcp'"
+                    )
+                # Set the default value if not provided
```
scosman: slightly awkward to get it setting a default value, then after check for missing.

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -47,20 +48,50 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+                if not isinstance(args, list):
+                    raise ValueError(
+                        "args must be a list for external tools of type 'local_mcp'"
+                    )
+                if not args:
+                    raise ValueError(
+                        "args is required for external tools of type 'local_mcp'"
+                    )
+
+                env_vars = self.properties.get("env_vars", {})
```
scosman, line 80: Should validate keys? A key with spaces for example? Not sure what the restrictions are.

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -52,3 +65,33 @@ async def mcp_client(
...
+        """
+        Create a session for a local MCP server.
+        """
+        command = tool_server.properties.get("command")
+        if not command:
+            raise ValueError("command is required")
+
+        args = tool_server.properties.get("args", [])
+        if not args:
+            raise ValueError("argument is required")
```
scosman: Check it's an array, but "not args" will fire on empty array which we don't want

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -29,13 +31,24 @@ async def mcp_client(
...
     ]:
-        # Only support remote MCP servers for now
         match tool_server.type:
             case ToolServerType.remote_mcp:
-                pass
+                async with self._create_remote_mcp_session(tool_server) as session:
+                    yield session
+            case ToolServerType.local_mcp:
+                async with self._create_local_mcp_session(tool_server) as session:
+                    yield session
```
scosman, line 40: clean!

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -52,3 +65,33 @@ async def mcp_client(
...
+    async def _create_local_mcp_session(
+        self,
+        tool_server: ExternalToolServer,
+    ) -> AsyncGenerator[ClientSession, None]:
+        """
+        Create a session for a local MCP server.
+        """
+        command = tool_server.properties.get("command")
+        if not command:
+            raise ValueError("command is required")
```
scosman: In general your errors are way too short. Any of these could be bubbled up to a user and give them some helpful info. What would "comment is required" mean to a random user who just hit run on the "Run" tab? nada. "Attempted to start local MCP server, but no command was provided". See other errors in the PR, most have this issue. Keep them short, but enough description to be helpful without a stack trace.

scosman (changes requested):

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: "<Other Allowed Directories e.g. /Users/username/Desktop>" this is better

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: haha i actually asked gpt for this and they recommended using "Connections" to keep the UI consistent, will name it something else.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: Agree

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: Use "Run Configuration" instead

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:202]: This re-uses existing component, will address this in future PR.

libs/core/kiln_ai/datamodel/external_tool_server.py
aiyakitori: key validation is now done at the request level

app/desktop/studio_server/tool_api.py
aiyakitori: Don't have any good solution on top of my head, send_ping() is also another API we could leverage but not sure if all implementation require the API_KEY.

scosman: yeah, probably not something we need / can solve

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -84,9 +84,6 @@ def validate_required_fields(self) -> "ExternalToolServer":
                     raise ValueError(
                         "env_vars must be a dictionary for external tools of type 'local_mcp'"
                     )
-                # Set the default value if not provided
-                if "env_vars" not in self.properties:
-                    self.properties["env_vars"] = {}
```
scosman: Don't set values in validators. Treat them as strictly validators (They should be read only). Just require it and callers will figure it out.

scosman (approved):
one small comment, but you can merge whenever

PR #557, opened 2025-09-02
fix: svelte-check warnings about unused ts annotations

leonardmq:
## What does this PR do?

Started getting errors in the svelte-check that runs in `./checks.sh`. This seems to be due to the API schema types having changed slightly such that the TS bypass annotations are no longer needed - which triggers an error.

I do not know why I started getting these errors today - my `openapi-typescript` package seems to be the same version as in `package.json`; and these errors are coming from files that I have not touched. Started getting them on my feature branch, but am also getting them on `main`.

To test if you are getting the same issue on `main`:
1. Edit a Svelte file (for example, `./app/web_ui/src/routes/(app)/settings`) and save
2. `git add . && ./checks.sh`
3. See if you get the svelte-check errors

And:
```sh
pushd app/web_ui/src/lib && ./generate_schema.sh && popd && git add . && ./checks.sh
```

To see if this PR introduces an issue for you:
1. Edit a Svelte file (for example, `./app/web_ui/src/routes/(app)/settings`) and save
2. `git add . && ./checks.sh`
3. See if you get other issues (maybe typecheck errors)

## Summary by CodeRabbit

- New Features
  - More flexible API payloads: many endpoints now accept generic key–value data for request and response bodies.
  - Public API surface simplified by removing legacy model metadata and a large model-name enum.

- Bug Fixes
  - Settings > Providers: “Custom models” now validates as an array of strings and raises a clear error if invalid.

- Refactor
  - Removed several TypeScript error-suppression comments so payloads are fully type-checked (no expected runtime changes).

### DIFF
--- a/app/web_ui/src/lib/api_schema.d.ts
+++ b/app/web_ui/src/lib/api_schema.d.ts
@@ -1240,7 +1240,9 @@ export interface components {
             name?: string | null;
             type: components["schemas"]["EvalConfigType"];
             /** Properties */
-            properties: Record<string, never>;
+            properties: {
+                [key: string]: unknown;
+            };
             /** Model Name */
             model_name: string;
             provider: components["schemas"]["ModelProviderName"];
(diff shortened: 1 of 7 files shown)


leonardmq:
Forgot to run `npm i` after switching branches - there must be a new dependency on my other branch that changes something with the codegen. Closing

PR #562, opened 2025-09-04
Fix tool calling + reasoning models

scosman:
Fix several issues with tool calling, traces, and reasoning when combined with tools. See commit messages for detailed breakdowns.

## Summary by CodeRabbit

- New Features
  - Manage external tools: add, view, and remove Remote/Local MCP tool servers per project; enable/disable demo tools; detailed server pages with available tools.
  - Tool-calling runs: select multiple tools per task; tools included in run requests; model dropdown warns/filters when tool support is required.
  - Per-project available tools store and per-task tool selection persistence.
- Improvements
  - Multi-select Fancy Select, collapse header badges, empty-state messaging, reusable settings components.
  - Models now display tool-calling support.
- Bug Fixes
  - Show “Output Model” only when model_id is a string.
- Chores
  - Added MCP CLI dependency.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
@@ -670,11 +670,14 @@ class LiteLlmAdapter(BaseAdapter):
         Extract the message from the model with type safety and validation.
         """
         content: str | None = None
+        reasoning_content: str | None = None
         tool_calls: List[ChatCompletionMessageToolCall] | None = None
         tool_calls_param: List[ChatCompletionMessageToolCallParam] = []
         if hasattr(response_choice, "message"):
             if hasattr(response_choice.message, "content"):
                 content = response_choice.message.content
+            if hasattr(response_choice.message, "reasoning_content"):
+                reasoning_content = response_choice.message.reasoning_content
             if hasattr(response_choice.message, "tool_calls"):
                 tool_calls = response_choice.message.tool_calls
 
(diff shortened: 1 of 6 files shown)


libs/core/kiln_ai/utils/open_ai_types.py
```diff
@@ -34,8 +34,12 @@
...
-    Almost exact copy of ChatCompletionAssistantMessageParam, but with List[T] instead of Iterable[T] for tool_calls.
+    Almost exact copy of ChatCompletionAssistantMessageParam, but two changes.
+
+    First change: List[T] instead of Iterable[T] for tool_calls. Addresses pydantic issue.
     https://github.com/pydantic/pydantic/issues/9541
+
+    Second change: Add reasoning_content to the message. A LiteLLM property for reasoning data.
     """
 
     role: Required[Literal["assistant"]]
```
aiyakitori, line 45: nit: use constant for "assistant"? Not sure if it's related but we are checking the role with raw strings above.
```
if raw_message.role != "assistant":
```

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -108,11 +117,18 @@ async def _run_model_turn(
             # count the usage
             usage += self.usage_from_response(model_response)
 
-            # Extract content and tool_calls safely
-            content, assistant_message, tool_calls = self.extract_assistant_response(
-                response_choice
-            )
-            messages.append(assistant_message)
+            # Extract content and tool calls
+            if not hasattr(response_choice, "message"):
```
aiyakitori, line 121: nit: we should probably use constants to access these attributes from a response.

libs/core/kiln_ai/utils/open_ai_types.py
scosman: these are openAI defined, not us, and they don't expose a constant. All their example also just use strings. Not sure what's "pythonic" here but I'll follow their lead.

PR #563, opened 2025-09-04
Store MCP Server Secrets in Settings

aiyakitori:
## What does this PR do?

Headers and Env Vars previous were stored as part of the data model which can be synced. This PR adds the ability for tool servers to save their "secrets" such as API Key inside Kiln Settings. 

* Add logic to handle secrets lifecycle in ExternalToolServer.
* Update _Tool Server_ screen to mask secrets and indicate if a key is missing simulating initial sync state. 
* Add status column to  _Manage Tool_ screen to indicate if a tool server has missing keys. 
* Update connect tool server screen to mark if a header or env var is a secret. 

## Checklists

image
image

## Summary by CodeRabbit

- **New Features**
  - Centralized MCP secret storage (MCP_SECRETS_KEY); tool-server secrets tracked separately, masked in UI, and listings expose missing_secrets with Status (Ready / Action Required).

- **Refactor**
  - Unified key/value model for headers/env vars with per-item "Secret" toggles in add-tool UIs. Creation payloads include secret_header_keys/secret_env_var_keys. Sessions merge secrets into requests/env without mutating originals. Shared UiProperty type introduced for property lists.

- **Tests**
  - Expanded tests for validation, secret lifecycle, session merging, UI flows, and edge cases.

### DIFF
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -322,7 +322,6 @@ def connect_tool_servers_api(app: FastAPI):
         project_id: str, tool_server_id: str
     ) -> ExternalToolServerApiDescription:
         tool_server = tool_server_from_id(project_id, tool_server_id)
-        print(f"tool_server: {tool_server}")
 
         # Get available tools based on server type
         available_tools = []
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -36,6 +36,7 @@ class ExternalToolServerCreationRequest(BaseModel):
     description: str | None = None
     server_url: str
     headers: Dict[str, str] = Field(default_factory=dict)
+    secret_header_keys: List[str] = Field(default_factory=list)
 
     @model_validator(mode="after")
     def validate_server_details(self):
@@ -86,6 +87,24 @@ class ExternalToolServerCreationRequest(BaseModel):
 
             # Update headers to validated version
             self.headers = validated_headers
+
+        # Validate secret header keys
+        if isinstance(self.secret_header_keys, list):
+            validated_secret_header_keys = []
+            for key in self.secret_header_keys:
+                key_str = key.strip() if isinstance(key, str) else str(key).strip()
+                if not key_str:
+                    raise ValueError("Secret header key is required")
+                # Check if the key is in the headers
+                if key_str not in self.headers:
+                    raise ValueError(
+                        f"Secret header key {key_str} is not in the headers"
+                    )
+
+                validated_secret_header_keys.append(key_str)
+
+            # Update secret header keys to validated version
+            self.secret_header_keys = validated_secret_header_keys
 
         return self
 
@@ -303,6 +322,7 @@ def connect_tool_servers_api(app: FastAPI):
         project_id: str, tool_server_id: str
     ) -> ExternalToolServerApiDescription:
         tool_server = tool_server_from_id(project_id, tool_server_id)
+        print(f"tool_server: {tool_server}")
 
         # Get available tools based on server type
         available_tools = []
@@ -337,7 +357,7 @@ def connect_tool_servers_api(app: FastAPI):
     ) -> ExternalToolServer:
         project = project_from_id(project_id)
 
-        # Create the ExternalToolServer with required fields
+        # Create the ExternalToolServer with required fields, temporarily store all headers in properties to validate the connectivity
         properties = {
             "server_url": tool_data.server_url,
             "headers": tool_data.headers,
@@ -354,8 +374,32 @@ def connect_tool_servers_api(app: FastAPI):
         # Validate the tool server connectivity
         await validate_tool_server_connectivity(tool_server)
 
+        # Now the tool server is validate
+        # Remove the secret headers from the properties and store them in the configuration
+        non_secret_headers = {
+            key: tool_data.headers[key]
+            for key in tool_data.headers.keys()
+            if key not in tool_data.secret_header_keys
+        }
+
+        tool_server.properties["headers"] = non_secret_headers
+        tool_server.properties["secret_header_keys"] = tool_data.secret_header_keys
+
         # Save the tool to file
         tool_server.save_to_file()
+
+        # Store secret headers in the configuration before validation
+        if tool_data.secret_header_keys:
+            config = Config.shared()
+            mcp_secrets = config.get_value("mcp_secrets") or dict[str, str]()
+
+            # Store secrets with the pattern: mcp_server_id::header_name
+            for header_name in tool_data.secret_header_keys:
+                header_value = tool_data.headers[header_name]
+                secret_key = f"{tool_server.id}::{header_name}"
+                mcp_secrets[secret_key] = header_value
+
+            config.update_settings({"mcp_secrets": mcp_secrets})
 
         return tool_server
 
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -358,7 +358,7 @@ def connect_tool_servers_api(app: FastAPI):
             missing_secrets=[],
         )
 
-        # Check if the tool server has missing secretes (e.g. new user syncing exisitng project)
+        # Check if the tool server has missing secretes (e.g. new user syncing exisiting project)
         # If there are missing secrets, add a requirement to the result and skip getting available tools.
         missing_secrets = tool_server.missing_secrets()
         if missing_secrets and len(missing_secrets) > 0:
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -29,6 +29,7 @@ class KilnToolServerDescription(BaseModel):
     id: ID_TYPE
     type: ToolServerType
     description: str | None
+    missing_secrets: list[str]
 
 
 class ExternalToolServerCreationRequest(BaseModel):
@@ -333,6 +334,7 @@ def connect_tool_servers_api(app: FastAPI):
                 id=tool.id,
                 type=tool.type,
                 description=tool.description,
+                missing_secrets=tool.missing_secrets(),
             )
             for tool in project.external_tool_servers()
         ]
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -194,6 +194,7 @@ class ExternalToolServerApiDescription(BaseModel):
     created_by: str | None
     properties: Dict[str, Any]
     available_tools: list[ExternalToolApiDescription]
+    missing_secrets: list[str]
 
 
 class ToolApiDescription(BaseModel):
@@ -342,6 +343,27 @@ def connect_tool_servers_api(app: FastAPI):
     ) -> ExternalToolServerApiDescription:
         tool_server = tool_server_from_id(project_id, tool_server_id)
 
+        # Create a default result with incomplete status
+        result = ExternalToolServerApiDescription(
+            id=tool_server.id,
+            name=tool_server.name,
+            type=tool_server.type,
+            description=tool_server.description,
+            created_at=tool_server.created_at,
+            created_by=tool_server.created_by,
+            properties=tool_server.properties,
+            available_tools=[],
+            missing_secrets=[],
+        )
+
+        # Check if the tool server has missing secretes (e.g. new user syncing exisitng project)
+        # If there are missing secrets, add a requirement to the result and skip getting available tools.
+        missing_secrets = tool_server.missing_secrets()
+        if missing_secrets and len(missing_secrets) > 0:
+            result.missing_secrets = list(missing_secrets)
+            return result
+
+        # If there are no missing secrets, get available tools
         # Get available tools based on server type
         available_tools = []
         match tool_server.type:
@@ -358,16 +380,10 @@ def connect_tool_servers_api(app: FastAPI):
             case _:
                 raise_exhaustive_enum_error(tool_server.type)
 
-        return ExternalToolServerApiDescription(
-            id=tool_server.id,
-            name=tool_server.name,
-            type=tool_server.type,
-            description=tool_server.description,
-            created_at=tool_server.created_at,
-            created_by=tool_server.created_by,
-            properties=tool_server.properties,
-            available_tools=available_tools,
-        )
+        # Update the result with the available tools
+        result.available_tools = available_tools
+
+        return result
 
     @app.post("/api/projects/{project_id}/connect_remote_mcp")
     async def connect_remote_mcp(
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -384,10 +384,7 @@ def connect_tool_servers_api(app: FastAPI):
         # Validate the tool server connectivity
         await validate_tool_server_connectivity(tool_server)
 
-        # Store secret headers in the configuration after validation
-        tool_server.save_secrets()
-
-        # Save the tool to file (save_to_file will automatically strip secrets)
+        # Save the tool to file
         tool_server.save_to_file()
 
         return tool_server
@@ -418,10 +415,7 @@ def connect_tool_servers_api(app: FastAPI):
         MCPSessionManager.shared().clear_shell_path_cache()
         await validate_tool_server_connectivity(tool_server)
 
-        # Store secret env vars in the configuration after validation
-        tool_server.save_secrets()
-
-        # Save the tool to file (save_to_file will automatically strip secrets)
+        # Save the tool to file
         tool_server.save_to_file()
 
         return tool_server
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -115,6 +115,7 @@ class LocalToolServerCreationRequest(BaseModel):
     command: str
     args: List[str]
     env_vars: Dict[str, str] = Field(default_factory=dict)
+    secret_env_var_keys: List[str] = Field(default_factory=list)
 
     @model_validator(mode="after")
     def validate_command(self):
@@ -138,6 +139,24 @@ class LocalToolServerCreationRequest(BaseModel):
                 raise ValueError(
                     f"Invalid environment variable key: {key}. Can only contain letters, digits, and underscores."
                 )
+
+        # Validate secret environment variable keys
+        if isinstance(self.secret_env_var_keys, list):
+            validated_secret_env_var_keys = []
+            for key in self.secret_env_var_keys:
+                key_str = key.strip() if isinstance(key, str) else str(key).strip()
+                if not key_str:
+                    raise ValueError("Secret environment variable key is required")
+                # Check if the key is in the env_vars
+                if key_str not in self.env_vars:
+                    raise ValueError(
+                        f"Secret environment variable key {key_str} is not in the list of environment variables"
+                    )
+
+                validated_secret_env_var_keys.append(key_str)
+
+            # Update secret env var keys to validated version
+            self.secret_env_var_keys = validated_secret_env_var_keys
 
         return self
 
@@ -408,7 +427,7 @@ def connect_tool_servers_api(app: FastAPI):
     ) -> ExternalToolServer:
         project = project_from_id(project_id)
 
-        # Create the ExternalToolServer with required fields
+        # Create the ExternalToolServer with required fields, temporarily store all env vars in properties to validate the connectivity
         properties = {
             "command": tool_data.command,
             "args": tool_data.args,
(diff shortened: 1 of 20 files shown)


app/desktop/studio_server/tool_api.py
```diff
@@ -87,6 +89,24 @@ def validate_server_details(self):
             # Update headers to validated version
             self.headers = validated_headers
 
+        # Validate secret header keys
+        if isinstance(self.secret_header_keys, list):
```
scosman: little strange to check the type here. Pydantic will check the type automatically so it's not needed. Doesn't hurt but not needed.

app/desktop/studio_server/tool_api.py
```diff
@@ -87,6 +89,24 @@ def validate_server_details(self):
...
+            validated_secret_header_keys = []
+            for key in self.secret_header_keys:
+                key_str = key.strip() if isinstance(key, str) else str(key).strip()
+                if not key_str:
+                    raise ValueError("Secret header key is required")
+                # Check if the key is in the headers
+                if key_str not in self.headers:
```
scosman: validate key is in, not key_str (stripped). You'll get into weird error states where the value isn't actually valid: `    key` will pass validation but really shouldn't.

Striping is good to check empty, but don't use it here.

app/desktop/studio_server/tool_api.py
```diff
@@ -87,6 +89,24 @@ def validate_server_details(self):
...
+                    raise ValueError(
+                        f"Secret header key {key_str} is not in the headers"
+                    )
+
+                validated_secret_header_keys.append(key_str)
+
+            # Update secret header keys to validated version
+            self.secret_header_keys = validated_secret_header_keys
```
scosman: hard rule: no mutations in validators. 

Just check it's valid above.

In general: kill "magic". We could spend an hour debugging why something sometimes works, only to find the validator is magically fixing a bad value. Just error on the bad values and we'll catch the bug in 10 seconds.

scosman: If you wanted to check "no leading/trailing whitespace" then check that directly, and raise a nice error explaining the specific issue so use can fix. But never just hide the issue. P3 to do that though.

app/desktop/studio_server/tool_api.py
```diff
@@ -120,6 +141,24 @@ def validate_command(self):
...
+                # Check if the key is in the env_vars
+                if key_str not in self.env_vars:
+                    raise ValueError(
+                        f"Secret environment variable key {key_str} is not in the list of environment variables"
+                    )
+
+                validated_secret_env_var_keys.append(key_str)
+
+            # Update secret env var keys to validated version
+            self.secret_env_var_keys = validated_secret_env_var_keys
```
scosman: ditto to all of above

app/desktop/studio_server/tool_api.py
```diff
@@ -304,6 +345,27 @@ async def get_tool_server(
...
+            created_by=tool_server.created_by,
+            properties=tool_server.properties,
+            available_tools=[],
+            missing_secrets=[],
+        )
+
+        # Check if the tool server has missing secretes (e.g. new user syncing exisiting project)
+        # If there are missing secrets, add a requirement to the result and skip getting available tools.
+        missing_secrets = tool_server.missing_secrets()
+        if missing_secrets and len(missing_secrets) > 0:
```
scosman: len check and other checks unnecessary. If it's an array of zero that's valid. Just assign directly (assuming method doesn't return None, but just make it return empty array and not None if it does).

app/desktop/studio_server/tool_api.py
```diff
@@ -304,6 +345,27 @@ async def get_tool_server(
...
+        # Create a default result with incomplete status
+        result = ExternalToolServerApiDescription(
+            id=tool_server.id,
+            name=tool_server.name,
+            type=tool_server.type,
+            description=tool_server.description,
+            created_at=tool_server.created_at,
+            created_by=tool_server.created_by,
+            properties=tool_server.properties,
+            available_tools=[],
```
scosman: nit: I like the old way more. Having a partly instantiated but incorrect object could introduce buggy. Just calculate values first and build the object when you have it (like you had)

This pattern is sometimes needed with dicts, but not objects.

app/desktop/studio_server/tool_api.py
```diff
@@ -320,32 +382,27 @@ async def get_tool_server(
...
         properties = {
             "server_url": tool_data.server_url,
             "headers": tool_data.headers,
+            "secret_header_keys": tool_data.secret_header_keys,
         }
 
         tool_server = ExternalToolServer(
-            name=tool_data.name,
-            type=ToolServerType.remote_mcp,  # Default to remote MCP type
+            name=tool_data.name.strip(),  # Trim whitespace for filename validation
```
scosman: no magic edits. Let it error on invalid inputs.

app/desktop/studio_server/tool_api.py
```diff
@@ -365,15 +425,16 @@ async def connect_local_mcp(
...
         properties = {
             "command": tool_data.command,
             "args": tool_data.args,
             "env_vars": tool_data.env_vars,
+            "secret_env_var_keys": tool_data.secret_env_var_keys,
         }
 
         tool_server = ExternalToolServer(
-            name=tool_data.name,
+            name=tool_data.name.strip(),  # Trim whitespace for filename validation
```
scosman: no magic

app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -24,6 +25,8 @@
     <div class="flex items-center text-gray-500 overflow-x-hidden">
```
scosman: put error up here `<div class="flex items-center overflow-x-hidden {property.error ? 'text-error' : 'text-gray-500'}">`

that way you can have errors and links, and it's not either-or.

app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -24,6 +25,8 @@
...
       {#if property.link}
         <a href={property.link} class="link">{property.value}</a>
+      {:else if property.error}
```
scosman: remove per above

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -155,6 +171,9 @@
                     Disable
                   </button>
                 </td>
+                <td class="text-sm">
+                  <span class="text-success">✓ Ready</span>
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Use same same code so it remains visually consistent: 

```
 <Warning
                      warning_message="Ready"
                      warning_color="success"
                      warning_icon="check"
                    />
```

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -7,21 +7,28 @@
...
     server_url: string
-    headers: { key: string; value: string; placeholder: string | null }[]
+    headers: KeyValuePair[]
   }
 
-  interface LocalMcpServer {
-    name: string
-    subtitle: string
-    description: string
+  interface LocalMcpServer extends BaseMcpServer {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: of fancy. If you do move to types, syntax will change but same thing possible.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -74,11 +82,13 @@
           key: "Authorization",
           value: "apikey REPLACE_WITH_TWELVE_DATA_API_KEY",
           placeholder: "apikey REPLACE_WITH_TWELVE_DATA_API_KEY",
+          is_secret: true,
         },
         {
           key: "X-OpenAPI-Key",
           value: "",
           placeholder: "REPLACE_WITH_YOUR_OPENAI_API_KEY",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:85]: I swear we already addressed this in a PR, but these value placeholder strings need work.

Placeholders don't need to be all caps, shouldn't have "REPLACE_". Can just use nice plain English "Your Firecrawl API Key", "Your OpenAI API Key". The REPLACE is ugly, we only used it for cases where we anted to prepopulate the value because it needs a special format (e.g. "Bearer ...."). I think all placeholders need work.

Again: I swear I did this.... is there another PR pending?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -7,21 +7,28 @@
 
   $: project_id = $page.params.project_id
 
-  interface RemoteMcpServer {
+  interface KeyValuePair {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: not sure why you're using interface instead of type. I think here you have  more typical types? I'm not really sure here though, not an expert. This might be better.

Are TS interfaces more for real interfaces/protocols (things external objects implement)?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -11,17 +11,18 @@
   import Warning from "$lib/ui/warning.svelte"
 
   // Environment Variables as array of key/value pairs
-  interface EnvVarPair {
+  interface KeyValuePair {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: repeated code several places. 

You should define a types using a .ts file. It can live in /src/lib/... then import it in all 3 places

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -54,23 +55,37 @@
...
+    envVarsObj: Record<string, string>
+    secret_env_var_keys: string[]
+  } {
     const envVarsObj: Record<string, string> = {}
+    const secretEnvVarKeys: string[] = []
 
     for (const envVar of env_vars) {
       if (envVar.key.trim() && envVar.value.trim()) {
-        envVarsObj[envVar.key.trim()] = envVar.value.trim()
+        const key = envVar.key.trim()
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: no magic

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -54,23 +55,37 @@
...
-  function buildEnvVarsObject(): Record<string, string> {
+  function buildEnvVarsObject(): {
+    envVarsObj: Record<string, string>
+    secret_env_var_keys: string[]
+  } {
     const envVarsObj: Record<string, string> = {}
+    const secretEnvVarKeys: string[] = []
 
     for (const envVar of env_vars) {
       if (envVar.key.trim() && envVar.value.trim()) {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:60]: okay here since we're ignoring completely blank rows (good UI design in UI layer)

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -54,23 +55,37 @@
...
-        envVarsObj[envVar.key.trim()] = envVar.value.trim()
+        const key = envVar.key.trim()
+        envVarsObj[key] = envVar.value.trim()
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: no magic (especially since spaces are valid values)

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -208,6 +225,20 @@
...
           </div>
+          <div class="flex-1 max-w-[100px]">
+            <FormElement
+              inputType="select"
+              label="Secret"
+              id="secret_{item_index}"
+              info_description="If this environment variable is a secret such as an API key, select 'Yes' to prevent it from being synced. Kiln will store the secret in your project's settings."
+              light_label={true}
+              select_options={[
+                [false, "No"],
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte]: "Secret" and "Not secret" (and remove "select 'yes'" above)

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -15,13 +15,14 @@
   let description = ""
 
   // Headers as array of key/value pairs
-  interface HeaderPair {
+  interface KeyValuePair {
     key: string
     value: string
     placeholder: string | null
+    is_secret: boolean
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: ditto

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -46,24 +47,36 @@
...
+    headersObj: Record<string, string>
+    secret_header_keys: string[]
+  } {
     const headersObj: Record<string, string> = {}
+    const secretHeaderKeys: string[] = []
 
     for (const header of headers) {
       if (header.key.trim() && header.value.trim()) {
-        headersObj[header.key.trim()] = header.value.trim()
+        const key = header.key.trim()
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: no magic

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -46,24 +47,36 @@
...
+        headersObj[key] = header.value.trim()
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: no magic (and headers can have spaces)

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte
```diff
@@ -179,6 +194,20 @@
...
           </div>
+          <div class="flex-1 max-w-[100px]">
+            <FormElement
+              inputType="select"
+              label="Secret"
+              id="secret_{item_index}"
+              info_description="If this header is a secret such as an API key, select 'Yes' to prevent it from being synced. Kiln will store the secret in your project's settings."
+              light_label={true}
+              select_options={[
+                [false, "No"],
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/+page.svelte]: ditto

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -134,6 +134,67 @@
     return properties
   }
 
+  type UiProperty = {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: don't re-define types. This already exists. (I see this mistake is made elsewhere, but don't copy that)

1) Two options: move it into exported type in a .ts file next to property_list.svelte, import it both here and in property_list.svelte
2) Just let typescript do its thing. Don't define the return value. Typescript should dynamically infer it, and give you warnings if the data is incorrect. Check this but should work.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -134,6 +134,67 @@
...
+    const properties: UiProperty[] = []
+    properties.push(
+      ...Object.entries(noneSecretProperties).map(([key, value]) => ({
+        name: key,
+        value: String(value ?? "N/A"),
+      })),
+    )
+
+    // Make sure secretKeys is an array of strings
+    const secretKeysArray: string[] = Array.isArray(secretKeys)
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: don't dynamically type check unless needed -- the method specifies an array of strings, so it will be. Type checker will catch callers passing the wrong thing. Keep the code as typed as possible. Use dynamic type checking to make it typed asap, but not constantly validating after it's typed.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
```diff
@@ -134,6 +134,67 @@
...
+
+  /**
+   * Helper function to build properties list with secret handling
+   * @param noneSecretProperties - Object containing non-secret key-value pairs
+   * @param secretKeys - Array of secret key names
+   * @param missingSecrets - Array of missing secret key names
+   * @returns Array of UiProperty objects
+   */
+  function buildPropertiesWithSecrets(
+    noneSecretProperties: Record<string, string>,
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: nit: nonSecretProperties, not none

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -62,6 +63,16 @@ def validate_required_fields(self) -> "ExternalToolServer":
                     raise ValueError(
                         "headers must be a dictionary for external tools of type 'remote_mcp'"
                     )
+
+                secret_header_keys = self.properties.get("secret_header_keys", None)
+                # Secret header keys are optional, but if they are set, they must be a list of strings
```
scosman, line 101: you're checking list, but not strings

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -83,7 +94,182 @@ def validate_required_fields(self) -> "ExternalToolServer":
                         "environment variables must be a dictionary for external tools of type 'local_mcp'"
                     )
 
+                secret_env_var_keys = self.properties.get("secret_env_var_keys", None)
```
scosman, line 131: ditto

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -83,7 +94,182 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+            # Look for secrets with the pattern: mcp_server_id::key_name
+            if mcp_secrets:  # Only proceed if mcp_secrets is not None
+                for key_name in secret_keys:
+                    secret_value = mcp_secrets.get(self._config_secret_key(key_name))
+                    if secret_value:
+                        secrets[key_name] = secret_value
+
+        return secrets
+
+    def missing_secrets(self) -> list[str]:
```
scosman: nit: almost same method as above. Combine, and do tuple return Tuple[secrets, missing_secrets]

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -83,7 +94,182 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+        secret_keys = self.get_secret_keys()
+
+        if not secret_keys:
+            return
+
+        if self.id is None:
+            raise ValueError("Server ID cannot be None when saving secrets")
+
+        # Extract secret values from properties based on server type
+        match self.type:
```
scosman: use `get_secret_keys`

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -83,7 +94,182 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+            case ToolServerType.remote_mcp:
+                secret_values = self.properties.get("headers", {})
+            case ToolServerType.local_mcp:
+                secret_values = self.properties.get("env_vars", {})
+            case _:
+                raise_exhaustive_enum_error(self.type)
+
+        config = Config.shared()
+        mcp_secrets = config.get_value(MCP_SECRETS_KEY) or dict[str, str]()
```
scosman: use explicit type definition on var instead of the `dict[str, str]()` to give it implicit type

`mcp_secrets: dict[str, str] = config.get_value(MCP_SECRETS_KEY) or {}`

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -83,7 +94,182 @@ def validate_required_fields(self) -> "ExternalToolServer":
...
+        # Remove secrets with the pattern: mcp_server_id::key_name
+        for key_name in secret_keys:
+            secret_key = self._config_secret_key(key_name)
+            if secret_key in mcp_secrets:
+                del mcp_secrets[secret_key]
+
+        # Always call update_settings to maintain consistency with the old behavior
+        config.update_settings({MCP_SECRETS_KEY: mcp_secrets})
+
+    def save_to_file(self) -> None:
```
scosman, line 252: You end up having 2 inconsistent object states for the same object: before save secrets are in headers, after they aren't. Callers could easily have bugs.

The object also doesn't work unless functions are called in the right order (if you call save to file before save_secrets, the secrets are already deleted).

Suggestion: Process everything in `__init__` instead. Convert to the correct representation right away (no secrets in header/env_vars, just keys in secret_header_keys/secret_env_var_keys). You can store the secrets in memory in an internal non-model variable by setting `this._unsaved_secrets = {...}` in init. Still override save_to_file, but just to check if _unsaved_secrets isn't empty, and then call  internal _save_secrets (remove the public save_secrets). That way you can't forget and just save file, or secrets, but not both. Might need to update retrieve_secrets to merge config secrets and local `this._unsaved_secrets`

app/desktop/studio_server/tool_api.py
aiyakitori:
```
def missing_secrets(self) -> list[str]:
```
This returns empty array guaranteed.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte]: used types for consistency

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/tool_servers/[tool_server_id]/+page.svelte]: Did 1, i had this initially.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:85]: Didn't see any pending PRs around this. We can open a small PR to fix the strings?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte
```diff
@@ -208,6 +218,20 @@
...
+              id="secret_{item_index}"
+              info_description="If this environment variable is a secret such as an API key, select 'Secret' to prevent it from being synced. Kiln will store the secret in your project's settings."
+              light_label={true}
+              select_options={[
+                [false, "No Secret"],
+                [true, "Secret"],
+              ]}
+              bind:value={env_vars[item_index].is_secret}
+            />
+          </div>
```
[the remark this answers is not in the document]
Comment (aiyakitori) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/+page.svelte:237]: We explicitly want a dropdown so it looks better.

scosman (approved):

PR #564, opened 2025-09-04
Tools UI fixes

scosman:
- clear the browser cache when adding/deleting tools
 - Fix how we define "empty" tools to include demo tools.


aiyakitori (approved):

leonardmq (commented):
We have a similar problem with AI Providers. When you connect a new one, it does not show up in the `Run` UI dropdown until you force a refresh if the list of models was already cached

scosman:
@leonardmq I'll add it to the list, but feel free to take that one!

PR #566, opened 2025-09-04
fix: invalidate available model list after adding or removing a provider

leonardmq:
## What does this PR do?

Similar problem as https://github.com/Kiln-AI/Kiln/pull/564

Bug:
1. Don't have OpenAI connected (hard refresh if you have just disconnected it)
2. Navigate to the `Run` page, check out the models in the dropdown in the right sidebar
3. Click on `Settings` in the nav sidebar to the left 
4. Go into `AI Providers`
5. Connect OpenAI, set an API key and connect
6. Go back into `Run` (via the sidebar)
7. `OpenAI` models are not showing up in the model dropdown in the sidebar to the left because the model list has not been refreshed

Fix:
- invalidate the model list in the store after adding / removing any provider

## Summary by CodeRabbit

* **Bug Fixes**
  * The available models list now refreshes immediately after connecting or disconnecting providers (including Ollama, Docker Model Runner, and OpenAI-compatible).
  * Updating provider settings—such as submitting API keys, adding new OpenAI-compatible providers, or removing them—now correctly updates the models list.
  * Prevents stale or outdated model listings after any provider connectivity or configuration change, ensuring users always see the current set of available models.

### DIFF
--- a/app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
+++ b/app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
@@ -11,6 +11,7 @@
   import { client, base_url } from "$lib/api_client"
   import Warning from "$lib/ui/warning.svelte"
   import { available_tuning_models } from "$lib/stores/fine_tune_store"
+  import { clear_available_models_cache } from "$lib/stores"
   import { get_provider_image } from "$lib/ui/provider_image"
   import posthog from "posthog-js"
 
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(fullscreen)/setup/(setup)/connect_providers/connect_providers.svelte
```diff
@@ -396,6 +397,8 @@
 
       // Clear the available models list
       available_tuning_models.set(null)
+      // Clear the available models cache so it refreshes next time
+      clear_available_models_cache()
```
leonardmq, line 401: @scosman - this one handles the disconnect case (unless I miss a different one elsewhere?)

PR #570, opened 2025-09-05
Edit tools

scosman:
Screenshot 2025-09-05 at 11 26 21 AM

## Summary by CodeRabbit

- New Features
  - Add ability to edit Local and Remote MCP tool servers via new PATCH API and dedicated UI editors.
  - Introduce “Edit Tool Server” page and an Edit button on tool server detail pages.
  - Improve forms with secret env vars/headers handling, validation, placeholders, and warnings; missing secret properties now link directly to the edit page.

- Tests
  - Add end-to-end coverage for editing flows, including success, 404 (not found), and 400 (type mismatch) cases.

- Chores
  - Add --staged-only to checks.sh; pre-commit now runs staged-only checks for faster commits.

### DIFF
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -455,7 +455,6 @@ def connect_tool_servers_api(app: FastAPI):
         tool_server = existing_tool_server
         tool_server.name = tool_data.name
         tool_server.description = tool_data.description
-        # TODO: the setter here isn't updating the properties for the for secrets. Need to make that logic reusable
         tool_server.properties = _local_tool_server_properties(tool_data)
 
         # Validate the tool server connectivity
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
@@ -65,7 +65,7 @@
     name = editing_tool_server.name
     description = editing_tool_server.description || ""
 
-    // TODO: fix the server type, so it's not "never" and we don't need a cast
+    // We should fix the server type, so it's not "never" and we don't need a cast
     const props = editing_tool_server.properties as Record<string, unknown>
 
     if (props.command && typeof props.command === "string") {
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte
@@ -55,7 +55,7 @@
     name = editing_tool_server.name
     description = editing_tool_server.description || ""
 
-    // TODO: fix the server type, so it's not "never" and we don't need a cast
+    // We should fix the server type, so it's not "never" and we don't need a cast
     const props = editing_tool_server.properties as Record<string, unknown>
 
     if (props.server_url && typeof props.server_url === "string") {


libs/core/kiln_ai/datamodel/test_exteral_tool_server.py
```diff
@@ -1,695 +0,0 @@
-import pytest
```
aiyakitori: I removed this in https://github.com/Kiln-AI/Kiln/pull/573/files. Should pull tools and see if there are conflicts with test_external_tool_server.py as well.

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -74,6 +84,16 @@ def __init__(self, **kwargs):
...
 
+    def __setattr__(self, name: str, value: Any) -> None:
+        """
+        Override __setattr__ to process secrets whenever properties are updated.
+        """
+        super().__setattr__(name, value)
+
+        # Process secrets whenever properties are updated
+        if name == "properties":
+            self._process_secrets_from_properties()
```
aiyakitori: I think one of the issue for this approach is modification of the properties's value wouldn't trigger the setter. i.e. `externalToolServer.properties["new secret"] = "secret"`, the setter is only triggered if a different dictionary object is set. 

Is this a corner case that we should care about? Since both create and update we will use new dictionaries?

app/desktop/studio_server/tool_api.py
```diff
@@ -420,6 +441,42 @@ async def connect_local_mcp(
...
+        if existing_tool_server.type != ToolServerType.local_mcp:
+            raise HTTPException(
+                status_code=400,
+                detail="Existing tool server is not a local MCP server. You can't edit a non-local MCP server with this endpoint.",
+            )
+
+        tool_server = existing_tool_server
+        tool_server.name = tool_data.name
+        tool_server.description = tool_data.description
+        # TODO: the setter here isn't updating the properties for the for secrets. Need to make that logic reusable
```
aiyakitori:
```suggestion
        # TODO: the setter here isn't updating the properties for the secrets. Need to make that logic reusable
```

aiyakitori: This should be supported now on tools via #573

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
```diff
@@ -0,0 +1,362 @@
...
+    } catch (e) {
+      error = createKilnError(e)
+    } finally {
+      submitting = false
+    }
+  }
+</script>
+
+<div>
+  {#if installation_instruction}
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte:239]: @scosman I thought the approach is to make the following another component, but i guess it's not needed since it's only used at this place?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte
```diff
@@ -0,0 +1,311 @@
...
+
+  function load_existing_tool_server() {
+    if (!editing_tool_server) {
+      return
+    }
+    name = editing_tool_server.name
+    description = editing_tool_server.description || ""
+
+    // TODO: fix the server type, so it's not "never" and we don't need a cast
+    const props = editing_tool_server.properties as Record<string, unknown>
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte:59]: This comes from schema, anyway we could have the schema to not have unknown?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
```diff
@@ -0,0 +1,362 @@
...
+      env_vars = env_vars
+    }
+
+    // Load the secret env vars to the form
+    let api_secret_env_vars = props.secret_env_var_keys
+    if (api_secret_env_vars && Array.isArray(api_secret_env_vars)) {
+      for (const key of api_secret_env_vars) {
+        env_vars.push({
+          key: key,
+          value: "SECRET_VALUE",
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte:109]: Is the convention never return the actual value back to the UI? Our current API doesn't return the value of the key but i wonder if this is something we should support?

libs/core/kiln_ai/datamodel/external_tool_server.py
scosman: this was from your PR, just merged in here.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
```diff
@@ -0,0 +1,362 @@
...
+        warning_icon="info"
+        large_icon={true}
+        warning_message={installation_instruction}
+      />
+    </div>
+  {/if}
+  {#if editing_requires_secrets}
+    <div class="mb-6 max-w-lg">
+      <Warning
+        warning_message="This server requires secrets, such as API keys. You must enter them below wherever it says 'SECRET_VALUE' to edit this server."
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte:252]: If a user just want to update say the description or arguments of the tool server this would still require them to dig out the keys and enter them, is this by design? or something we should fix in the future?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte:109]: We don't have auth on our API, do I don't want to make any APIs that return secrets. Agree it's not the best UX, but it's pretty niche, not every day thing.

libs/core/kiln_ai/datamodel/external_tool_server.py
aiyakitori: I realized we might have a corner case there. Probably something we should fix, thoughts?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/remote_mcp/edit_remote_tool.svelte:59]: Properties is broken in the schema by using "Any" in python, it ends up being "never" in TS. One of the refactors I want to do

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/local_mcp/edit_local_tool.svelte:252]: Not ideal, but it would be a much bigger project to have partial update. We can add it to the backlog, but not that critical either.

PR #571, opened 2025-09-05
Better error messages when connecting tool servers 

aiyakitori:
## What does this PR do?

"unhandled errors in a TaskGroup" is often returned back to UI when failing to connect to mcp servers. Catch common errors for both remote & local mcp servers and return user friendly error messages. 

## Checklists

## Summary by CodeRabbit

* **Bug Fixes**
  * Clear, user-friendly errors during remote/local MCP session setup (HTTP status, connectivity, startup failures), and preserved original error causes.
  * Prevent unintended mutation when merging secret headers and environment variables.
  * More reliable PATH and shell resolution across platforms.

* **Tests**
  * Extensive coverage for error mapping, nested/aggregated exceptions, secret/header/env merging, session creation scenarios, and path handling.

* **Chores**
  * Conditional dependency added for ExceptionGroup compatibility on older Python versions.

### DIFF
--- a/libs/core/kiln_ai/tools/mcp_session_manager.py
+++ b/libs/core/kiln_ai/tools/mcp_session_manager.py
@@ -16,13 +16,6 @@ from kiln_ai.datamodel.external_tool_server import ExternalToolServer, ToolServe
 from kiln_ai.utils.config import Config
 from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 
-# Import ExceptionGroup with proper fallback for Python 3.10
-try:
-    from exceptiongroup import ExceptionGroup  # type: ignore[import-untyped]
-except ImportError:
-    # Python 3.11+ has ExceptionGroup in builtins
-    pass
-
 logger = logging.getLogger(__name__)
 
 
@@ -61,18 +54,24 @@ class MCPSessionManager:
                 raise_exhaustive_enum_error(tool_server.type)
 
     def _extract_first_exception(
-        self, exception: Exception, type: type | tuple[type, ...]
+        self, exception: Exception, target_type: type | tuple[type, ...]
     ) -> Exception | None:
         """
         Extract first relevant exception from ExceptionGroup or handle direct exceptions
         """
-        if isinstance(exception, type):
+        # Check if the exception itself is of the target type
+        if isinstance(exception, target_type):
             return exception
-        if isinstance(exception, ExceptionGroup):
-            for nested_exc in exception.exceptions:  # type: ignore[attr-defined]
-                result = self._extract_first_exception(nested_exc, type)
-                if result:
-                    return result
+
+        # Handle ExceptionGroup
+        if hasattr(exception, "exceptions"):
+            exceptions_attr = getattr(exception, "exceptions", None)
+            if exceptions_attr:
+                for nested_exc in exceptions_attr:
+                    result = self._extract_first_exception(nested_exc, target_type)
+                    if result:
+                        return result
+
         return None
 
     @asynccontextmanager
@@ -111,46 +110,26 @@ class MCPSessionManager:
             # Check for HTTPStatusError
             http_error = self._extract_first_exception(e, httpx.HTTPStatusError)
             if http_error and isinstance(http_error, httpx.HTTPStatusError):
-                # Create user-friendly error message based on status code
-                status_code = http_error.response.status_code
-                if status_code == 400:
-                    raise ValueError(
-                        "The MCP server rejected the request. Please verify your authentication token and server URL, and consult the server documentation for the correct request format."
-                    )
-                elif status_code == 401:
-                    raise ValueError(
-                        "Authentication to the MCP server failed. Please verify your token is valid and has not expired, and consult the server documentation if needed."
-                    )
-                elif status_code == 403:
-                    raise ValueError(
-                        "Access to the MCP server is forbidden. Please ensure your authentication token has the required permissions."
-                    )
-                elif status_code == 404:
-                    raise ValueError(
-                        "MCP server not found. Please verify the server URL is correct and the server is running."
-                    )
-                elif status_code >= 500:
-                    raise ValueError(
-                        "The MCP server encountered an internal error. This is a server-side problem — please try again later or contact the server administrator."
-                    )
-                else:
-                    raise ValueError(
-                        "Failed to connect to the MCP server. Please verify the server URL and authentication settings, and consult the server documentation for troubleshooting."
-                    )
-
-            # Check for connection errors
-            connection_error = self._extract_first_exception(
-                e, (ConnectionError, OSError)
-            )
-            if connection_error and isinstance(
-                connection_error, (ConnectionError, OSError)
-            ):
                 raise ValueError(
-                    f"Unable to connect to MCP server due to: '{connection_error}'. Please verify the configurations are correct, the server is running, and your network connection is working."
+                    f"The MCP server rejected the request. "
+                    f"Status {http_error.response.status_code}. "
+                    f"Response from server:\n{http_error.response.reason_phrase}"
                 )
 
+            # Check for connection errors
+            connection_error_types = (ConnectionError, OSError, httpx.RequestError)
+            connection_error = self._extract_first_exception(e, connection_error_types)
+            if connection_error and isinstance(
+                connection_error, connection_error_types
+            ):
+                raise RuntimeError(
+                    f"Unable to connect to MCP server. Please verify the configurations are correct, the server is running, and your network connection is working. Original error: {connection_error}"
+                ) from e
+
             # If no known error types found, re-raise the original exception
-            raise
+            raise RuntimeError(
+                f"Failed to connect to the MCP Server. Check the server's docs for troubleshooting. Original error: {e}"
+            ) from e
 
     @asynccontextmanager
     async def _create_local_mcp_session(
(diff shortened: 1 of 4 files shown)


libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -51,6 +60,21 @@ async def mcp_client(
...
+    def _extract_first_exception(
+        self, exception: Exception, type: type | tuple[type, ...]
+    ) -> Exception | None:
+        """
+        Extract first relevant exception from ExceptionGroup or handle direct exceptions
+        """
+        if isinstance(exception, type):
+            return exception
+        if isinstance(exception, ExceptionGroup):
+            for nested_exc in exception.exceptions:  # type: ignore[attr-defined]
```
scosman: no type-ignore outside of tests. Use get then verify type.

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -5,15 +5,24 @@
...
 from kiln_ai.datamodel.external_tool_server import ExternalToolServer, ToolServerType
 from kiln_ai.utils.config import Config
 from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 
+# Import ExceptionGroup with proper fallback for Python 3.10
+try:
+    from exceptiongroup import ExceptionGroup  # type: ignore[import-untyped]
+except ImportError:
+    # Python 3.11+ has ExceptionGroup in builtins
+    pass
```
scosman: Can we do this without exception for normal cases?

Maybe:

```
import builtins

if hasattr(builtins, "ExceptionGroup"):
    ExceptionGroup = builtins.ExceptionGroup
else:
    from exceptiongroup import ExceptionGroup
```

libs/core/kiln_ai/tools/mcp_session_manager.py
scosman: But also, you check type above so this shouldn't be needed. Something is off. How you import might be breaking the type system.

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -71,15 +95,62 @@ async def _create_remote_mcp_session(
...
+                    raise ValueError(
+                        "MCP server not found. Please verify the server URL is correct and the server is running."
+                    )
+                elif status_code >= 500:
+                    raise ValueError(
+                        "The MCP server encountered an internal error. This is a server-side problem — please try again later or contact the server administrator."
+                    )
+                else:
+                    raise ValueError(
+                        "Failed to connect to the MCP server. Please verify the server URL and authentication settings, and consult the server documentation for troubleshooting."
```
scosman: headers, not "authentication settings". We call them just headers in our UI, they aren't necessarily auth.

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -71,15 +95,62 @@ async def _create_remote_mcp_session(
...
+            # Handle HTTP errors with user-friendly messages
+
+            # Check for HTTPStatusError
+            http_error = self._extract_first_exception(e, httpx.HTTPStatusError)
+            if http_error and isinstance(http_error, httpx.HTTPStatusError):
+                # Create user-friendly error message based on status code
+                status_code = http_error.response.status_code
+                if status_code == 400:
+                    raise ValueError(
+                        "The MCP server rejected the request. Please verify your authentication token and server URL, and consult the server documentation for the correct request format."
```
scosman: I'd be okay using 1 error for all HTTP errors to simplify code. Our message, then the status code, then `http_error.response.text`. That what if the server has helpful errors, they get passed through. The user might need the code and text to debug.

"The MCP server rejected the request. Status {http_error.response.status_code}. Response from server:\n {http_error.response.text}"

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -71,15 +95,62 @@ async def _create_remote_mcp_session(
...
+                    )
+
+            # Check for connection errors
+            connection_error = self._extract_first_exception(
+                e, (ConnectionError, OSError)
+            )
+            if connection_error and isinstance(
+                connection_error, (ConnectionError, OSError)
+            ):
+                raise ValueError(
```
scosman: this isn't a ValueError (only for bad values), use RuntimeError.

`raise RuntimeError("Unable to connect to MCP server. Please verify the configurations are correct, the server is running, and your network connection is working. Original error: {e}") from. e`

note the "from e" which improves stack traces

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -71,15 +95,62 @@ async def _create_remote_mcp_session(
...
+                    f"Unable to connect to MCP server due to: '{connection_error}'. Please verify the configurations are correct, the server is running, and your network connection is working."
+                )
+
+            # If no known error types found, re-raise the original exception
+            raise
```
scosman: We should still wrap it. The original exception could be anything (like the initial bug, totally useless to the user). We can say something better than that about the issue being starting the MCP server.

to raise and wrap pattern:

`raise RuntimeError("Failed to connect to the MCP Server. Check the server's docs for troubleshooting. Original error: {e}") from e)`

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -119,10 +190,31 @@ async def _create_local_mcp_session(
...
+            file_not_found = self._extract_first_exception(e, FileNotFoundError)
+            if file_not_found and isinstance(file_not_found, FileNotFoundError):
+                raise ValueError(
+                    f"Command '{command}' not found. Please follow the instructions in the MCP server documentation to install it."
+                )
+
+            # Check for MCP errors (from session initialization)
+            mcp_error = self._extract_first_exception(e, McpError)
+            if mcp_error and isinstance(mcp_error, McpError):
+                error_message = str(mcp_error).strip()
```
scosman: no casting ever, but especially here where we know the type isn't string. Use the wrap pattern above

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -119,10 +190,31 @@ async def _create_local_mcp_session(
...
+                raise ValueError(
+                    f"MCP server failed to start due to: '{error_message}'. Please verify your command, arguments, and environment variables, and consult the server's documentation for the correct setup."
+                )
+
+            # If no known error types found, re-raise the original exception
+            raise
```
scosman: use the wrap pattern above.

Also: I don't think it's worth a special case for `McpError`. We don't say anything there that we wouldn't also say here.

`raise RuntimeError("MCP server failed to start. Please verify your command, arguments, and environment variables, and consult the server's documentation for the correct setup.") from e`

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -119,10 +190,31 @@ async def _create_local_mcp_session(
...
+            async with stdio_client(server_params) as (read, write):
+                async with ClientSession(read, write) as session:
+                    await session.initialize()
+                    yield session
+        except Exception as e:
+            # Handle local MCP server startup errors with user-friendly messages
+
+            # Check for FileNotFoundError (command not found)
+            file_not_found = self._extract_first_exception(e, FileNotFoundError)
+            if file_not_found and isinstance(file_not_found, FileNotFoundError):
```
scosman: Seems a bit general. FileNotFoundError might be a command issue, but might not be. We're assuming all filesystem issues should get that message, but it's not true.

Is this needed? The message seemed okay for this one already. This is better when it's right (but it might be wrong).

If you want to keep it I think you need to check specifically the error is "No such file or directory: 'command'" and not any FileNotFoundError. Or leave the general handler below.

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -119,10 +190,31 @@ async def _create_local_mcp_session(
...
+                raise ValueError(
+                    f"Command '{command}' not found. Please follow the instructions in the MCP server documentation to install it."
+                )
+
+            # Check for MCP errors (from session initialization)
+            mcp_error = self._extract_first_exception(e, McpError)
+            if mcp_error and isinstance(mcp_error, McpError):
+                error_message = str(mcp_error).strip()
+                raise ValueError(
```
scosman: remove McpError check? It doesn't say anything we're not going to include in the general version below.

libs/core/kiln_ai/tools/test_mcp_session_manager.py
```diff
@@ -37,6 +47,176 @@ def test_singleton_reset_for_testing(self):
...
+
+        # Should not find non-existent exception type
+        result = manager._extract_first_exception(outer_group, KeyError)
+        assert result is None
+
+        # Should not find another unrelated exception type
+        result = manager._extract_first_exception(outer_group, LookupError)
+        assert result is None
+
+    def test_extract_first_exception_empty_exception_group(self):
```
scosman: there's too much test code here for the use case. Did you read all of it and are signing up to maintain it all? No point in tests unless we're reviewed them and verify they all check the right things. Otherwise it's just maintenance cost. Sonnet will write tests that pass, but don't check important things or check incorrect things.

To eliminate repeated code, use pytest paramterize. Write the method once, and takes 2 parameters: the error to throw, and a string to check the resulting error matches. Should eliminate 90% of the code.

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: I like this better as well.

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: I'm more in the camp of not handling this. This was just an extra case when I typed the npx command wrong and the error message was a bit too vague. We can remove this and use more general error for this case.

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: This is needed in the case where MCP fails to start the process i.e. wrong argument. If we don't extract the error and fallback to the general version you often would get ` unhandled errors in a TaskGroup (1 sub-exception)` which is the reason why we added error handling the first place. 

If we handle the error, we would get "Connection closed" instead and that error comes from initializing the server.

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: .text cannot be read because it has already been closed when the error is caught. Used `response.reason_phrase` instead.

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: Actually if we use this approach below we don't need the imports at all. 
```
        # Handle ExceptionGroup
        if hasattr(exception, "exceptions"):
            exceptions_attr = getattr(exception, "exceptions", None)
            if exceptions_attr:
```

scosman (approved):

PR #583, opened 2025-09-08
 Add ability to set a default run config ID to a Task

sfierro:
## Summary by CodeRabbit

- **New Features**
  - Tasks now support an optional default run configuration ID, letting a task have a preselected run configuration when executed.
  - The API surface exposes this as an optional string/null field on Task; behavior is unchanged when unset.

- **Tests**
  - Added unit tests to validate reading, setting, and clearing a task’s default run configuration ID.


scosman (approved):
this is fine, but no UI?

PR #587, opened 2025-09-09
Update sample server instruction 

aiyakitori:
## What does this PR do?

Update sample mcp server instruction to remind user to restart Kiln so PATH would take affect. 

## Checklists

## Summary by CodeRabbit

* **Documentation**
  * Clarified setup instructions for select local tools: Web Search & Scrape (Node.js), Run Python Code (deno), and Access Files (Node.js).
  * Added guidance to restart Kiln after installing required runtimes (Node.js or Deno) before connecting these servers.
  * Remote servers and other tools are unaffected; no functional behavior changes—only improved on-screen installation guidance.

### DIFF
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
@@ -105,7 +105,7 @@
         },
       ],
       installation_instruction:
-        "You must have Node.js installed: https://nodejs.org",
+        "You must have Node.js installed: https://nodejs.org. If you had to install node, restart Kiln before connecting the server.",
     },
     {
       name: "Run Python Code",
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
```diff
@@ -105,7 +105,7 @@
         },
       ],
       installation_instruction:
-        "You must have Node.js installed: https://nodejs.org",
+        "You must have Node.js installed: https://nodejs.org. If you had to install node, restart Kiln before connecting the server.",
     },
```
Comment (aiyakitori) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:109]: Do we also want the local mcp page to include this permanently on the page (probably not)? Something like "`If you had to install new package manager like Node.js, restart Kiln before connecting the server.`"

scosman (approved):

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/+page.svelte:109]: Naaaa. It's confusing 90% of the time, helpful 10%. At least here there's context.

PR #597, opened 2025-09-11
New Synth Data UI

scosman:
- Uses a table
 - Proper menus and dropdown
 - Allows previewing outputs before saving, so you can curate output generation as well
 - Add links to docs
 - New "steps" UI and state

## Summary by CodeRabbit

* **New Features**
  * Two-stage data flow: generate preview samples then explicitly save; generated items tagged "synthetic" and optionally "synthetic_session_{id}".
  * New stepper-based Generate UI with per-step controls, progress dialogs, table-style sample list, expand/preview and per-row actions.

* **API / Schema**
  * Added generate_sample (optional session_id) and save_sample endpoints; client schema/types expose generated output type.

* **Public Changes**
  * Field renamed human_guidance → guidance; project path accepts Path-like values; samples now include an output field.


scosman:
@coderabbit review

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -221,6 +230,31 @@
     // Case 5: No state - wait for the user to setup via UI
   }
 
+  function load_initial_step() {
+    // Load which step to start on based on status
+
+    // Nothing! Start at step 1 for topics
+    if ($saved_state.root_node.sub_topics.length === 0) {
+      set_current_step(1)
```
Comment (sfierro) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:238]: This was a little annoying when I started generating without topics and was on step 3 and went out then back in and it brought me to step 1, not sure if there's a flag we can set to know if the user wanted to skip topics from the get go or not. But also not a big deal if we don't fix if we really want them to generate topics.

PR #598, opened 2025-09-12
feat: llama_index pypdf extractor

leonardmq:
## What does this PR do?

The model extractor for PDFs has noise (e.g. starting transcriptions with a preface like "Here's your transcription") and the splitting by page may cause some issues - seems to be rare but I had a case with one corrupted page of a PDF making the PDF fail.

This PR adds the llama_index's `PDFReader` that uses `pypdf` (which we already had installed to do per-page extraction).

WIP:
- work on the UI to allow picking non-models in the dropdown

## Checklists

## Summary by CodeRabbit

* New Features
  * Choose extractor type when creating extractor configs, including a new Llama PDF Reader.
  * Added support for the “nomic_text_embedding_v1_5” embedding model.
* API
  * Creating extractor configs now requires an extractor_type in the request.
  * Model/provider fields may be optional based on the selected extractor type.
* UI
  * Create Extractor form now submits the selected extractor type.
* Validation
  * Improved handling and clear errors for invalid extractor types.

### DIFF
--- a/app/web_ui/src/lib/api_schema.d.ts
+++ b/app/web_ui/src/lib/api_schema.d.ts
@@ -1884,6 +1884,8 @@ export interface components {
              * @description The description of the extractor config
              */
             description?: string | null;
+            /** @description The type of extractor to use */
+            extractor_type: components["schemas"]["ExtractorType"];
             /** @description The name of the model provider to use for the extractor config. */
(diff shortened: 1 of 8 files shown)


leonardmq:
Might do later - for now we have some local alternatives via Ollama

PR #604, opened 2025-09-15
Run Options dropdown to select from saved / default

sfierro:
## Summary by CodeRabbit

- New Features
  - Centralized Run Configuration component with save/set-default actions, advanced options, and per-task caching.
  - Inline action button in form fields; detailed model/prompt display helpers.
- UI Improvements
  - Empty states for selects with message/subtitle/link.
  - Tools & Search always visible with clearer empty/ loading states.
  - Model/prompt selectors accept task context; “Eval Prompt” renamed to “Frozen Prompt”.
- Changes
  - Shortened frozen prompt description in app and tests.
- Refactor
  - Pages migrated to shared stores/components; import paths consolidated.
- Tests
  - Updated expectations accordingly.
- Chores
  - Misspell check excludes build directory.


app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -176,6 +180,179 @@
...
+            project_id: $current_project?.id ?? "",
+            task_id: $current_task?.id ?? "",
+          },
+        },
+        body: {
+          // @ts-expect-error openapi-fetch generates the wrong type for this: Record<string, never>
+          default_run_config_id: (selected_run_config as TaskRunConfig).id,
+        },
+      })
+      await load_current_task($current_project)
```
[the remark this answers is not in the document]
sfierro: this code rabbit comment is wrong

app/web_ui/src/lib/stores/run_configs_store.ts
```diff
@@ -0,0 +1,136 @@
...
+>({})
+
+// Promise map to avoid parallel requests for run configs per task
+const loading_task_run_configs: Record<string, Promise<void>> = {}
+
+export async function load_task_run_configs(
+  project_id: string,
+  task_id: string,
+  force_refresh: boolean = false,
+) {
```
scosman: missing return type

app/web_ui/src/lib/stores/run_configs_store.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      // Update the store with the new data for this specific task using composite key
+      run_configs_by_task_composite_id.update((configs) => ({
+        ...configs,
+        [composite_key]: data || [],
+      }))
+    } catch (error) {
+      console.error(
+        "Failed to load task run configs:",
+        createKilnError(error).getMessage(),
```
scosman: createKilnError unnecessary for logging.

app/web_ui/src/lib/stores/run_configs_store.ts
```diff
@@ -0,0 +1,136 @@
...
+      )
+      // Set empty array for this task on error using composite key
+      run_configs_by_task_composite_id.update((configs) => ({
+        ...configs,
+        [composite_key]: [],
```
scosman: empty isn't correct. Empty is a valid value, but this was an error not empty set. By setting UI will be wrong, and it will block retrying.

Can make the store allow errors:
```
export const run_configs_by_task_composite_id = writable<
  Record<string, TaskRunConfig[] | KilnError>
>({})
```

Or add a new error store UI can check
```
export const run_configs_errors_by_task_composite_id = writable<
  Record<string, KilnError>
>({})
```

I prefer the latter. Your probably want a 3rd one for "loading" state as well

app/web_ui/src/lib/stores/run_configs_store.ts
```diff
@@ -0,0 +1,136 @@
...
+        run_config_properties,
+      },
+    },
+  )
+
+  if (error) {
+    throw error
+  }
+
+  // Reload the run configs to include the new one (force refresh to get fresh data)
```
scosman, line 154: maybe move this up above error? There's a chance it saved to disk then failed. Reload is safe bet.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte
```diff
@@ -14,10 +14,19 @@
...
+  import {
+    load_task_run_configs,
+    run_configs_by_task_composite_id,
+    get_task_composite_id,
+  } from "$lib/stores/run_configs_store"
   import { prompt_link } from "$lib/utils/link_builder"
 
   $: project_id = $page.params.project_id
   $: task_id = $page.params.task_id
+  $: current_task_run_configs =
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte:26]: default to null, not empty, so we can tell the difference between loaded and empty, still loading, and error.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte
```diff
@@ -70,40 +79,10 @@
...
-      run_configs_loading = false
-    }
-  }
-
   let toggle_eval_favourite_error: KilnError | null = null
 
-  $: loading = evals_loading || run_configs_loading
-  $: error = evals_error || run_configs_error || toggle_eval_favourite_error
+  $: loading = evals_loading
+  $: error = evals_error || toggle_eval_favourite_error
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte]: bring back the error here. By removing we'll act like it's empty, when it's actually an error.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte
```diff
@@ -70,40 +79,10 @@
...
-    } finally {
-      run_configs_loading = false
-    }
-  }
-
   let toggle_eval_favourite_error: KilnError | null = null
 
-  $: loading = evals_loading || run_configs_loading
-  $: error = evals_error || run_configs_error || toggle_eval_favourite_error
+  $: loading = evals_loading
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/+page.svelte]: bring back loading check for run configs here. Will show the UI before it's loaded without it.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/+page.svelte
```diff
@@ -51,19 +56,20 @@
...
-  let task_run_configs_loading = true
-
   let score_summary: EvalResultSummary | null = null
   let score_summary_error: KilnError | null = null
 
   // Note: not including score_summary_error, because it's not a critical error we should block the UI for
-  $: loading = eval_loading || eval_configs_loading || task_run_configs_loading
-  $: error = eval_error || eval_configs_error || task_run_configs_error
+  $: loading = eval_loading || eval_configs_loading
+  $: error = eval_error || eval_configs_error
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/+page.svelte]: 2x ditto

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/+page.svelte
```diff
@@ -51,19 +56,20 @@
...
+
+  $: current_task_run_configs =
+    $run_configs_by_task_composite_id[
+      get_task_composite_id(project_id, task_id)
+    ] || []
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/+page.svelte]: ditto

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/add_run_method.svelte
```diff
@@ -15,7 +15,7 @@
   import AvailableModelsDropdown from "../../../../../run/available_models_dropdown.svelte"
   import PromptTypeSelector from "../../../../../run/prompt_type_selector.svelte"
   import Collapse from "$lib/ui/collapse.svelte"
-  import RunOptions from "$lib/ui/run_options.svelte"
+  import AdvancedRunOptions from "../../../../../run/advanced_run_options.svelte"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/add_run_method.svelte]: Rename is good, but it should be back in the prior folder `$lib/ui/` since it really is a shared UI component. That eliminates these ugly relative imports.

Ones above aren't great either but legacy. We shouldn't be moving shared components from UI to routes

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -44,9 +49,7 @@
   let selectedModels: (string | null)[] = [null, null] // Track selected model for each column
 
   // Run configs state
-  let task_run_configs: TaskRunConfig[] | null = null
   let loading = true
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:52]: bring back error handling!

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -293,8 +271,13 @@
   }
 
   // Generate dropdown options from run configs
+  $: current_task_run_configs =
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:286]: don't default to empty. Checks for !current_task_run_configs above are broken by setting a default value

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:52]: Loading is likely broken? You set it to true, and deleted the code which set it to false. Not sure how this ever gets past the spinner now.

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -524,15 +509,6 @@
     <div class="w-full min-h-[50vh] flex justify-center items-center">
       <div class="loading loading-spinner loading-lg"></div>
     </div>
-  {:else if error}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:527]: bring back error case

app/web_ui/src/routes/(app)/run/prompt_type_selector.svelte
```diff
@@ -86,7 +86,10 @@
       if (prompt.chain_of_thought_instructions && exclude_cot) {
         continue
       }
-      static_prompts.push({ value: prompt.id, label: prompt.name })
+      static_prompts.push({
+        value: prompt.id,
+        label: `${prompt.name} (${prompt.generator_id ? prompt_name_from_id(prompt.generator_id, current_task_prompts) : "Custom"})`,
```
scosman: Feels like we'll end up with prompt.name twice? assuming prompt_name_from_id prints the name?

Might be worth new method adjacent to prompt_name_from_id... detailed_prompt_name_from_id?

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+  } from "$lib/stores"
+  import { getRunConfigPromptDisplayName } from "$lib/utils/run_config_formatters"
+  import { onMount } from "svelte"
+  import {
+    load_task_run_configs,
+    run_configs_by_task_composite_id,
+    get_task_composite_id,
+  } from "$lib/stores/run_configs_store"
+
+  export let selected_run_config_id: string | "custom" | null
```
scosman: nit: `string | "custom"` is redundant.

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+  export let default_run_config_id: string | null
+
+  onMount(async () => {
+    await load_available_prompts()
+    await load_model_info()
+  })
+
+  $: if ($current_project?.id && $current_task?.id) {
```
scosman: don't use `current_project.id` or `current_task.id` in any new files. Make project_id and task_id required parameters (`export let project_id: string`), and pass in from the parent `+page.svelte`. We'll keep the mistake isolated there.

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+    load_task_run_configs($current_project.id, $current_task.id)
+  }
+
+  // Build the options for the dropdown
+  $: options = (() => {
```
scosman: nit (P3, but just sharing): I usually like to do it like the example below. Bit more explicit for type checking (return value not dynamic, will catch errors like new early returns), and you explicitly define which parameters you're reactive to. This is also fine, but just sharing.

```
$: options = build_options(reactive params...)
function build_options(reactive params...) -> OptionsGroup[] {
   ...
}
```

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+    const options: OptionGroup[] = []
+
+    // Add new custom configuration option
+    options.push({
+      label: "",
+      options: [
+        {
+          value: "custom",
+          label: "None",
+          description: "Run with your current manually selected options.",
```
scosman: maybe: 
 - "Custom" + descrioption="Run with the options specified below. This config is not saved."

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+  $: options = (() => {
+    const options: OptionGroup[] = []
+
+    // Add new custom configuration option
+    options.push({
+      label: "",
+      options: [
+        {
+          value: "custom",
+          label: "None",
```
scosman: I'm still leaning "Unsaved" or "Custom'

None feels weird. I don't want "None/no run configuration" I want a custom run configuration. There's still is a run configuration even when this is set to "None", so None feels strange.

"Unsaved" is kinda nice. Bit of a call to action (you should save if you want to keep this), plus contrasts the "Saved Configurations" section nicely. However, people are going to confuse that with the primary function (Run), which is also saving data (samples into dataset).

Still think "Custom" is the most descriptive. We can put the unsaved part in the description so it at it doesn't feel like the UI is in a bad state (which "Unsaved" kinda does).

app/web_ui/src/lib/stores/run_configs_store.ts
```diff
@@ -0,0 +1,136 @@
+import type { TaskRunConfig, RunConfigProperties } from "$lib/types"
+import { writable } from "svelte/store"
+import { client } from "$lib/api_client"
+import { createKilnError } from "$lib/utils/error_handlers"
+
+// Helper function to create composite keys for a task
+export function get_task_composite_id(
```
scosman, line 10: totally optional, but if you want to get fancy you can define a custom type here that forces people to use the right type of string.

```
type TaskCompositeId = string & { __brand: "TaskCompositeId" };

export function get_task_composite_id(
  project_id: string,
  task_id: string,
): TaskCompositeId {
  return `${project_id}:${task_id} as TaskCompositeId`
}

export const run_configs_by_task_composite_id = writable<
  Record<TaskCompositeId, TaskRunConfig[]>
>({})
```

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+
+    // Add saved configurations if they exist
+    let saved_configuration_options: Option[] = []
+
+    // Add default configuration first if it exists
+    if (default_run_config_id) {
+      const default_config = (
+        $run_configs_by_task_composite_id[
+          get_task_composite_id(
+            $current_project?.id ?? "",
```
scosman: this will be cleaner with change above making these properties required.

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+            $current_task?.id ?? "",
+          )
+        ] ?? ([] as TaskRunConfig[])
```
scosman: shouldn't need the `as` here. We try to never cast.

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+      ).find((config) => config.id === default_run_config_id)
+
+      if (default_config) {
+        saved_configuration_options.push({
+          value: default_run_config_id,
+          label: `${default_config.name} (Default)`,
+          description: `Model: ${model_name(default_config.run_config_properties.model_name, $model_info)} (${provider_name_from_id(default_config.run_config_properties.model_provider_name)})
```
scosman: make a helper for this, since it's repeated

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
```diff
@@ -0,0 +1,112 @@
...
+      })
+    }
+
+    return options
+  })()
+</script>
+
+<FormElement
+  label="Run Configuration"
+  info_description="Select a saved run configuration to automatically apply its settings, or choose None to manually configure run options and save them for future use."
```
scosman: Yeah, None feels weird. I don't want "None/no run configuration" I want a custom run configuration. There's still a run configuration even when this is set to "None", so None feels strange.

String a bit confusing. Selecting None doesn't save them for future use.

Maybe make this dynamic:
 - if saved options > 1: "Select a saved run configuration which includes model, prompt, tools and properties. Alternatively choose 'Custom' to manually configure this run."
 - if saved_options.len == 0: "You can save your run configuration including model, prompt, tools and properties. This makes it easy to return to later."

app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -6,7 +6,15 @@
     ui_state,
     available_models,
     available_model_details,
+    load_available_prompts,
+    load_current_task,
   } from "$lib/stores"
+  import {
```
scosman: Can we add all the new logic in Run to a new component run_config.svelte (in lib/ui). run is already way too big, and we'll need this logic pulled out anyways.

In a separate PR we can update the evals screen to use the new component which would let them manage configs there too?

app/web_ui/src/routes/(app)/run/run_options_dropdown.svelte
sfierro: thinking just Run with the options specified below. since longer descriptions there look overwhelming and we say you can save it in the info description (tooltip)

app/web_ui/src/routes/(app)/run/prompt_type_selector.svelte
sfierro: it is two different names, one is the auto generated prompt name (Chilly Falcon) and one is the generator name (Zero Shot)

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:527]: oops!

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/add_run_method.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_methods/add_run_method.svelte]: oops not sure how that happened

app/web_ui/src/lib/stores/run_configs_store.ts
sfierro: love this!

app/web_ui/src/lib/utils/form_element.svelte
```diff
@@ -111,6 +111,7 @@
         <span class="pl-1 text-xs text-gray-500 flex-none"
           >{info_msg || (optional ? "Optional" : "")}</span
         >
+        <slot name="inline-actions"></slot>
```
scosman: For design consistency, I'd give them a typed input, not a slot. Giving them a slot let's them mess up the design. If we want to change the design for the app, we'd need to go update all instances. 

We use slots pretty sparingly for that reason (mostly in layout files),

Example:
```
export type InlineAction = {
  handler: () => void
  label: string
}

export let inlineAction: InlineAction | null = null
```

The put the presentation code here instead of a slot. Callers can specify text/action, but can't mess up the style.
```
{#if inlineAction}
  <button class="link font-medium text-grey-500" on:click{inlineAction.handler}>{inlineAction.label}</button>
{/if>
```

app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -189,47 +170,35 @@
...
-              bind:top_p
-              bind:structured_output_mode
-              has_structured_output={requires_structured_output}
-              project_id={$current_project?.id}
-              task_id={$current_task?.id || ""}
-            />
-          </Collapse>
-        {/if}
-      </div>
+      <RunConfigComponent
```
scosman: 🎉

app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -175,6 +140,22 @@
     run_complete = false
     clear_all()
   }
+
+  // Helper function to convert run options to server run_config_properties format
```
scosman: You could move this into RunConfigComponent. I think you can stop exporting the N small properties, and just expose this function. Will remove a lot of code in page.

Either as `export function...` adding this function, or refactor it to use a RunConfigProperties internally for state and just expose that (both work).

app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -189,47 +170,35 @@
...
+        bind:prompt_method
+        {requires_structured_output}
+        {requires_tool_support}
+        onModelChange={(newModel) => (model = newModel)}
+        onTemperatureChange={(newTemp) => (temperature = newTemp)}
+        onTopPChange={(newTopP) => (top_p = newTopP)}
+        onStructuredOutputModeChange={(newMode) =>
+          (structured_output_mode = newMode)}
+        onToolsChange={(newTools) => (tools = newTools)}
+        onPromptMethodChange={(newMethod) => (prompt_method = newMethod)}
```
scosman: None of these should be necessary, "bind" should be doing this for you.

scosman: See comment "You could move this into RunConfigComponent. " above which I believe will fix this for you.

But separately it's good to know this pattern shouldn't be needed. "bind" alone should propagate the properties up from child to parent.

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+  export let temperature: number = 1.0
+  export let top_p: number = 1.0
+  export let structured_output_mode: StructuredOutputMode = "default"
+  export let tools: string[] = []
+  export let prompt_method = "simple_prompt_builder"
+  export let requires_structured_output: boolean = false
+  export let requires_tool_support: boolean = false
+
+  // Events
+  export let onModelChange: (model: string) => void
```
scosman: See other comments. These aren't needed.

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+    AvailableModels,
+  } from "$lib/types"
+  import AvailableModelsDropdown from "./available_models_dropdown.svelte"
+  import AdvancedRunOptions from "$lib/ui/advanced_run_options.svelte"
+  import Collapse from "$lib/ui/collapse.svelte"
+  import { tick, onMount } from "svelte"
+  import SavedRunConfigurationsDropdown from "./saved_run_configs_dropdown.svelte"
+
+  // Props
+  export let project_id: string
```
scosman, line 26: See other comment about maybe using a RunConfigProperties here (as two options). Either way you can not export all these 1:1 and export as a RunConfigProperties (function or propetly)

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+  import Collapse from "$lib/ui/collapse.svelte"
+  import { tick, onMount } from "svelte"
+  import SavedRunConfigurationsDropdown from "./saved_run_configs_dropdown.svelte"
+
+  // Props
+  export let project_id: string
+  export let task_id: string
+  export let default_run_config_id: string | null = null
+  export let model: string = ""
+  export let temperature: number = 1.0
```
scosman: Worth bringing in the comment about where these magic values come from into here (I think it'll be deleted from +page.svelte)

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+    // Load prompts for the specified task
+    if (task_id && project_id) {
+      await load_available_prompts()
+    }
+  })
+
+  let save_config_error: KilnError | null = null
+  let set_default_error: KilnError | null = null
+
+  let selected_run_config_id: string | null = "custom"
```
scosman: Do we want to allow null here? Feels like we should always have a value.

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+
+  // RunOptionsDropdown Logic
+
+  // Update form values when selected_run_config changes
+  $: if (selected_run_config !== "custom") {
+    update_current_run_options()
+  }
+
+  async function update_current_run_options() {
+    updating_current_run_options = true
```
scosman: this probably doesn't do anything. JS is single threaded, so without an "await" somewhere in here, this won't ever be interrupted and get the "true" value.

nit: for flags I like to use this pattern for error handling:
```
try {
   updating_current_run_options = true
   ...
} finally {
   updating_current_run_options = false
}   
```

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+      config.run_config_properties.model_provider_name +
+      "/" +
+      config.run_config_properties.model_name
+    temperature = config.run_config_properties.temperature
+    top_p = config.run_config_properties.top_p
+    structured_output_mode = config.run_config_properties.structured_output_mode
+    tools = [...(config.run_config_properties.tools_config?.tools ?? [])]
+
+    // Notify parent of changes
+    onPromptMethodChange(prompt_method)
```
scosman: Should remove these (see other comments)

app/web_ui/src/routes/(app)/run/+page.svelte
scosman: Update, you already have it! Just add an export to it.

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+  // Update form values when selected_run_config changes
+  $: if (selected_run_config !== "custom") {
+    update_current_run_options()
+  }
+
+  async function update_current_run_options() {
+    updating_current_run_options = true
+
+    // Populate values from saved configuration
+    const config = selected_run_config as TaskRunConfig
```
scosman: No casting allowed. Something like this will fix your type checking.

```
if (selected_run_config === "custom) {
  return
}
```

The `selected_run_config: TaskRunConfig | "custom"` is a bit wonky. Kinda duplicates selected_run_config_id and I'm worried about keeping them in sync. Maybe make selected_run_config a function, which uses selected_run_config_id and returns the TaskRun when called?

app/web_ui/src/lib/ui/run_config_component.svelte
```diff
@@ -0,0 +1,368 @@
...
+      if (
+        config_properties.model_name !== current_model_name ||
+        config_properties.model_provider_name !== current_provider_name ||
+        config_properties.prompt_id !== prompt_method ||
+        config_properties.temperature !== temperature ||
+        config_properties.top_p !== top_p ||
+        config_properties.structured_output_mode !== structured_output_mode ||
+        !arrays_equal(config_properties.tools_config?.tools ?? [], tools)
+      ) {
+        selected_run_config_id = "custom"
```
scosman: should you also be setting selected_run_config = custom?

sfierro: it'll change with the reactive statement

app/web_ui/src/lib/ui/run_config_component.svelte
sfierro: yeah let me look into not having the duplication aspect

scosman (approved):

PR #605, opened 2025-09-16
feat: RagConfig filters docs by tags

leonardmq:
## What does this PR do?

Allow selecting one or multiple tags when creating a RagConfig to restrict the set of documents that the RagConfig sees. When multiple tags are selected, the matching set of documents is every document where at least one tag is in the union of selected tags (e.g. select `car` and `truck`, every document with either one will be matched).

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Tag-based filtering for RAG configs (null/empty = all documents).
  * Tag Selector when creating RAG configs; selected tags persisted and surfaced in config lists/details.
  * Workflow runs respect config tags and progress outputs now include run logs.
  * New project-level endpoint to list document tags.

* **Validation**
  * Tags validated (no empty tags or tags containing spaces).

* **UI**
  * Tag badges in config rows and details; configurable empty-label for multi-select.

* **Tests**
  * Extensive tests covering tags, progress, runners, and deduplication.

### DIFF
--- a/libs/core/kiln_ai/datamodel/rag.py
+++ b/libs/core/kiln_ai/datamodel/rag.py
@@ -1,6 +1,6 @@
 from typing import TYPE_CHECKING, Union
 
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from kiln_ai.datamodel.basemodel import ID_TYPE, FilenameString, KilnParentedModel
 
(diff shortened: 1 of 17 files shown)


libs/core/kiln_ai/datamodel/rag.py
```diff
@@ -34,8 +34,26 @@ class RagConfig(KilnParentedModel):
...
+
+    @model_validator(mode="after")
+    def validate_tags(self):
+        if self.tags is not None:
+            if len(self.tags) == 0:
+                raise ValueError("Tags cannot be an empty list.")
+            for tag in self.tags:
+                if not tag:
+                    raise ValueError("Tags cannot be empty.")
+                if " " in tag:
```
scosman, line 56: not sure this is true! Spaces work fine in dataset.

scosman (commented):
Made some UI changes -- make sure you like them before merging. But looks good!

libs/core/kiln_ai/datamodel/rag.py
leonardmq: Spaces are not allowed in the dataset as far as I can tell: 
https://github.com/Kiln-AI/Kiln/blob/5671c795ca260db0cd8fc5aad9b47f5130539685/libs/core/kiln_ai/datamodel/task_run.py#L256

Just went with the same scheme as we had elsewhere for uniformity - but we could probably allow spaces if we want to

PR #606, opened 2025-09-16
refactor: increase lock acquisition timeout

leonardmq:
## What does this PR do?

Increase RAG step lock acquisition timeout to 1 hour (up from 1 minute). This will allow for more lenient handling of RAG Config runs triggered simultaneously when their subconfigs are shared.

For example, if you have 3 RagConfig that have the same extractor; the following sequence is likely to happen:
1. You upload 10 documents.
2. Frontend triggers all the RagConfigs that start concurrently.
3. The 3 RagConfigs try to start, each one tries to get the same lock for the `Extraction` substep (because the same extractor config is used for all 3).
4. One RagConfig acquires the lock while the other two wait.
5. Timeout duration elapses; but the lock-holding RagConfig is still not done extracting. The two configs waiting on the lock give up -> `Completed with Errors` -> user needs to run the configs manually afterwards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased internal lock timeout for AI processing workflows to 1 hour to prevent premature timeouts on long-running or large jobs.
  * Reduces intermittent lock/timeout errors and improves reliability across extraction, chunking, embedding, and indexing steps.
  * No configuration changes required; existing workflows benefit automatically.
  * Improves overall stability without altering output or user-facing behavior.

### DIFF
--- a/libs/core/kiln_ai/adapters/rag/rag_runners.py
+++ b/libs/core/kiln_ai/adapters/rag/rag_runners.py
@@ -39,6 +39,12 @@ from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 from kiln_ai.utils.lock import shared_async_lock_manager
 from pydantic import BaseModel, ConfigDict, Field
 
+# We set the timeout high because current UX is likely to cause the user triggering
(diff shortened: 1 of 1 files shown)


leonardmq:
@coderabbitai review

PR #609, opened 2025-09-16
Rag UI Updates

scosman:
- Functional "Easy Mode" for creating search tools
 - Fully integrate into the existing tools UI. RAG is listed in /settings/tools, can create RAG from setting, etc.
 - Rename "Vector store" to "Search Index" since it might just be full text search, and better for noobs
 - Lots of string and design tweaks.

Note: will re-point at rag branch after rag-tool is merged.

https://github.com/user-attachments/assets/222148d6-3940-4e12-b66a-c643933b4768

## Summary by CodeRabbit

* **New Features**
  * Project-level Search Tools API and an "Add Search Tool" guided workflow with templates and an Add Search Tool page.

* **UI**
  * Feature carousel; Manage Tools surfaces RAG search tools; unified clickable settings items; many label/text updates (e.g., "Vector Store" → "Search Index") and spacing/typography tweaks.

* **Improvements**
  * Per-tool name/description for search tools, friendlier embedding model names, extractor MIME-type hints, improved RAG progress completion status, provider-required badges, prompt generators, and tool-name validation.

* **Tests**
  * Expanded coverage for search-tools API, RAG labeling, embedding names, tool-name validation, and streaming/error flows.


app/web_ui/src/lib/utils/input_validators.ts
```diff
@@ -103,5 +104,10 @@ export const tool_name_validator: (value: unknown) => string | null = (
     return "Must start with a lowercase letter"
   }
 
+  // Check length
+  if (name.length > 64) {
+    return "Must be less than 64 characters long"
```
leonardmq, line 109: Message says less than 64, but if checks if more than 64. 

Is exactly 64 allowed?

app/desktop/studio_server/tool_api.py
```diff
@@ -242,6 +242,23 @@ async def get_available_tools(
...
 
+        # Add search tools (RAG)
+        rag_configs = project.rag_configs(readonly=True)
+        if rag_configs:
+            tool_sets.append(
+                ToolSetApiDescription(
+                    set_name="Search Tools (RAG)",
+                    tools=[
+                        ToolApiDescription(
+                            id=f"{RAG_TOOL_ID_PREFIX}{rag_config.id}",
```
leonardmq, line 253: Should we update the `@app.get("/api/projects/{project_id}/search_tools")` endpoint to return these IDs too? 

Right now we return `id=rag_config.id` over there; maybe confusing later if no prefix there

app/web_ui/src/lib/utils/input_validators.ts
scosman: haha, good catch

app/desktop/studio_server/tool_api.py
scosman: I think it's fine? One is the real ID, the other is tool ID. Both are correct in the context.

PR #612, opened 2025-09-17
fix: reconcile vector store by deleting docs no longer targeted

leonardmq:
## What does this PR do?

Problem:
- deleting documents or untagging them take them out of the set of documents targeted by a RAG Config
- if the documents being removed from the set were already indexed in the db, they will stay there

This PR adds a reconciliation step after the indexing is complete to iterate over all the nodes in the store, and remove them if they are not in the set of known documents.

## Checklists

## Summary by CodeRabbit

- New Features
  - Automatic vector index cleanup after indexing, removing entries for documents that were deleted or no longer match tags.
  - Tag-aware maintenance: only keeps vectors for documents matching configured tags; clearer feedback when no documents match.
  - Improved consistency between project content and the vector index, reducing stale data and storage usage.

- Tests
  - Extensive tests added for tag-filtered indexing and cleanup, including empty/complete sets and uninitialized/empty store scenarios.

### DIFF
--- a/app/web_ui/src/lib/api_schema.d.ts
+++ b/app/web_ui/src/lib/api_schema.d.ts
@@ -403,6 +403,23 @@ export interface paths {
         patch?: never;
         trace?: never;
     };
+    "/api/projects/{project_id}/documents/tags": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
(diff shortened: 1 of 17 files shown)


leonardmq:
Cherry-picked and rebased, but it closed the PR. Opened a new PR for the new branch here: https://github.com/Kiln-AI/Kiln/pull/613

PR #616, opened 2025-09-18
feat: trigger all rag configs after tagging or doc deletion

leonardmq:
## What does this PR do?

Tagging or deleting a document may take it out of or add it into a RagConfig - e.g. a doc is tagged with the tag `truck`, and a RagConfig filters for this tag, now it needs to process and index it; similarly if such a doc is untagged or deleted, the index must be updated to remove the doc chunks.

Solution:
- after tagging / deletion, we trigger a client-side flow that retrieves all the rag configs and triggers all of them to run; this allows for UI updates the normal way
- browsers limit the number of connections that can be open in parallel, so this PR also adds a semaphore to allow queueing those requests (and avoid kicking off as many connections as the user has RagConfigs in the project + whatever replay they may have triggered by queued operations)

Some notes:
- not a problem if the user does rapid fire tagging / deletion - we call the client method to trigger all rag configs every time but:
  -  they get queued up on the frontend so eventually all of them will run (and loading state will be the state of whichever one was triggered, which is fine); 
  - they also get queued on the backend via a server side lock that timeouts acquisition attempts only after 1 hour, so all of them will eventually run or time out
- we indiscriminately trigger all rag configs to run even those that are not impacted by the operation (because the doc is and was not in the set the config targets), but that is fine since those will be very fast to complete because they have nothing to do; so their work is only marginally heavier than checking current progress

## Checklists

### DIFF
--- a/app/web_ui/src/lib/stores/rag_progress_store.ts
+++ b/app/web_ui/src/lib/stores/rag_progress_store.ts
@@ -3,6 +3,7 @@ import { base_url, client } from "$lib/api_client"
 import type { LogMessage, RagConfigWithSubConfigs, RagProgress } from "../types"
 import { createKilnError, type KilnError } from "../utils/error_handlers"
 import { progress_ui_state } from "./progress_ui_store"
+import { Semaphore } from "$lib/utils/semaphore"
 
 export type RagConfigurationStatus =
   | "not_started"
@@ -26,6 +27,10 @@ export const ragProgressStore = createRagProgressStore()
 export const allRagConfigs = derived(ragProgressStore, ($store) => {
   return sortRagConfigs(Object.values($store.rag_configs), "created_at")
 })
+
+// cap concurrent EventSource connections browsers have different limits,
+// we keep the limit low enough to be safe for most browsers
+const sseSlots = new Semaphore(5)
 
 function createRagProgressStore() {
   const { subscribe, set, update } = writable<RagConfigurationProgressState>({
@@ -63,7 +68,7 @@ function createRagProgressStore() {
   }
 
   async function run_all_rag_configs(project_id: string): Promise<boolean> {
-    // retrieve all rag configs
+    // get all rag configs
     const { data, error } = await client.GET(
       "/api/projects/{project_id}/rag_configs",
       {
@@ -79,17 +84,34 @@ function createRagProgressStore() {
       return false
     }
 
+    // process rag configs with concurrency limit to avoid hitting browser limits
+    const tasks: Promise<unknown>[] = []
     for (const rag_config of data) {
       if (!rag_config.id) {
         continue
       }
-      run_rag_config(project_id, rag_config.id)
+
+      tasks.push(
+        (async () => {
+          await sseSlots.acquire()
+          try {
+            await run_rag_config(project_id, String(rag_config.id))
+          } finally {
+            sseSlots.release()
+          }
+        })(),
+      )
     }
 
-    return true
+    const results = await Promise.allSettled(tasks)
+    const failed = results.filter((r) => r.status === "rejected")
+    return failed.length === 0
   }
 
-  function run_rag_config(project_id: string, rag_config_id: string): boolean {
+  function run_rag_config(
+    project_id: string,
+    rag_config_id: string,
+  ): Promise<void> {
     update((state) => ({
       ...state,
       status: { ...state.status, [rag_config_id]: "running" },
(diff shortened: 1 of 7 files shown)


app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -79,17 +86,36 @@ function createRagProgressStore() {
...
+        (async () => {
+          await sseSlots.acquire()
+          try {
+            await run_rag_config(project_id, String(rag_config.id))
+          } finally {
+            sseSlots.release()
+          }
+        })(),
+      )
     }
```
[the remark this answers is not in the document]
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/616/commits/b9e3614d3092b0a36199c22a6376540f32e4dfc1

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -79,17 +86,36 @@ function createRagProgressStore() {
...
 
-    return true
+    const results = await Promise.allSettled(tasks)
+    const failed = results.filter((r) => r.status === "rejected")
+    return failed.length === 0
   }
```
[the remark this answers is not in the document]
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/616/commits/b9e3614d3092b0a36199c22a6376540f32e4dfc1

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -115,85 +129,149 @@ function createRagProgressStore() {
...
+              (extraction_incomplete ||
+                chunking_incomplete ||
+                embedding_incomplete ||
+                indexing_incomplete)
+
+            const completion_status =
+              has_errors(rag_config_id) || failed_implicitly
+                ? "completed_with_errors"
+                : "complete"
+
```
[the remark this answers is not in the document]
leonardmq: No functional impact because at the moment, they are the same thing. But should be fixed - I fixed it in a separate PR (that has not been merged yet) already in response to another coderabbitai comment pointing it out over there

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -123,85 +133,149 @@ function createRagProgressStore() {
       current_step: 0,
     })
 
+    // we will send a log to say that the config is queued if we are still blocked for more than 10s
```
scosman, line 136: for later, but we should do something similar for page progress. They can feel stuck when processing lots of pages.

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -123,85 +133,149 @@ function createRagProgressStore() {
...
+        showed_queued_log = true
+        update((state) => ({
+          ...state,
+          logs: {
+            ...state.logs,
+            [rag_config_id]: [
+              ...(state.logs[rag_config_id] ?? []),
+              {
+                level: "info",
+                message: `A maximum of ${MAX_CONCURRENT_RAG_CONFIGS} configs can run at the same time. This config is queued and will start automatically when a slot is free. Refreshing the browser window will cancel the queue and any config that is currently running or waiting to start.`,
```
scosman: Bit wordy. We're explaining the internals but the user message can be more focused on the state and needed action (just wait):

"Queued. Will start automatically once other runs finish."

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -123,85 +133,149 @@ function createRagProgressStore() {
...
+            { level: "info", message: "Started running" },
+          ],
+        },
+      }))
+    }
+
+    try {
+      await run_rag_config_unsafe(project_id, rag_config_id)
+    } finally {
+      sseSlots.release()
```
scosman: Counting semaphores are dangerous. At minimum let's extend the try block around the acquire so there's zero chance of introducing a crash which leaves the semaphore unreleased. That still has the issue of breaking the count if acquire fails.

Generally I prefer block based style where you can't forget to release. It releases as soon as you leave the block

```
let lock = ConcurrentRunner(concurrency=2) 
for (a in b) { 
  lock.run(() => {do_something(a)} 
}
```

ChatGPT's version (not reviewed):

```
function createLimiter(max) {
  let active = 0;
  const queue = [];

  function next() {
    if (queue.length === 0 || active >= max) return;
    active++;
    const { fn, resolve } = queue.shift();
    fn().finally(() => {
      active--;
      next();
    }).then(resolve);
  }

  return function run(fn) {
    return new Promise(resolve => {
      queue.push({ fn, resolve });
      next();
    });
  };
}

// Usage:
const run = createLimiter(2);
for (const a of b) {
  run(() => do_something(a));
}

```

scosman (changes requested):

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -79,25 +85,29 @@ function createRagProgressStore() {
...
-  function run_rag_config(project_id: string, rag_config_id: string): boolean {
+  async function run_rag_config(
+    project_id: string,
+    rag_config_id: string,
+  ): Promise<void> {
+    // we update the status even if waiting for slot, so that it looks
+    // like it is running to the user
     update((state) => ({
       ...state,
       status: { ...state.status, [rag_config_id]: "running" },
```
leonardmq, line 113: Should also update state of substeps here

app/web_ui/src/lib/stores/rag_progress_store.ts
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/616/commits/ae29cef5291916cf1ab258ab9734fdba179ca267

Refactored to do mostly the ChatGPT version - works fine

app/web_ui/src/lib/stores/rag_progress_store.ts
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/616/commits/ae29cef5291916cf1ab258ab9734fdba179ca267

PR #617, opened 2025-09-18
feat: add archiving / unarchiving for RagConfig

leonardmq:
## What does this PR do?

Issue:
- users have no way of deleting or disabling RAG Configs; they are very likely to click around just to see what it does, and should have a way of disabling the configs they no longer want to avoid wasted requests / time
- SSE appears to monopolize the browser after only 6 open requests (every request to the server hangs after that); so we cannot afford to run unnecessary configs from this perspective too

This PR adds a way of archiving (and unarchiving) existing configs.

Add `Archive` / Unarchive` button on RagConfig detailed page (with warning when it is archived); with following side effects:
1. Archived configs are skipped when we trigger all rag configs
2. Archived configs are filtered out in `@app.get("/api/projects/{project_id}/available_tools")`
3. Archived configs are filtered out in `@app.get("/api/projects/{project_id}/search_tools")`
4. Archived configs return an error if we try to run them
5. Table UI -> hide number of documents, hide action button (e.g. `Run`, etc.) and override status (complete / incomplete) with just `Archived` and grey out the row

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Archive/unarchive Search Tool configurations; archived items are hidden from active listings and prevented from running.

* **UI**
  * Archive/Unarchive toggle, warning banner for archived items, "Test Search Tool" section, disabled inputs for archived configs, and an "Archived" badge; archived items shown after active ones.

* **API & State**
  * Archived flag in responses, partial updates supported, client tracks archived state and excludes archived configs from bulk runs.

* **Tests**
  * Added coverage for archived behavior across API and UI flows.

### DIFF
--- a/libs/server/kiln_server/document_api.py
+++ b/libs/server/kiln_server/document_api.py
@@ -1494,6 +1494,12 @@ def connect_document_api(app: FastAPI):
                 detail="RAG config not found",
             )
 
+        if rag_config.is_archived:
+            raise HTTPException(
+                status_code=422,
+                detail="This RAG configuration is archived. You must unarchive it to use it.",
+            )
+
         # Empty query - return empty results
         if not request.query.strip():
             return RagSearchResponse(results=[])
--- a/libs/server/kiln_server/test_document_api.py
+++ b/libs/server/kiln_server/test_document_api.py
@@ -2535,6 +2535,24 @@ async def test_search_rag_config_not_found(client, mock_project):
 
 
 @pytest.mark.asyncio
+async def test_search_rag_config_archived(client, mock_project, mock_rag_config):
+    """Test search with archived RAG config"""
+    with patch("kiln_server.document_api.project_from_id") as mock_project_from_id:
+        mock_project_from_id.return_value = mock_project
+
+        mock_rag_config.is_archived = True
+        mock_rag_config.save_to_file()
+
+        response = client.post(
+            f"/api/projects/{mock_project.id}/rag_configs/{mock_rag_config.id}/search",
+            json={"query": "test query"},
+        )
+
+    assert response.status_code == 422, response.text
+    assert "archived" in response.json()["message"]
+
+
+@pytest.mark.asyncio
 async def test_search_rag_config_vector_store_not_found(
     client, mock_project, mock_rag_config
 ):


app/desktop/studio_server/tool_api.py
```diff
@@ -530,4 +531,5 @@ async def get_search_tools(project_id: str) -> list[SearchToolApiDescription]:
                 description=rag_config.tool_description,
             )
             for rag_config in project.rag_configs(readonly=True)
+            if not rag_config.is_archived
```
leonardmq, line 534: Wasn't sure if we should filter them out here - it seems to hide them from what is being counted in the Settings > Tools page / table, which I suppose makes sense since an archived config should behave as if it did not exist mostly

@scosman

app/desktop/studio_server/tool_api.py
```diff
@@ -255,6 +255,7 @@ async def get_available_tools(
                             description=f"{rag_config.name}: {rag_config.tool_description}",
                         )
                         for rag_config in rag_configs
+                        if not rag_config.is_archived
```
leonardmq, line 258: Wondering if this is risky for the client? I saw there is a tool store on the frontend that uses index db; it does not seem to break, but I could be missing some downstream side effect impact

@scosman

scosman: Anything server side should be fine, can't impact the indexDB

app/desktop/studio_server/tool_api.py
scosman: Agreed.

libs/core/kiln_ai/datamodel/rag.py
```diff
@@ -14,6 +14,11 @@ class RagConfig(KilnParentedModel):
         description="A name to identify this RAG configuration for your own reference.",
     )
 
+    is_archived: bool = Field(
```
scosman, line 17: The frontend seems to respect this, but shouldn't the backend error somewhere if we try to process or run an archived config?

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
```diff
@@ -32,7 +32,17 @@
 
   $: status = $ragProgressStore.status[rag_config.id || ""]
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte:33]: Not relevant to this line: but sort the configs in the table view so archived are last.

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
```diff
@@ -75,7 +85,11 @@
 
 {#if rag_progress && rag_config}
   <tr
-    class="{row_hovered ? 'hover' : ''} cursor-pointer hover:bg-base-200"
+    class="{row_hovered
+      ? 'hover'
+      : ''} cursor-pointer hover:bg-base-200 {rag_config.is_archived
+      ? 'opacity-60 bg-gray-50'
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte]: Remove this. bg-gray-50 isn in our approved pallet and is too close to the hover:bg-base-200. Opacity messes with our color pallet and legibility. The badge is fine on it's own.

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
```diff
@@ -142,7 +156,11 @@
                 ? 'badge-success'
                 : ''} {status_badge_props?.error
                 ? 'badge-error'
-                : ''} {status_badge_props?.primary ? 'badge-primary' : ''}"
+                : ''} {status_badge_props?.primary
+                ? 'badge-primary'
+                : ''} {status_badge_props?.archived
+                ? 'text-gray-500 border-gray-300'
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte]: Just badge-secondary (and remove badge-outline, put it in the others but not this one)

Nit: always use our approved colors/sytles for consistency. Never hard coded custom values (gray-500 is approved for text, all others are named)

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/[rag_config_id]/rag_config/+page.svelte
```diff
@@ -83,6 +85,56 @@
     }
   }
 
+  export async function archive_rag_config() {
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/[rag_config_id]/rag_config/+page.svelte]: This block can be cleaned up:

 - duplicate code, just make on function and take a bool param or name it "toggle_rag_archived_status"
 - throws an error inconsistently (KilnError in one, not the other)
 - No UI for errors. Throwing it alone won't do anything. Since you're defining "const" I'm pretty sure you aren't setting the global error var. Do `const { error: patch_error } = ...` and `error = patch_error` to set the UI visible error (confirm).

libs/core/kiln_ai/datamodel/rag.py
leonardmq: We do check for archived status in `@app.get("/api/projects/{project_id}/rag_configs/{rag_config_id}/run")` and reject the run if it is archived: 
https://github.com/Kiln-AI/Kiln/blob/c4def1e462c6606b18bd293263e37722139e1e99/libs/server/kiln_server/document_api.py#L1441

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte:33]: Addressed in https://github.com/Kiln-AI/Kiln/pull/617/commits/72eda7b4b8b334e3934747bc986851f36f53f5ef

Not archived first, archived last; and within each of these two groups sorted by date created desc.

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte]: Addressed in https://github.com/Kiln-AI/Kiln/pull/617/commits/72eda7b4b8b334e3934747bc986851f36f53f5ef

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/table_rag_config_row.svelte]: Addressed in https://github.com/Kiln-AI/Kiln/pull/617/commits/72eda7b4b8b334e3934747bc986851f36f53f5ef

It does arguably look less archived now though (secondary badge is vantablack ultra)

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/[rag_config_id]/rag_config/+page.svelte
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/[rag_config_id]/rag_config/+page.svelte]: Addressed in https://github.com/Kiln-AI/Kiln/pull/617/commits/72eda7b4b8b334e3934747bc986851f36f53f5ef and https://github.com/Kiln-AI/Kiln/pull/617/commits/cc875e0ccd9239bc9d292ffa91faaf6fbc441757

Merged into one function; added proper error handling / visibility.

Set the error to be a separate one (the main one is full page hard error), that shows at the top like this:
image

image

libs/core/kiln_ai/datamodel/rag.py
leonardmq: Added rejection also for the `/search` endpoint in https://github.com/Kiln-AI/Kiln/pull/617/commits/03595909a8531a325689ac54b015583584bb3a5e

Also open question whether we should delete the local lancedb files when the user archives a config as they may take up space. Not necessary for V1.

PR #625, opened 2025-09-19
fix: deduplicate consecutive whitespace and newlines before chunking

leonardmq:
## What does this PR do?

Sometimes the model responds with many whitespace or newline chars (a lot of then - I had a case with 1.5M such consecutive characters). The chunker cannot handle it and crashes.

Fix:
- deduplicate consecutive `\n` and spaces in response before chunking
- increase temperature (lowering it seems to have made this worse)

## Checklists

## Summary by CodeRabbit

- New Features
  - Text is now automatically cleaned before chunking: consecutive newlines are limited to six and spaces to fifty to normalize input and improve chunk quality.
- Bug Fixes
  - Whitespace-only inputs now produce no chunks.
  - More robust handling of very large texts and extreme whitespace sequences.
- Tests
  - Added comprehensive tests for whitespace normalization, large-input scenarios, and pre-chunking cleaning.

### DIFF
--- a/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
+++ b/libs/core/kiln_ai/adapters/chunkers/base_chunker.py
@@ -3,6 +3,7 @@ from abc import ABC, abstractmethod
 
 from pydantic import BaseModel, Field
 
+from kiln_ai.adapters.chunkers.helpers import clean_up_text
 from kiln_ai.datamodel.chunk import ChunkerConfig
 
 logger = logging.getLogger(__name__)
@@ -30,7 +31,11 @@ class BaseChunker(ABC):
         if not text:
             return ChunkingResult(chunks=[])
 
-        return await self._chunk(text)
+        sanitized_text = clean_up_text(text)
+        if not sanitized_text:
+            return ChunkingResult(chunks=[])
+
+        return await self._chunk(sanitized_text)
 
     @abstractmethod
     async def _chunk(self, text: str) -> ChunkingResult:
--- a/libs/core/kiln_ai/adapters/chunkers/helpers.py
+++ b/libs/core/kiln_ai/adapters/chunkers/helpers.py
@@ -3,9 +3,21 @@ import re
 
 def clean_up_text(text: str) -> str:
     """
-    Clean up text by removing consecutive newlines and consecutive whitespace. Models sometimes send a lot of those.
+    Clean up text by limiting consecutive newlines and consecutive whitespace. Models sometimes send a lot of those.
     It seems to happen more when the transcription is done at low temperature.
+
+    - Replaces 6+ consecutive newlines with exactly 6 newlines
+    - Replaces 50+ consecutive spaces with exactly 50 spaces
+    - Leaves 1-5 consecutive newlines unchanged
+    - Leaves 1-49 consecutive spaces unchanged
     """
-    text = re.sub(r"\n+", "\n", text)
-    text = re.sub(" +", " ", text)
+    max_consecutive_newlines = 6
+    max_consecutive_whitespace = 50
+
+    # Replace 6+ consecutive newlines with exactly 6 newlines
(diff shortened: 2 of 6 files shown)


libs/core/kiln_ai/adapters/chunkers/base_chunker.py
```diff
@@ -30,7 +31,11 @@ async def chunk(self, text: str) -> ChunkingResult:
         if not text:
             return ChunkingResult(chunks=[])
 
-        return await self._chunk(text)
+        sanitized_text = clean_up_text(text)
+        if not sanitized_text:
+            return ChunkingResult(chunks=[])
+
+        return await self._chunk(sanitized_text)
```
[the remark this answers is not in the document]
leonardmq: Should be ok to remove those pre-chunking. We preserve the original text in the extraction, but chunks do not benefit much (arguably the opposite is true) from more than 1 space and more than 1 newline. MistralOCR for example even removes all newlines it seems

libs/core/kiln_ai/adapters/chunkers/helpers.py
```diff
@@ -0,0 +1,11 @@
+import re
+
+
+def clean_up_text(text: str) -> str:
+    """
+    Clean up text by removing consecutive newlines and consecutive whitespace. Models sometimes send a lot of those.
+    It seems to happen more when the transcription is done at low temperature.
+    """
+    text = re.sub(r"\n+", "\n", text)
```
scosman: Maybe `text = re.sub(r"\n{4,}", "\n\n", text)`

I feel like we should allow 1,2,3 newlines in a row, as there's a semantic difference between 1 and two newlines (but less so between 7 and 9).

libs/core/kiln_ai/adapters/chunkers/helpers.py
```diff
@@ -0,0 +1,11 @@
...
+    text = re.sub(" +", " ", text)
```
scosman: How big of an issue is this this? I feel like we're going to risk semantic changes. Example:

`ACCOUNT                       STATUS` has different meaning than `ACCOUNT STATUS`

Model bugs aren't great, but patching over them in code, which the user can't impact with the prompt, might be even worse. I vote to remove this, ideally fix with temp, and optionally add prompt guidance instead.

leonardmq: I think that very small semantic difference is offset by the negative of risking space-only chunks that are filled with spaces, or chunking breaking. 

One case I got 1 million+ spaces in a row. That was probably the temperature + weird PDF. 

Higher temperature should solve that though, but not bulletproof either. 

We could increase the threshold of number of consecutive occurrences to 4+ to dedupe

scosman: Yeah, it just feels like one of those bugs that will just keep popping up in other forms (repeated periods, repeated dashes, repeated tabs, loops, etc). Coding special cases will make things less predictable, and harder to fix at model/param level.

Let's do something like 6+ for newlines, and 50+ for spaces? Takes care of extreme cases with minimal change to model output? In both cases, collapse it to the max (20 newlines becomes 6 newlines, 1M spaces become 50, so semantically it's still "big gap" and doesn't accidentally turn the biggest gaps into the smallest)?

leonardmq: I'll cap the more egregious cases

Can we send events to analytics from the backend? 

Would be interesting to log how often the cleaning has an impact - maybe with higher temperature it never even happens (so far seems to be the case, but hard to be sure).

leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/625/commits/6470abb880b0f389f1de256f7de8e5470d137e81 - capped newlines at 6, and spaces at 50

scosman (approved):

PR #626, opened 2025-09-19
feat: filesystem cache for pdf pages

leonardmq:
## What does this PR do?

Extracting long PDFs is slow. The PDF splitting itself is fast, getting the response from the LLM is slow. Sometimes, it may happen that a PDF fails after extracting a number of its pages successfully - when that happens on the next run, we start processing it from the first page again, which is slow and expensive.

Fix:
- add filesystem cache; before extracting a PDF page, we query the cache; after each successful extraction we store the extraction as bytes in a file
- the filesystem cache is implemented as a singleton pointing to a temporary directory that is created on app start (if the user quits the app, the next time around, the dir will be different; but it survives refreshes etc.)
- key for each PDF page is `{extractor_config_id}_{hash of [pdf_path _ page number]}`

This allows resuming processing a PDF from where it last left off.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/utils/filesystem_cache.py
+++ b/libs/core/kiln_ai/utils/filesystem_cache.py
@@ -1,0 +1,57 @@
+import logging
+import tempfile
+from pathlib import Path
+
+from kiln_ai.datamodel.basemodel import name_validator
+
+logger = logging.getLogger(__name__)
+
+
+class FilesystemCache:
+    def __init__(self, path: Path):
+        # the key must be a valid filename
+        validate_key = name_validator(min_length=1, max_length=120)
+        self.validate_key = validate_key
+        self.cache_dir_path = path
+
+    def get_path(self, key: str) -> Path:
+        self.validate_key(key)
+        return self.cache_dir_path / key
+
+    def get(self, key: str) -> bytes | None:
+        # check if the file exists
+        if not self.get_path(key).exists():
+            return None
+
+        # we don't want to raise because of internal cache corruption issues
+        try:
+            with open(self.get_path(key), "rb") as f:
+                return f.read()
+        except Exception:
+            logger.error(f"Error reading file {self.get_path(key)}", exc_info=True)
+            return None
+
+    def set(self, key: str, value: bytes) -> Path:
+        logger.debug(f"Caching {key} at {self.get_path(key)}")
+        self.validate_key(key)
+        path = self.get_path(key)
+        path.write_bytes(value)
+        return path
+
+
+class TemporaryFilesystemCache:
+    _shared_instance = None
+
+    def __init__(self):
+        self._cache_temp_dir = tempfile.mkdtemp(prefix="kiln_cache_")
+        self.filesystem_cache = FilesystemCache(path=Path(self._cache_temp_dir))
+
+        logger.debug(
+            f"Created temporary filesystem cache directory: {self._cache_temp_dir}"
+        )
+
+    @classmethod
+    def shared(cls) -> FilesystemCache:
+        if cls._shared_instance is None:
+            cls._shared_instance = cls()
+        return cls._shared_instance.filesystem_cache
--- a/libs/core/kiln_ai/utils/test_filesystem_cache.py
+++ b/libs/core/kiln_ai/utils/test_filesystem_cache.py
@@ -1,0 +1,322 @@
+import tempfile
+from pathlib import Path
+from unittest.mock import mock_open, patch
+
+import pytest
+
+from kiln_ai.utils.filesystem_cache import FilesystemCache, TemporaryFilesystemCache
+
+
+class TestFilesystemCache:
+    @pytest.fixture
+    def temp_dir(self):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            yield Path(tmp_dir)
+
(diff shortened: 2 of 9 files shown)


libs/core/kiln_ai/utils/test_filesystem_cache.py
```diff
@@ -0,0 +1,322 @@
...
+
+        with pytest.raises(ValueError):
+            cache.set(invalid_key, b"content")
+
+
+class TestTemporaryFilesystemCache:
+    def test_temporary_cache_creation(self):
+        """Test that TemporaryFilesystemCache creates a temporary directory."""
+        # Reset the singleton for clean testing
+        TemporaryFilesystemCache._shared_instance = None
```
scosman: don't mutate _shared_instance_ or any globals in tests. Test should work fine without this.

libs/core/kiln_ai/utils/test_filesystem_cache.py
```diff
@@ -0,0 +1,322 @@
...
+        # Should have a FilesystemCache instance
+        assert isinstance(temp_cache.filesystem_cache, FilesystemCache)
+        assert temp_cache.filesystem_cache.cache_dir_path == Path(
+            temp_cache._cache_temp_dir
+        )
+
+    def test_multiple_instances_share_same_cache(self):
+        """Test that multiple calls to shared() return the same cache instance."""
+        # Reset the singleton for clean testing
+        TemporaryFilesystemCache._shared_instance = None
```
scosman: ditto

libs/core/kiln_ai/utils/filesystem_cache.py
```diff
@@ -0,0 +1,57 @@
...
+from kiln_ai.datamodel.basemodel import name_validator
+
+logger = logging.getLogger(__name__)
+
+
+class FilesystemCache:
+    def __init__(self, path: Path):
+        # the key must be a valid filename
+        validate_key = name_validator(min_length=1, max_length=120)
+        self.validate_key = validate_key
```
scosman: saving a function to local var is strange. Just create it when you need it? P2

libs/core/kiln_ai/utils/filesystem_cache.py
```diff
@@ -0,0 +1,57 @@
...
+        return self.cache_dir_path / key
+
+    def get(self, key: str) -> bytes | None:
+        # check if the file exists
+        if not self.get_path(key).exists():
+            return None
+
+        # we don't want to raise because of internal cache corruption issues
+        try:
+            with open(self.get_path(key), "rb") as f:
```
scosman: make this async

libs/core/kiln_ai/utils/filesystem_cache.py
```diff
@@ -0,0 +1,57 @@
...
+                return f.read()
+        except Exception:
+            logger.error(f"Error reading file {self.get_path(key)}", exc_info=True)
+            return None
+
+    def set(self, key: str, value: bytes) -> Path:
+        logger.debug(f"Caching {key} at {self.get_path(key)}")
+        self.validate_key(key)
+        path = self.get_path(key)
+        path.write_bytes(value)
```
scosman: also async

libs/core/kiln_ai/utils/filesystem_cache.py
leonardmq: I thought you meant not assigning the function to any variable - so moved it into its own method in https://github.com/Kiln-AI/Kiln/pull/626/commits/92abcfd714bea29469ec8bd7f01521de0fdfdcce. 

But maybe you meant not assigning it to `validate_key` (and just assigning to `self.validate_key` directly)

libs/core/kiln_ai/utils/filesystem_cache.py
leonardmq: Refactored to use `anyio` - we already had it in `core` via transitive dependencies (`groq`, `mcp` and other things already needed it), but it is now added explicitly as well.

https://github.com/Kiln-AI/Kiln/pull/626/commits/92abcfd714bea29469ec8bd7f01521de0fdfdcce

libs/core/kiln_ai/utils/filesystem_cache.py
leonardmq: Refactored to use `anyio` - we already had it in `core` via transitive dependencies (`groq`, `mcp` and other things already needed it), but it is now added explicitly as well.

https://github.com/Kiln-AI/Kiln/pull/626/commits/92abcfd714bea29469ec8bd7f01521de0fdfdcce

libs/core/kiln_ai/utils/filesystem_cache.py
scosman: this works

libs/core/kiln_ai/utils/test_filesystem_cache.py
leonardmq: This is fixed too

libs/core/kiln_ai/utils/test_filesystem_cache.py
leonardmq: This is fixed too

scosman (approved):

PR #636, opened 2025-09-20
uncache search tools in UI when you create, archive or unarchive

scosman:
## Summary by CodeRabbit

- **Bug Fixes**
  - The list of available tools now refreshes immediately after creating a RAG config, saving a template, or archiving/unarchiving a config, preventing stale tool options from appearing in the Docs UI.
  - Ensures the tools list is up to date before navigating back to the RAG config list or recording related events, improving accuracy and consistency for end-users.


leonardmq (approved):

scosman:
@leonardmq missed that comment, but added an item in trello to do it

PR #647, opened 2025-09-23
refactor: split up extractor list and add mistral ocr extractor

leonardmq:
## What does this PR do?

Some extractors work very differently from normal models - they have different constraints, endpoints, outputs and so on. For example, `docling` and `MistralOCR`. They do not support inference for runs either, and as such have very little overlap with other models. Making them fit the shape of normal models risks making both extractors and models more fragile.

Changes:
- refactor: split up the extractor list from the ml_model_list
- feat: add Mistral OCR

WIP:
- provider connection UI not set up yet - but can test by setting `MISTRAL_API_KEY=` in the `.env`
- the parts that work are also unfinished

### Open questions:
1. Should the extractor list and the `ml_model_list` be completely decoupled? An alternative would be to dynamically append the extractors that are derived from the `ml_model_list`. The only requirement for a model to be an extractor is that it supports some mimetypes - that information will need to be in the `ml_model_list` in any case for when we add support for file inputs in messages.
2. Mistral OCR requires an API key, but for an extractor in this case. Should it be displayed among the other providers, or in a somewhat distinct UI?

## Summary by CodeRabbit

- New Features
  - Added support for Mistral as a connectable provider, including API key connect/disconnect.
  - Introduced Mistral OCR extractor and broader extractor support.
  - New “Available Extractors” selector in the extractor creation form, grouped by provider with MIME type hints.
  - Provider imagery updated to include the Mistral logo.

- Refactor
  - Unified extractor handling across server and UI, enabling listing and selection of both model-backed and non-model extractors.

- Chores
  - Added runtime dependency for Mistral client.

### DIFF
--- a/app/desktop/studio_server/provider_api.py
+++ b/app/desktop/studio_server/provider_api.py
@@ -13,6 +13,7 @@ from kiln_ai.adapters.docker_model_runner_tools import (
     DockerModelRunnerConnection,
     get_docker_model_runner_connection,
 )
+from kiln_ai.adapters.extractor_list import ExtractorProviderName, built_in_extractors
 from kiln_ai.adapters.ml_embedding_model_list import (
     EmbeddingModelName,
     KilnEmbeddingModel,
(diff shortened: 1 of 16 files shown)


leonardmq:
Mistral OCR here is added as more of a proof of concept / sanity check - might remove to keep the PR focused on the refactoring

leonardmq:
PR is stale - closing for now. Might still do at some point

PR #648, opened 2025-09-24
MCP Server for Kiln Tasks (RAG, Task as Tools)

scosman:
## Summary
- replace the FastAPI MCP application with a fastmcp-based server and supporting runtime helpers
- add a tool registry that prepares Kiln tool contexts for MCP and update the CLI/tests accordingly
- depend directly on mcp instead of fastapi-mcp and drop the workspace override pin

## Testing
- uv run pytest libs/server/kiln_server/mcp/test_mcp_server.py
- uv run ruff check libs/server/kiln_server/mcp

------
https://chatgpt.com/codex/tasks/task_e_68d20bcefca48332ac71a5a13da1a473

## Summary by CodeRabbit

* **New Features**
  * Kiln MCP Server (beta) with kiln_mcp CLI: run over stdio/SSE/streamable-http, host/port, mount path, log level, and selective tool exposure; CLI-driven REST server options.

* **Documentation**
  * Updated main README (commands, REST API wording) and added MCP server README with setup, usage, and examples.

* **Chores**
  * Added MCP dependency and registered kiln_server / kiln_mcp CLI entry points.

* **Tests**
  * Added comprehensive MCP server tests for discovery, context preparation, server calls, transports, and CLI parsing.

### DIFF
--- a/libs/server/kiln_server/mcp/README.md
+++ b/libs/server/kiln_server/mcp/README.md
@@ -50,7 +50,7 @@ usage: kiln_mcp [-h] [--tool-ids TOOL_IDS] [--transport {stdio,sse,streamable-ht
 Run the Kiln MCP server for a project.
 
 positional arguments:
-  project               Path to the kiln.project file
+  project               Path to the project.kiln file
 
 options:
   -h, --help            show this help message and exit
--- a/libs/server/kiln_server/mcp/mcp.py
+++ b/libs/server/kiln_server/mcp/mcp.py
@@ -105,25 +105,37 @@ def _list_tools(project_path: Path) -> None:
     stdout.write("Listing all available tools\n\n")
     stdout.write("Search Tools / RAG (ID -- name -- description):\n")
     has_rag_tools = False
+    archived_rag_tools = 0
     for rag_config in project.rag_configs(readonly=True):
+        if rag_config.is_archived:
+            archived_rag_tools += 1
+            continue
         stdout.write(
             f"{build_rag_tool_id(rag_config.id)} -- {rag_config.tool_name} -- {rag_config.tool_description}\n"
         )
         has_rag_tools = True
     if not has_rag_tools:
         stdout.write("No RAG tools found.\n")
+    if archived_rag_tools > 0:
+        stdout.write(f"Archived RAG tool count (not listed): {archived_rag_tools}\n")
 
     stdout.write("\nKiln Task Tools (ID -- name  -- description):\n")
     has_kiln_task_tools = False
+    archived_kiln_task_tools = 0
     for tool_server in project.external_tool_servers(readonly=True):
-        if (
-            tool_server.type == ToolServerType.kiln_task
-            and not tool_server.properties.get("is_archived", False)
-        ):
+        if tool_server.type == ToolServerType.kiln_task:
+            if tool_server.properties.get("is_archived", False):
+                archived_kiln_task_tools += 1
+                continue
             stdout.write(
                 f"{build_kiln_task_tool_id(tool_server.id)} -- {tool_server.name} -- {tool_server.description or 'No description'}\n"
             )
             has_kiln_task_tools = True
+    if archived_kiln_task_tools > 0:
+        stdout.write(
+            f"Archived Kiln task tool count (not listed): {archived_kiln_task_tools}\n"
+        )
+
     if not has_kiln_task_tools:
         stdout.write("No Kiln task tools found.\n")
 


libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -17,7 +32,7 @@ async def run(self, **kwargs) -> Any:
         pass
 
     @abstractmethod
-    async def toolcall_definition(self) -> Dict[str, Any]:
+    async def toolcall_definition(self) -> ToolCallDefinition:
```
leonardmq, line 44: Might want to update the return types on the concrete class implementations too

libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -65,7 +80,7 @@ async def name(self) -> str:
     async def description(self) -> str:
         return self._description
 
-    async def toolcall_definition(self) -> Dict[str, Any]:
+    async def toolcall_definition(self) -> ToolCallDefinition:
```
leonardmq, line 92: Might want to update the return types on the concrete class implementations too

libs/server/kiln_server/server.py
```diff
@@ -49,12 +50,41 @@ def ping():
...
+        "--log-level",
+        default="info",
+        help="Log level for the server when using network transports.",
+    )
+    parser.add_argument(
+        "--auto-reload",
+        default=False,
+        help="Enable auto-reload for the server.",
+    )
+    return parser
```
leonardmq, line 81: Coderabbit is correct about `--auto-reload` parsing needing `action="store_true"`

libs/server/kiln_server/mcp/README.md
```diff
@@ -0,0 +1,64 @@
...
+This MCP server allows you to call Kiln tools, such as a Kiln Search Tool, from any MCP client.
+
+**Beta** This is a beta intended for local usage. It isn't designed for production workloads.
+
+### Installation
+
+We suggest installing with uv:
+
+```bash
+uv tool install kiln_server
```
leonardmq, line 14: If I understand the install: we first install the whole library with `uv tool install kiln_server`, which exposes two scripts (`kiln_server` and `kiln_mcp`) in `pyproject.toml` and they can be run as CLI.

Is this correct?

libs/server/kiln_server/mcp/mcp_server_tool_utils.py
```diff
@@ -0,0 +1,107 @@
...
+) -> list[ToolContext]:
+    """Prepare ``ToolContext`` objects for the provided tool resolutions."""
+
+    contexts: list[ToolContext] = []
+    seen_names: set[str] = set()
+
+    for resolution in resolutions:
+        context = await _build_tool_context(resolution)
+        tool_name = context.definition.name
+        if tool_name in seen_names:
```
leonardmq, line 102: The UI lets the user create duplicate tool names. If you load such a project, you get hit with this error on startup and it is not immediately clear what the user should do about it. Could add a suggestion to tell them to select specific tool-ids (for tools that do not have the same name) in the error message

libs/server/kiln_server/mcp/mcp.py
```diff
@@ -0,0 +1,114 @@
...
+        return None
+    return [token.strip() for token in raw.split(",") if token.strip()]
+
+
+def build_argument_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description="Run the Kiln MCP server for a project."
+    )
+    parser.add_argument("project", type=Path, help="Path to the project.kiln file")
+    parser.add_argument(
```
leonardmq, line 34: Users do not know what tool ids are or what they refer to or how we build them. For RAG Configs, it is `kiln_tool::rag::{rag config id}`, but we do not expose that in the UI.

In the Search Tool (RAG) detailed page, in the `Properties` panel, we show an `ID`, which is the ID of the RAG Config - above the `Tool name` and `Tool description`. Maybe we should show the `Tool ID` instead or in addition to the RAG Config ID.

Could also add a command in the CLI to `list-tools`

libs/server/kiln_server/mcp/tool_selection.py
```diff
@@ -0,0 +1,111 @@
...
+    return bool(getattr(config, "is_archived", False))
+
+
+def _iterate_rag_configs(project: Project) -> Iterable[RagConfig]:
+    """Yield the project's RAG configurations in read-only mode."""
+
+    return project.rag_configs(readonly=True)
+
+
+def _build_rag_tool_id(rag_config: RagConfig) -> str:
```
leonardmq: Can we move this into a common package (probably here: https://github.com/Kiln-AI/Kiln/blob/97a0b4be30fc0abd54a8d99a7ad09d51a98823f6/libs/core/kiln_ai/datamodel/tool_id.py#L106) so we can also use it here: https://github.com/Kiln-AI/Kiln/blob/97a0b4be30fc0abd54a8d99a7ad09d51a98823f6/app/desktop/studio_server/tool_api.py#L250

libs/server/kiln_server/mcp/tool_selection.py
```diff
@@ -0,0 +1,111 @@
...
+    tool: KilnToolInterface
+
+
+ToolFactory = Callable[[str, RagConfig], KilnToolInterface]
+
+
+def _default_rag_tool_factory(tool_id: str, rag_config: RagConfig) -> KilnToolInterface:
+    """Instantiate a :class:`~kiln_ai.tools.rag_tools.RagTool`."""
+
+    from kiln_ai.tools.rag_tools import RagTool
```
leonardmq: Why not import at the top?

libs/server/kiln_server/mcp/tool_selection.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return RagTool(tool_id, rag_config)
+
+
+def _is_archived(config: RagConfig) -> bool:
+    """Check whether a RAG configuration is archived."""
+
+    archived_flag = getattr(config, "archived", None)
```
leonardmq: The flag is `is_archived` (I don't recall `archived` being a possibility, and if it was, it would no longer work) so seems like this `_is_archived` method is not needed?

libs/server/kiln_server/mcp/tool_selection.py
```diff
@@ -0,0 +1,111 @@
...
+    if archived_flag is not None:
+        return bool(archived_flag)
+    return bool(getattr(config, "is_archived", False))
+
+
+def _iterate_rag_configs(project: Project) -> Iterable[RagConfig]:
```
leonardmq: Feels like this whole method can be removed, and it does not seem to be used or mock in tests either

libs/server/kiln_server/mcp/tool_selection.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+            continue
+
+        tool = factory(tool_id, rag_config)
+        resolutions.append(ToolResolution(tool_id=tool_id, tool=tool))
+        missing_ids.discard(tool_id)
+
+    if missing_ids:
+        raise ValueError(
+            "Requested tool IDs were not found or are not eligible: "
```
leonardmq: If "eligible" only means "not archived", we should probably say "archived" instead of "not eligible"

libs/server/kiln_server/mcp/runtime.py
```diff
@@ -0,0 +1,105 @@
...
+        if context is None:
+            raise ValueError(f"Unknown tool requested: {name}")
+
+        try:
+            result = await context.resolution.tool.run(**arguments)
+        except Exception as exc:  # pragma: no cover - defensive logging
+            logger.exception("Tool %s failed", name)
+            raise ValueError(str(exc)) from exc
+
+        if context.requires_structured_output:
```
leonardmq, line 81: Shouldn't validating the output against an output schema already be handled in the `tool.run()` implementation?

libs/server/kiln_server/mcp/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Once installed, you can run `kiln_mcp` from your terminal. The last parameter must be the path to the project.kiln file you wish to use. Example:
+
+```bash
+kiln_mcp --transport streamable-http "/Users/username/Kiln Projects/Project Name/project.kiln"
+```
+
+### Running kiln_mcp in Cursor, VSCode, etc
+
+Many MCP clients require running MCP server over the stdio transport. See this example `mcp.json`:
```
leonardmq: Some people (such as me a few weeks ago) could think that they need to do this plus in addition to having the server running via CLI. Maybe helpful to add a clarification like "- in which case you do not need to run the server explicitly yourself" (if there is a good less verbose way to phrase this)

libs/server/kiln_server/mcp/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```json
+{
+  "mcpServers": {
+    "kilnMCP": {
+      "command": "kiln_mcp",
+      "args": ["/Users/username/Kiln Projects/Project Name/project.kiln"]
+    }
+  }
+}
```
leonardmq, line 39: Unrelated to this line, but bridging UI with code is difficult at the moment as you need to manually navigate in a lot of places in the UI to collect IDs, paths, etc. To understand what goes where, you more or less need to understand the data model and how it works under the hood.

One way I see that could make the bridging much smoother is in the corner on specific pages, have a code button, you click it and get a dialog sort of UI with snippets of Python for the particular thing (e.g. RAG config, Run config, etc.) the page is about.

Could have something like that on the `Search Tool (RAG)` detail page where you would get both the loader code for the config with ID prefilled, etc. and the MCP code

scosman: I like this. We'll see if people ask (and make sure MCP server is stable) before adding to the UI, but will be really cool

libs/server/kiln_server/mcp/runtime.py
scosman: tool.run() is just an interface, not a base class. No guarantee if validates.

libs/server/kiln_server/mcp/README.md
scosman: yup!

Can do `uv tool install --force ./lib/server` for a local install to try it

PR #659, opened 2025-09-26
fix: rag store does not clear on project switch

leonardmq:
## What does this PR do?

RAG artefacts (documents, configs, etc.) are parented under `project`. The store where we keep progress data does not update on project switching - if a user is on Project A, and switches through internal navigation to Project B, they will see their Project A's RAG configs.

Fix:
- refactor the store to key all the rag data under `project_id` and add a `currentProjectRagProgressStore` derived store that subscribes to the current project id (that was already there in the `ui_state` store)

## Summary by CodeRabbit

- New Features
  - Per-project RAG progress and controls: run all or individual configs within the selected project.
  - Per-project overall completion percentage and sorting of configs by creation date.

- Bug Fixes
  - Errors, running states, logs, and archive/unarchive actions are isolated and reliably applied per project.

- Refactor
  - Global progress/state handling migrated to project-scoped state for consistent UI behavior.

### DIFF
--- a/app/web_ui/src/lib/stores/rag_progress_store.ts
+++ b/app/web_ui/src/lib/stores/rag_progress_store.ts
@@ -374,6 +374,7 @@ function createRagProgressStore() {
     subscribe,
     set,
     update,
+    updateProjectState,
     run_all_rag_configs,
     run_rag_config,
     getProjectState,
@@ -464,10 +465,7 @@ export async function load_all_rag_config_progress(projectId: string) {
 
     if (!error && data) {
       for (const [rag_config_id, progress] of Object.entries(data)) {
-        ragProgressStore.update((globalState) => {
-          const projectState =
-            globalState[projectId] ||
-            ragProgressStore.getProjectState(projectId)
+        ragProgressStore.updateProjectState(projectId, (projectState) => {
           const newProgress = {
             ...projectState.progress,
             [rag_config_id]: {
@@ -486,23 +484,17 @@ export async function load_all_rag_config_progress(projectId: string) {
           }
 
           return {
-            ...globalState,
-            [projectId]: {
-              ...projectState,
-              progress: newProgress,
-              status: newStatus,
-            },
+            ...projectState,
+            progress: newProgress,
+            status: newStatus,
           }
         })
       }
     }
   } catch (e) {
-    ragProgressStore.update((globalState) => ({
-      ...globalState,
-      [projectId]: {
-        ...ragProgressStore.getProjectState(projectId),
-        error: createKilnError(e),
-      },
+    ragProgressStore.updateProjectState(projectId, (projectState) => ({
+      ...projectState,
+      error: createKilnError(e),
     }))
   }
 }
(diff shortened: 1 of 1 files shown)


app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
...
+    if (!currentProjectId) {
+      return ragProgressStore.getProjectState(null)
+    }
+    return (
+      // return the project state if it exists, otherwise return new default project state
+      ragProgress.projects[currentProjectId] ||
+      ragProgressStore.getProjectState(currentProjectId)
+    )
+  },
+)
```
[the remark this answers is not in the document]
leonardmq: The dashboard is always on a project - when no project is selected, the user lands on a page where all they can do is select a project. So should be fine

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -531,11 +626,20 @@ export async function update_rag_config_archived_state(
...
+      [project_id]: {
+        ...ragProgressStore.getProjectState(project_id),
+        is_archived: {
+          ...ragProgressStore.getProjectState(project_id).is_archived,
+          [rag_config_id]: is_archived,
+        },
+      },
     },
   }))
 }
```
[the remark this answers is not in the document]
leonardmq: They are tracked separately at the moment - was intentional - but this whole store is due for a refactor (we will move a lot of the logic related to progress tracking to the backend)

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
...
 export const ragProgressStore = createRagProgressStore()
 
-export const allRagConfigs = derived(ragProgressStore, ($store) => {
-  return sortRagConfigs(Object.values($store.rag_configs), "created_at")
-})
+// derived store that provides the rag state for the currently selected project
+export const currentProjectRagProgressStore = derived(
+  [ragProgressStore, ui_state],
+  ([ragProgress, uiState]) => {
+    const currentProjectId = (uiState as UIState).current_project_id
```
scosman: We should remove this. The library has a project ID in the URL, which may or may not be the current project. We should respect the exact project they link to, and ignore current_project_id.
In general: current_project_id is deprecated. There's going to be one layout allowed to use it (just on fresh page load, just to redirect to a proper URL with project_id in it), but the URL based parameters are the source of truth, and current_project_id is the source of bugs. It gets messed up in lots of ways (deeplinks, multiple tabs, etc)
Just use something like $: progressStore = ragProgressStore[$page.params.project_id] in callers, it works great

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
...
+      // error
+      error: null,
+
+      // last started rag config id
+      last_started_rag_config_id: null,
+    }
+  }
+
+  function getCurrentProjectId(): string | null {
+    return (get(ui_state) as UIState).current_project_id
```
scosman: delete

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
...
+  }
+
+  function getProjectState(
+    project_id: string | null,
+  ): RagConfigurationProgressState {
+    if (!project_id) return getDefaultProjectState()
```
scosman: raise exception? Seems invalid, and will hide that null is a possible value (which is is if we're allowing null project_ids), which will cause bugs

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
   last_started_rag_config_id: string | null
 }
 
+interface ProjectRagProgressState {
+  projects: Record<string, RagConfigurationProgressState>
```
scosman: nit: naming. This doesn't contain projects, it contains rag_config_state_by_project_id

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -25,45 +26,103 @@ interface RagConfigurationProgressState {
...
-  const { subscribe, set, update } = writable<RagConfigurationProgressState>({
-    // rag configuration id -> rag config
-    rag_configs: {},
+  const { subscribe, set, update } = writable<ProjectRagProgressState>({
+    projects: {},
+  })
 
-    // rag configuration id -> progress; only applies to rag configurations that are running
-    progress: {},
+  function getDefaultProjectState(): RagConfigurationProgressState {
```
scosman: Feels like giant refactor isn't needed?

Can we just do this:

```
export const ragProgressStore = writable<ProjectRagProgressState>({
    projects: {},
}
```

But then otherwise leave this class alone? Caller can get the right object by ID, then use it. Feels strange that every function needs to take in an ID.

scosman (changes requested):

app/web_ui/src/lib/stores/rag_progress_store.ts
leonardmq: `$: progressStore = ragProgressStore[$page.params.project_id]` does not work, because we need to initialize the state for the project if it does not already exist - and initialization shape is complex.

But now thinking, why not create one store instance per project and access the whole store instance like `$: progressStore = ragProgressStores[$page.params.project_id]`, or was that what you meant?

leonardmq:
Alternative version without derived store and where  consumers access the state object directly, and where all the initialization logic is done in the consumers: 
- https://github.com/Kiln-AI/Kiln/tree/leonard/kil-57-bug-rag-store-stays-there-on-project-switch-no-derived

It seems much worse because every consumer needs to initialize the internal store state in a flaky way instead of having the progress lazy-initialized

app/web_ui/src/lib/stores/rag_progress_store.ts
```diff
@@ -60,11 +54,48 @@ function createRagProgressStore() {
...
+      state: RagConfigurationProgressState,
+    ) => RagConfigurationProgressState,
+  ) {
+    update((globalState) => ({
+      ...globalState,
+      [project_id]: updater(
+        globalState[project_id] || getDefaultProjectState(),
+      ),
+    }))
+  }
```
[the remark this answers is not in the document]
leonardmq: `project_id` should always be there. Any reason to check it even though we don't allow the type to be null?

PR #660, opened 2025-09-26
perf: much faster tagging on large datasets

leonardmq:
## What does this PR do?

The current tagging on `TaskRun` and `Document` is slow with large datasets mostly because we do a linear scan in the parent folder for every child model (`TaskRun` and `Document`) we are looking for. The tagging is slower as the dataset size (number of files in the directory to scan) and the number of models to tag (number of times we need to do that scanning) increases.

Fix:
- add `from_ids_and_parent_path` on `KilnParentedModel` - a variant of the single lookup (`from_id_and_parent_path`) that collects all the ids in one scan
- refactor `TaskRun` and `Document` tagging to batch the lookups (in smaller batches to avoid filling up memory) with the bulk lookup method

The FS methods on the `KilnParentedModel` are synchronous, which does not help - I tried implementing an async variant of the `save_to_file` method (using `anyio`) and running those in parallel, but that did not have any particular positive performance impact. 

Benchmark results:

The benchmark test for `from_ids_and_parent_path` is in `test_from_ids_and_parent_path_benchmark`. It compares how long it takes to look up the children in bulk versus individually.

This test assumes we retrieve all the children in the dataset (the case during `Select All > Tagging`).

| Model Count | Bulk Duration (s) | Bulk Ops/sec | Individual Duration (s) | Individual Ops/sec | Bulk Speedup (× faster) | Time Savings |
|---------------------------------|-------------------|--------------|--------------------------|---------------------|---------------------|--------------|
| 100                             | 0.0050            | 20042.59     | 0.1578                   | 633.76             | 31.62×              | 96.8%        |
| 1,000                           | 0.0425            | 23511.94     | 6.6871                   | 149.54             | 157.23×             | 99.4%        |
| 2,500                           | 0.1068            | 23406.58     | 42.7666                  | 58.46              | 400.41×             | 99.8%        |
| 5,000                           | 0.2175            | 22985.72     | 175.0355                 | 28.57              | 804.66×             | 99.9%        |

Here is a benchmark for the tagging speed now - still becomes slower as size grows, but orders of magnitude faster than it was:
| Number of Runs | Total Duration (s) | Runs/sec | Time per Run (ms) |
|----------------|---------------------|----------|-------------------|
| 100            | 0.023               | 4293.4   | 0.23              |
| 1,000          | 0.407               | 2457.8   | 0.41              |
| 10,000         | 7.837               | 1275.9   | 0.78              |
| 50,000         | 89.701              | 557.4    | 1.79              |

## Checklists


libs/server/kiln_server/document_api.py
```diff
@@ -729,12 +729,20 @@ async def edit_tags(
...
-            if not document:
-                failed_documents.append(document_id)
-            else:
+
+        # all the runs we need to tag
+        doc_ids_set: set[str] = set(document_ids)
+        docs_found_set: set[str] = set()
+
+        batch_size = 500
+        for i in range(0, len(document_ids), batch_size):
```
scosman, line 738: Add a `await sleep(0)` per batch, since we're not properly async?

libs/server/kiln_server/run_api.py
```diff
@@ -243,12 +243,18 @@ async def edit_tags(
...
-                failed_runs.append(run_id)
-            else:
+
+        # all the runs we need to tag
+        run_ids_set: set[str] = set(run_ids)
+        runs_found_set: set[str] = set()
+
+        batch_size = 500
+        for i in range(0, len(run_ids), batch_size):
+            batch_run_ids = run_ids[i : i + batch_size]
```
scosman, line 257: ditto: sleep(0)

libs/server/kiln_server/document_api.py
leonardmq: Good catch, I had it but got rid of it by accident during clean up - added it back just now. 

To confirm also that it does work like we think it does - dashboard is totally blocked because the server cannot process any incoming request while blocked and it works decently when we have the sleep

PR #665, opened 2025-09-27
Launch stdio server in cache path

aiyakitori:
## Checklists


libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -164,10 +164,15 @@ async def _create_local_mcp_session(
             env_vars["PATH"] = self._get_path()
 
         # Set the server parameters
+        # On Windows, set cwd to cache directory to avoid permission issues
```
scosman: Why only windows?

scosman (approved):

PR #669, opened 2025-09-29
feat: loader with lancedb cloud support

leonardmq:
## What does this PR do?

We should make it easy to port a RagConfig over to a prod environment. A first step towards that is having a simple way to load existing data into an off-Kiln vector store as a static corpus. 

Changes:
- refactor internals of `LanceDBAdapter` for reuse in the loader
- add `LanceDBLoader` that exposes an iterator to get the data as llama_index `TextNode` and another method that writes the nodes to an existing LanceDB deployment using their own LanceDB client (including support for LanceDB Cloud and other such deployment options)

The `LanceDBLoader` is distinct from the `LanceDBAdapter` because it exposes a simplified, higher-level interface without internals (e.g. traversing the filesystem to collect the target models, understanding how they relate to one another, etc.), without the idiosyncrasies of our own requirements (avoiding duplicate inserts, trying to recover from partial inconsistencies, etc.), data types and complex dependency initialization.

## Checklists

## Summary by CodeRabbit

- New Features
  - Introduces a VectorStoreLoader to export RAG data as LlamaIndex-compatible nodes with async batching.
  - Enhances LanceDB integration with configurable query types, deterministic chunk IDs, and optional store injection for flexible deployments.

- Refactor
  - Simplifies node creation and ID handling for more predictable behavior across vector store operations.

- Documentation
  - Adds a “Taking Kiln RAG to production” guide, including an end-to-end LanceDB Cloud example.

- Tests
  - Adds comprehensive tests for the loader, LanceDB helpers, and adapter behavior across local and cloud setups.


libs/core/kiln_ai/adapters/vector_store_loaders/test_lancedb_loader.py
```diff
@@ -0,0 +1,357 @@
...
+@pytest.mark.parametrize(
+    "vector_store_type",
+    [
+        VectorStoreType.LANCE_DB_FTS,
+        VectorStoreType.LANCE_DB_VECTOR,
+        VectorStoreType.LANCE_DB_HYBRID,
+    ],
+)
+@pytest.mark.paid
+async def test_lancedb_loader_insert_nodes_lancedb_cloud(
```
leonardmq, line 227: This test uses the real LanceDB Cloud

libs/core/kiln_ai/adapters/vector_store_loaders/lancedb_loader.py
```diff
@@ -0,0 +1,58 @@
...
+        rag_config: RagConfig,
+        vector_store_config: VectorStoreConfig,
+        lancedb_vector_store: LanceDBVectorStore,
+    ):
+        self.project = project
+        self.rag_config = rag_config
+        self.lancedb_vector_store = lancedb_vector_store
+        self.vector_store_config = vector_store_config
+
+    async def iter_llama_index_nodes(self) -> AsyncGenerator[TextNode, None]:
```
scosman: can we move this into the base? Doesn't seem lancedb specific? It will make it easier to sub-class.

libs/core/kiln_ai/adapters/vector_store_loaders/test_lancedb_loader.py
```diff
@@ -0,0 +1,282 @@
...
+
+    lancedb_region = os.getenv("LANCE_DB_REGION")
+    assert lancedb_region is not None, (
+        "LANCE_DB_REGION is not set - test requires lancedb cloud"
+    )
+    return LanceDBCloudEnvVars(
+        uri=lancedb_uri,
+        api_key=lancedb_api_key,
+        region=lancedb_region,
+    )
```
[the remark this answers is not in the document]
leonardmq: Test is marked as `@pytest.mark.paid`, which means it only runs when we run this one test explicitly (otherwise skipped). As it is only ever going to run when we are trying to run this one test, then it makes sense to just throw if the preconditions (env vars) are not set.

PR #674, opened 2025-10-01
Tasks as Tools

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Kiln Task Tools: create/edit/list/detail UI, API endpoints, surfaced in tool selectors and available tools with archival state and counts.
  * Run page UX: saved run-config dropdown, separate Tools selector, per-task prompt loading/isolation.
  * Trace UX: clickable Kiln Task links, Tool Messages dialog, and improved tool-call rendering.
  * Misc UI: Fancy Select disabled prop, dataset run link helper, upload accessible by Enter/Space, View Logs action.

* **Bug Fixes**
  * Kiln Task tools auto-archived when their task is deleted; archived tools excluded from listings.

* **Tests**
  * Expanded coverage for Kiln Task tooling, tool IDs, tool execution, session/error flows, and PDF/image extractor behavior.

---

> [!NOTE]
> Introduce Kiln Tasks as callable tools (create/edit/list/use), extend tool/data models and adapters, and update UI/trace/flows to select and manage them alongside MCP and RAG tools.
> 
> - **Backend/API**:
>   - Add `kiln_task` tool servers (`ToolServerType.kiln_task`) with `KilnTaskServerProperties`, endpoints to add/edit (`add_kiln_task_tool`, `edit_kiln_task_tool`), list (`/kiln_task_tools`), and include in `/available_tools`.
>   - Implement `KilnTaskTool` and register in `tool_registry`; parse `kiln_task::<server_id>` IDs.
>   - Disallow kiln_task via MCP session; `available_mcp_tools` now errors for `kiln_task` type.
>   - Extend tool sets with typed grouping (`ToolSetType`: search/mcp/kiln_task/demo).
> - **Datamodel**:
>   - Add `DataSourceType.tool_call` and propagate through `DataSource`/`TaskOutput`.
>   - Enhance `BaseAdapter.generate_run` to accept `output_source_type`; include conditional properties; support tool call context.
>   - `KilnToolInterface` gains `run_with_context` and `ToolCallContext`; `LiteLlmAdapter` uses it.
>   - Update `tool_id` helpers with kiln task prefix and parser.
> - **UI**:
>   - New flows/pages to create, view, list, and archive Kiln Task tools; integrate into Manage Tools.
>   - Tool selection refactor (`ToolsSelector`), typed tool sets ordering, and disabled FancySelect support.
>   - Run config component split: externalized save, prompts/run-config stores, and improved error handling.
>   - Trace links to tool detail and dataset run; improved content parsing.
>   - Minor UX: property list wrapping/info text; badges use "＋"; settings section reorder.
> - **Types/OpenAPI**:
>   - Update schema/types for new servers, endpoints, tool set typing, and datasource/tool-call additions.
> - **Tests**:
>   - Add coverage for kiln_task exclusions in MCP tools and updated adapter behaviors; numerous UI/backend test updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 49d9316764f8d84b64290e26d86830d5dd45a478. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/web_ui/src/lib/ui/fancy_select.svelte
+++ b/app/web_ui/src/lib/ui/fancy_select.svelte
@@ -325,11 +325,17 @@
 
   // Handle key input when dropdown is open
   function handleKeyInput(event: KeyboardEvent) {
-    // Don't interfere with navigation keys or if we're already focused on search input
+    // Don't interfere if we're already focused on search input
     if (isSearching && document.activeElement === searchInputElement) {
       return
     }
 
+    // Don't interfere if the fancy select is not focused either
+    if (document.activeElement !== selectedElement) {
+      return
+    }
+
+    // Don't interfere with navigation keys
     if (
       event.key === "ArrowDown" ||
       event.key === "ArrowUp" ||
--- a/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
@@ -167,9 +167,9 @@
     }
   }
 
-  export async function save_new_run_config() {
+  export async function save_new_run_config(): Promise<TaskRunConfig | null> {
     if (!project_id || !current_task.id) {
-      return
+      return null
     }
     try {
       save_config_error = null
@@ -180,14 +180,14 @@
       )
       // Reload prompts to update the dropdown with the new static prompt that is made from saving a new run config
       await load_task_prompts(project_id, current_task.id)
-      if (saved_config.id) {
-        selected_run_config_id = saved_config.id
-      } else {
+      if (!saved_config || !saved_config.id) {
         throw new Error("Saved config id not found")
       }
+      return saved_config
     } catch (e) {
       save_config_error = createKilnError(e)
     }
+    return null
   }
 
   async function get_selected_run_config(): Promise<
--- a/app/web_ui/src/lib/ui/run_config_component/saved_run_configs_dropdown.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/saved_run_configs_dropdown.svelte
@@ -33,7 +33,8 @@
   export let project_id: string
   export let current_task: Task
   export let selected_run_config_id: string | null = null // This will be null until the default_run_config_id is set
-  export let save_new_run_config: (() => Promise<void>) | null = null
+  export let save_new_run_config: (() => Promise<TaskRunConfig | null>) | null =
+    null
   export let save_config_error: KilnError | null = null
   export let set_default_error: KilnError | null = null
   export let info_description: string = ""
@@ -180,7 +181,10 @@
 
   async function handle_save() {
     if (save_new_run_config) {
-      await save_new_run_config()
+      const saved_run_config = await save_new_run_config()
+      if (saved_run_config?.id) {
+        selected_run_config_id = saved_run_config.id
+      }
     }
   }
 
--- a/app/web_ui/src/routes/(app)/run/+page.svelte
+++ b/app/web_ui/src/routes/(app)/run/+page.svelte
@@ -6,7 +6,7 @@
   import { KilnError } from "$lib/utils/error_handlers"
   import Run from "./run.svelte"
   import { client } from "$lib/api_client"
-  import type { TaskRun } from "$lib/types"
+  import type { TaskRun, TaskRunConfig } from "$lib/types"
   import RunInputForm from "./run_input_form.svelte"
   import posthog from "posthog-js"
   import { tick } from "svelte"
@@ -148,15 +148,16 @@
     }
   }
 
-  async function handle_save_new_run_config() {
+  async function handle_save_new_run_config(): Promise<TaskRunConfig | null> {
     try {
       if (!run_config_component) {
         throw new Error("Run config component is not loaded")
       }
-      await run_config_component.save_new_run_config()
+      return await run_config_component.save_new_run_config()
     } catch (e) {
       save_config_error = createKilnError(e)
     }
+    return null
   }
 </script>
 
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/kiln_task/add_kiln_task_tool.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/add_tools/kiln_task/add_kiln_task_tool.svelte
@@ -71,8 +71,16 @@
       .replace(/^_|_$/g, "")
   }
 
-  $: if (selected_task) {
-    name = to_snake_case(selected_task.name)
+  $: selected_task, handle_task_change()
+
+  function handle_task_change() {
+    clear_error_if_present()
+    selected_run_config_id = null
+    name = ""
+    description = ""
+    if (selected_task) {
+      name = to_snake_case(selected_task.name)
+    }
   }
 
   $: task_options = data_loaded ? create_task_options(tasks) : []
@@ -151,7 +159,10 @@
   async function create_new_run_config() {
     loading = true
     if (run_config_component) {
-      await run_config_component.save_new_run_config()
+      const saved_run_config = await run_config_component.save_new_run_config()
+      if (saved_run_config?.id) {
+        selected_run_config_id = saved_run_config.id
+      }
     }
     loading = false
   }
@@ -258,7 +269,6 @@
           fancy_select_options={task_options}
           disabled={!data_loaded}
           description="The Kiln task the tool will call."
-          on:change={clear_error_if_present}
         />
 
         {#if selected_task}
@@ -324,7 +334,6 @@
         <RunConfigComponent
           bind:this={run_config_component}
           {project_id}
-          bind:selected_run_config_id
           current_task={selected_task}
           hide_create_kiln_task_tool_button={true}
           {save_config_error}


app/desktop/studio_server/tool_api.py
```diff
@@ -256,10 +337,45 @@ async def get_available_tool_servers(
...
+                                tool_description=tool_server.properties.get(
+                                    "description"
+                                ),
+                                task_id=task_id,
+                                task_name=task.name,
+                                task_description=task.description,
+                                is_archived=tool_server.properties.get(
+                                    "is_archived", False
+                                ),
+                                created_at=tool_server.created_at.isoformat(),
```
scosman: return an actual datetime. We have full type support for datetime through to the frontend, including nice date formatters ("3 minutes ago")

app/desktop/studio_server/tool_api.py
```diff
@@ -296,6 +412,13 @@ async def get_tool_server(
                         ExternalToolApiDescription.tool_from_mcp_tool(tool)
                         for tool in tools_result.tools
                     ]
+            case ToolServerType.kiln_task:
+                available_tools = [
+                    await ExternalToolApiDescription.tool_from_kiln_task_tool(
+                        KilnTaskTool(project_id, tool_server.id or "", tool_server)
```
scosman: nit: can use tool_server_id instead of `tool_server.id or ""`. It can't be empty, and we don't want to return empty.

app/desktop/studio_server/tool_api.py
```diff
@@ -430,6 +553,61 @@ def _local_tool_server_properties(
             "secret_env_var_keys": tool_data.secret_env_var_keys,
         }
 
+    @app.post("/api/projects/{project_id}/add_kiln_task_tool")
```
scosman: We probably make this mistake everywhere, but we don't need the verb "add" (or "edit" below) in the URL when the API is POST (or PATCH). The API definition itself describes the action, the URI describes the resource. P2, I think we do a lot of this.

app/desktop/studio_server/tool_api.py
```diff
@@ -430,6 +553,61 @@ def _local_tool_server_properties(
...
+    ) -> ExternalToolServer:
+        existing_tool_server = tool_server_from_id(project_id, tool_server_id)
+        if existing_tool_server.type != ToolServerType.kiln_task:
+            raise HTTPException(
+                status_code=400,
+                detail="Existing tool server is not a kiln task tool. You can't edit a non-kiln task tool with this endpoint.",
+            )
+
+        # Create a deep copy of the existing tool server so if any validation fails we don't cache the bad data in memory
+        tool_server = existing_tool_server.model_copy(deep=True)
```
scosman: Oh, tool_server_from_id getting a read only copy is really dangerous. We should fix that so this isn't needed. All APIs should return mutable copies by default, unless you specifically ask for "readonly=true".

Not introduced here, but let's fix it before is causes a big bug

Just update tool_server_from_id to return `model_copy(deep=true)`, then can remove this here and above

app/desktop/studio_server/tool_api.py
```diff
@@ -430,6 +539,61 @@ def _local_tool_server_properties(
...
+            "description": tool_data.description,
+            "task_id": tool_data.task_id,
+            "run_config_id": tool_data.run_config_id,
+            "is_archived": tool_data.is_archived,
+        }
+
+        # Save the tool to file
+        tool_server.save_to_file()
+
+        return tool_server
```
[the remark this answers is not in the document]
scosman: agreed. Just fetch the task and fetch the run_config and make sure these IDs exist.

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
```diff
@@ -227,48 +241,35 @@
...
-  <div>
-    <PromptTypeSelector
-      bind:prompt_method
-      info_description="Choose a prompt. Learn more on the 'Prompts' tab."
-      bind:linked_model_selection={model}
-    />
-  </div>
   <Collapse
     title="Advanced Options"
     badge={tools.length > 0 ? "" + tools.length : null}
```
scosman: Remove the badge, since we moved tools outside of the collapse

app/web_ui/src/lib/ui/trace/trace.svelte
```diff
@@ -184,17 +252,32 @@
...
+                    />
                   </div>
                 {/if}
                 <div>
                   <div class="text-xs text-gray-500 font-bold mb-1">
                     Tool Result
+                    {#if run_link}
+                      <a
+                        href={run_link}
+                        class="text-blue-500 hover:text-blue-700 ml-2 underline"
```
scosman: - test-primary (don't use non-standard colors)
- we have a class "link" for underlining links, for our standard link style
- make this the same font and weight as the other link in this section. They should match.

app/web_ui/src/lib/ui/trace/tool_call.svelte
```diff
@@ -4,11 +4,44 @@
...
 </script>
 
 <div class="grid grid-cols-[auto,1fr] gap-x-3 gap-y-0 text-xs">
   <div class="font-medium text-gray-500">{nameTag}:</div>
-  <div class="font-mono">{tool_call.function.name}</div>
+  <div class="font-mono">
+    {#if get_tool_link()}
+      <a
+        href={get_tool_link()}
+        class="text-blue-500 hover:text-blue-700 underline"
```
scosman: - text-primary (always used named colors)
 - "link" class instead of underline (same thing basically, but if we ever want to edit will apply to all links)

app/web_ui/src/lib/ui/trace/tool_call.svelte
```diff
@@ -4,11 +4,44 @@
...
+        target="_blank"
+        rel="noopener noreferrer"
```
scosman: not needed for internal localhost links. These are SEO things

app/web_ui/src/lib/ui/trace/trace.svelte
```diff
@@ -114,6 +127,60 @@
...
+          }
+        }
+      } catch (e) {
+        // Content is not JSON, return null
+      }
+    }
+    return null
+  }
+
+  function tool_id_from_message(message: TraceMessage): string | undefined {
```
scosman: nit/P2: add tool_id to kiln_task_data_from_message above to prevent repeated parsing and repetitive code. 

Replace
```
@const run_link = get_dataset_run_link(message)
```

With
```
@const task_data = kiln_task_data_from_message(message)
```

which makes it only parse once per loop, then use the data.

app/web_ui/src/lib/ui/trace/trace.svelte
```diff
@@ -114,6 +127,60 @@
...
+          return parsed.tool_id
+        }
+      } catch (e) {
+        // Content is not JSON, return undefined
+      }
+    }
+    return undefined
+  }
+
+  function get_dataset_run_link(message: TraceMessage): string | null {
```
scosman: can delete this an just inline dataset_item_link(task_data) using the existing parsed data.

app/web_ui/src/lib/ui/fancy_select.svelte
```diff
@@ -630,8 +636,9 @@
                       </div>
                       {#if item.badge}
                         <div
-                          class="badge badge-sm px-2 {item.badge_color ===
-                          'primary'
+                          class="badge badge-sm text-xs {item.badge.length <= 2
+                            ? 'rounded-full w-5 h-5'
```
scosman, line 646: Image

app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -19,23 +19,31 @@
...
-        {#if property.link}
-          <a href={property.link} class="link">{property.value}</a>
-        {:else}
-          {property.value}
+      <div class="flex flex-col">
+        <div
+          class="flex items-start overflow-x-hidden {property.error
+            ? 'text-error'
+            : 'text-gray-500'}"
+          style="overflow-wrap: anywhere;"
```
scosman: Use class `wrap-anywhere`. There's almost always a tailwind class, and we use those. style tags usually a smell, sometimes used for dynamic values.

app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -19,23 +19,31 @@
...
-            warning_icon="exclaim"
-            warning_color="warning"
-            tight={true}
-          />
-        {/if}
-        {#if property.link}
-          <a href={property.link} class="link">{property.value}</a>
-        {:else}
-          {property.value}
+      <div class="flex flex-col">
```
scosman: Why new div and flex-col? Normally would just put it after the text (just like above one) so it's on same line.

app/web_ui/src/lib/ui/fancy_select.svelte
sfierro: discussed offline, doing in follow up

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
sfierro: good catch!

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+    ):
+        self._project_id = project_id
+        self._tool_server_model = data_model
+        self._tool_id = tool_id
+
+        self._name = data_model.properties.get("name", "")
+        self._description = data_model.properties.get("description", "")
+        self._task_id = data_model.properties.get("task_id", "")
+        self._run_config_id = data_model.properties.get("run_config_id", "")
+
```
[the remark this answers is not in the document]
sfierro: this is wrong

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task_tools/+page.svelte
```diff
@@ -0,0 +1,169 @@
...
+  function navigateToToolDetails(tool: KilnTaskToolDescription) {
+    goto(
+      `/settings/manage_tools/${project_id}/kiln_task/${tool.tool_server_id}`,
+    )
+  }
+</script>
+
+<div class="max-w-[1400px]">
+  <AppPage
+    title="Kiln Task Tools"
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task_tools/+page.svelte]: "Manage Kiln Tasks as Tools"?

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task/[tool_server_id]/+page.svelte
```diff
@@ -0,0 +1,442 @@
...
+        warning_message={unarchive_error.getMessage() ||
+          "An unknown error occurred"}
+        large_icon={true}
+        warning_color="error"
+        outline={true}
+      />
+    {/if}
+    {#if is_archived}
+      <Warning
+        warning_message="This Kiln Task tool is archived. You may unarchive it to use it again."
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task/[tool_server_id]/+page.svelte]: Task doesn't need to be capitalized

app/web_ui/src/lib/ui/property_list.svelte
sfierro: I think this was leftover code, info description is unused. reverting

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -208,7 +244,56 @@
...
               </tr>
             {/if}
-            {#each tools || [] as tool}
+            {#if kiln_task_tools && kiln_task_tools.length > 0 && unarchived_kiln_task_tools}
+              <tr
+                class="hover:bg-base-200 cursor-pointer"
+                on:click={() =>
+                  goto(`/settings/manage_tools/${project_id}/kiln_task_tools`)}
+                on:keydown={(e) =>
+                  e.key === "enter" &&
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte]: Not working. I think "Enter" should be capitalized.

Fix the RAG one above too please!

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -90,14 +90,18 @@ async def invoke(
         self,
         input: Dict | str,
         input_source: DataSource | None = None,
+        output_source_type: DataSourceType | None = None,
```
scosman: I think we can revert this whole file? We went back to output always synthetic, which was a good call.

Since this is the base_adapter (calling a LLM source) it's reasonable to enforce that output=synthetic and not take it as an option.

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -609,7 +609,13 @@ async def process_tool_calls(
                     f"Failed to validate arguments for tool '{tool_name}'. The arguments didn't match the tool's schema. The arguments were: {parsed_args}\n The error was: {e}"
                 ) from e
 
-            result = await tool.run(**parsed_args)
+            # Create context with the calling task's allow_saving setting
+            from kiln_ai.tools.base_tool import ToolCallContext
```
scosman: imports should be at top of the file. Might need to move ToolCallContext class into it's own file to prevent circular dep, but that's a good thing.

libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -1,10 +1,18 @@
...
 
 from kiln_ai.datamodel.json_schema import validate_schema_dict
 from kiln_ai.datamodel.tool_id import KilnBuiltInToolId, ToolId
 
 
+@dataclass
+class ToolCallContext:
+    """Context passed to tools when they are called, containing information from the calling task."""
+
+    allow_saving: bool = True
```
scosman, line 14: might want to document what this is for here. Most tools (MCP, etc) don't have this concept and the name isn't obvious.

libs/core/kiln_ai/datamodel/external_tool_server.py
```diff
@@ -231,6 +246,25 @@ def validate_required_fields(cls, data: dict) -> dict:
...
+            case ToolServerType.kiln_task:
+                tool_name_validator(properties.get("name", ""))
+                err_msg_prefix = "Kiln task server properties:"
+                validate_return_dict_prop(
+                    properties, "description", str, err_msg_prefix
+                )
+                description = properties.get("description", "")
+                if len(description) > 128:
+                    raise ValueError(
+                        "description must be less than 128 characters long"
```
scosman: 128 or fewer, not less than 128 (128 is valid).

Leonard made the same comment on my PR so you get it here :P

libs/core/kiln_ai/datamodel/task_output.py
```diff
@@ -189,64 +190,88 @@ class DataSourceProperty(BaseModel):
...
+                DataSourceType.file_import,
+                DataSourceType.tool_call,
+            ],
         ),
         DataSourceProperty(
             name="model_name",
             type=str,
-            required_for=[DataSourceType.synthetic],
-            not_allowed_for=[DataSourceType.human, DataSourceType.file_import],
+            required_for=[DataSourceType.synthetic, DataSourceType.tool_call],
```
scosman: I thought we were removing these? Tool_call mode is only used for input, so doesn't have model_name/provider/adapter.

Tool call should be not allowed for them all no?

libs/server/kiln_server/task_api.py
```diff
@@ -95,6 +99,30 @@ async def delete_task(project_id: str, task_id: str) -> None:
         task = task_from_id(project_id, task_id)
         task.delete()
 
+        # Archive any kiln task tools that have this task set as their task_id
+        parent_project = task.parent_project()
+        if parent_project is not None:
+            for tool_server in parent_project.external_tool_servers(readonly=True):
```
scosman: +1 to the other comment. Not readonly.

libs/server/kiln_server/task_api.py
```diff
@@ -95,6 +99,30 @@ async def delete_task(project_id: str, task_id: str) -> None:
...
+                if (
+                    tool_server.type == ToolServerType.kiln_task
+                    and tool_server.properties.get("task_id") == task_id
+                ):
+                    # For kiln task tools, we know the properties are KilnTaskServerProperties
+                    # Create a copy and update the is_archived flag
+                    updated_properties: KilnTaskServerProperties = {
```
scosman: Bunch of code, and its risky. We're still working with untyped accessors (get), and if we ever add a property we'd need to remember to update this code (I would forget).

Let's just do this since is_archived isn't optional?

```
if (
    tool_server.type == ToolServerType.kiln_task
    and tool_server.properties.get("task_id") == task_id
):
    if "is_archived" in tool_server.properties:
        tool_server.properties["is_archived"] = True
    else:
        raise TypeError("expected archiveable tool task server")
    tool_server.save_to_file()
```

I don't love the type ignore, but in this case it's safer (we just checked type, new properties won't be lost). I think Pydantic classes might be able to solve this for us over typedDict.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+        self,
+        project_id: str,
+        tool_id: str,
+        data_model: ExternalToolServer,
+    ):
+        self._project_id = project_id
+        self._tool_server_model = data_model
+        self._tool_id = tool_id
+
+        self._name = data_model.properties.get("name", "")
```
scosman, line 37: @aiyakitori how do we get the benefits of the enw typing? Shouldn't need `|| ""` It's getting treated like a normal dict

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+        return {
+            "type": "function",
+            "function": {
+                "name": await self.name(),
+                "description": await self.description(),
+                "parameters": await self.get_parameters_schema(),
+            },
+        }
+
+    async def run(self, **kwargs) -> str:
```
scosman: should we just delete this from the interface, and update the 3 other implementations? 

Can be a cleanup PR.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+            {
+                "output": task_run.output.output,
+                "task_run_id": task_run.id,
+                "task_id": task_run.parent.id if task_run.parent else None,
+                "project_id": self._project_id,
+                "tool_id": self._tool_id,
+            }
+        )
+
+    async def _get_task(self) -> Task:
```
scosman: Check out `@cached_property`. Easy lazy loader. Don't need to define it in __init__ either

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+        # Determine the input format
+        if task.input_json_schema:
+            # Structured input - pass kwargs directly
+            input = kwargs
+        else:
+            # Plaintext input - extract from 'input' parameter or convert kwargs
+            if "input" in kwargs:
+                input = kwargs["input"]
+            else:
+                # Convert kwargs to a descriptive string
```
scosman: just raise here? the tool schema requires input. This isn't valid.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                input = json.dumps(kwargs, indent=2)
+
+        # These imports are here to avoid circular chains
+        from kiln_ai.adapters.adapter_registry import adapter_for_task
```
scosman, line 81: imports should be at the top. Does it create circular?

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                allow_saving=context.allow_saving,
+                default_tags=["tool_call"],
+            ),
+        )
+        task_run = await adapter.invoke(
+            input,
+            input_source=DataSource(
+                type=DataSourceType.tool_call,
+                properties={
+                    "model_name": run_config.run_config_properties.model_name,
```
scosman: Remove all the properties. These properties all relate to the model/provider/adapter from the run_config, which is how we generate the output. input_source is supposed to describe where the input came from (not output). These probably are not correct for the input, which could come from any model.

just type=tool_call is enough. You can lookup that tool call and get it's output properties if you want the correct value where.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                    "model_provider": run_config.run_config_properties.model_provider_name,
+                    "adapter_name": adapter.adapter_name(),
+                },
+                run_config=run_config.run_config_properties,
+            ),
+            output_source_type=DataSourceType.tool_call,
```
scosman: remove this. It should still by synthetic. The output is synthetic (a LLM is generating it), it's just the input which is a tool call.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+
+            # Load the task from the project
+            self._task = Task.from_id_and_parent_path(self._task_id, project.path)
+            if self._task is None:
+                raise ValueError(
+                    f"Task not found: {self._task_id} in project {self._project_id}"
+                )
+        return self._task
+
+    async def _get_run_config(self) -> TaskRunConfig:
```
scosman: ditto for cached_property

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                ),
+                None,
+            )
+            if self._run_config is None:
+                raise ValueError(
+                    f"Task run config not found: {self._run_config_id} for task {self._task_id} in project {self._project_id}"
+                )
+        return self._run_config
+
+    async def get_parameters_schema(self) -> Dict[str, Any]:
```
scosman: ditto for cached_property

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                # Use the task's input schema directly if it exists
+                self._parameters_schema = task.input_schema()
+            else:
+                # For plaintext tasks, create a simple string input parameter
+                self._parameters_schema = {
+                    "type": "object",
+                    "properties": {
+                        "input": {
+                            "type": "string",
+                            "description": f"Plaintext input for the task: {task.instruction}",
```
scosman: Critical: task.instructions could be 5000 words. Way too long and unbounded for a tool description. It's also a prompt, which isn't the right thing to put in a description. 

Just "input to the tool" is fine here I think (task is Kiln specific word, model sees tools). The task description can be tuned, and this should work as well as any general purpose string.

scosman (changes requested):

libs/core/kiln_ai/tools/kiln_task_tool.py
scosman: This also makes the tool mutable, whenever you change instruction, the tool itself changes. Kills the benefits of locking the run config.

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,183 @@
...
+                    "model_provider": run_config.run_config_properties.model_provider_name,
+                    "adapter_name": adapter.adapter_name(),
+                },
+                run_config=run_config.run_config_properties,
+            ),
+            output_source_type=DataSourceType.tool_call,
+        )
+
+        # Return structured information about the created task as tool run
+        return json.dumps(
```
scosman: P0: you can't stuff trace messages like this. The content field needs to be the real content and only the content (`task_run.output.output`). This will break evals, fine tuning... kinda everything. We use it all over the place, and it's a well know schema outside of just Kiln (OpenAI defined trace schema).

You need to make a custom class, similar to ChatCompletionToolMessageParam in open_ai_types.py like we did for ChatCompletionAssistantMessageParamWrapper. You can add custom fields there extending the spec, but can't redefine the content of well known fields.

Will probably require changing the return type too (str, ToolResponseProperties)

I'd probably put them into 1 composite field: "kiln_tool_run": "project_id::task_id::run_id"

scosman (changes requested):

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
sfierro: yep already did but my commit never pushed

libs/core/kiln_ai/tools/kiln_task_tool.py
sfierro: good catch!

libs/core/kiln_ai/tools/kiln_task_tool.py
aiyakitori: Typed dict is still a dict so it's dictionary style access. there is no run time checking, the benefit is pyright static checks and IDE hints mostly. There is no run time checks. The real strong typed would be pydantic models but you lose the dictionary flexibility on "properties". 

Making both local/remote server properties pydantic model would work but that's a very big change vs typed dict that buys a bit more convenience without too much effort.

app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte
```diff
@@ -208,7 +244,56 @@
...
+                  {:else}
+                    {unarchived_kiln_task_tools.length} Kiln task tools available
+                  {/if}
+                  {#if unarchived_kiln_task_tools.length > 0}
+                    <InfoTooltip
+                      tooltip_text="Tools: {unarchived_kiln_task_tools
+                        .map((tool) => tool.tool_name)
+                        .join(', ')}"
+                      no_pad={true}
+                    />
```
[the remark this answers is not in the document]
Comment (sfierro) [reply app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/+page.svelte:282]: this is wrong

libs/core/kiln_ai/tools/kiln_task_tool.py
sfierro: filed a Linear issue

libs/core/kiln_ai/tools/kiln_task_tool.py
sfierro: added a follow up task

libs/core/kiln_ai/tools/kiln_task_tool.py
```diff
@@ -0,0 +1,160 @@
...
+from kiln_ai.datamodel import Task
+from kiln_ai.datamodel.external_tool_server import ExternalToolServer
+from kiln_ai.datamodel.task import TaskRunConfig
+from kiln_ai.datamodel.task_output import DataSource, DataSourceType
+from kiln_ai.datamodel.tool_id import ToolId
+from kiln_ai.tools.base_tool import KilnToolInterface, ToolCallContext
+from kiln_ai.utils.project_utils import project_from_id
+
+
+class KilnTaskToolResult:
```
scosman, line 15: Check out  Python`@dataclass`. Let's you skip writing an init method. Nothing wrong with this, just good to know.

libs/core/kiln_ai/utils/open_ai_types.py
```diff
@@ -84,11 +84,28 @@ class ChatCompletionAssistantMessageParamWrapper(TypedDict, total=False):
...
+    role: Required[Literal["tool"]]
+    """The role of the messages author, in this case `tool`."""
+
+    tool_call_id: Required[str]
+    """Tool call that this message is responding to."""
+
+    kiln_task_tool_data: Optional[str]
+    """The data for the Kiln task tool that this message is responding to.
+    
+    Formatted as `<project_id>,<tool_id>,<task_id>,<run_id>`
```
scosman: We use `::` delimiters. 

Good to follow the convention, but more importantly commas might be allowed in IDs so might be serious (not positive but let's follow convention anyways and save 15 seconds checking 😆 ). I know we allow a bunch of chars even though they are usually numbers.

scosman (changes requested):
one small important one, but then good to go!

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -2241,45 +2248,232 @@ class KilnModel(BaseModel):
...
+                multimodal_mime_types=[
+                    KilnMimeType.JPG,
+                    KilnMimeType.PNG,
+                    KilnMimeType.PDF,
+                    KilnMimeType.TXT,
+                    KilnMimeType.MD,
+                ],
+                multimodal_requires_pdf_as_image=True,
+            ),
+            KilnModelProvider(
```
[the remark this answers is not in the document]
leonardmq: Just double checked again and the slugs really are this - the `7b` inconsistency is also correct:
```
qwen/qwen-2.5-vl-7b-instruct
qwen/qwen2.5-vl-32b-instruct
qwen/qwen2.5-vl-72b-instruct
qwen/qwen3-vl-235b-a22b-instruct
```

List is here: https://openrouter.ai/models?q=qwen%20vl

PR #676, opened 2025-10-01
feat: add more providers for Qwen3 VL and Qwen2.5 VL

leonardmq:
…en2.5 vl

## What does this PR do?

Add mode providers for Qwen2.5 VL (SiliconFlow, Together, Fireworks) and add Qwen3 VL on OpenRouter. All with the new pipeline to flatten PDFs into images - otherwise it does not work.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/ml_model_list.py
+++ b/libs/core/kiln_ai/adapters/ml_model_list.py
@@ -170,6 +170,7 @@ class ModelName(str, Enum):
     qwen_3_235b_a22b = "qwen_3_235b_a22b"
     qwen_3_235b_a22b_2507_no_thinking = "qwen_3_235b_a22b_2507_no_thinking"
     qwen_3_235b_a22b_no_thinking = "qwen_3_235b_a22b_no_thinking"
+    qwen_3_vl_235b_a22b_no_thinking = "qwen_3_vl_235b_a22b_no_thinking"
     qwen_long_l1_32b = "qwen_long_l1_32b"
     kimi_k2 = "kimi_k2"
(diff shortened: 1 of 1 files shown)


libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -2363,8 +2364,38 @@ class KilnModel(BaseModel):
                 multimodal_requires_pdf_as_image=True,
                 max_parallel_requests=1,
             ),
-            # openrouter has the model, but annotation format returned by this model is rejected
-            # by litellm internal validation
+            KilnModelProvider(
+                name=ModelProviderName.openrouter,
```
scosman, line 2368: These two are 7b version of model, but container is for 3B

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3701,6 +3825,34 @@ class KilnModel(BaseModel):
             ),
         ],
     ),
+    # Qwen 3 235B (22B Active) VL Non-Thinking
+    KilnModel(
+        family=ModelFamily.qwen,
+        name=ModelName.qwen_3_vl_235b_a22b_no_thinking,
+        friendly_name="Qwen 3 235B (22B Active) VL Non-Thinking",
```
scosman, line 3832: "Qwen 3 VL 235B (22B Active) Instruct"

Important to move the VL up, it gets lost after the rest, and there are a bunch of other Qwen 3 models. "Instruct" to match their official name (prefer those when we can, "Non Thinking" more for a if a model has 2 modes and we want to differentiate)

PR #677, opened 2025-10-01
Capture MCP server stderr

aiyakitori:
## What does this PR do?

Capture MCP server's stderr to provide more accurate error message to users when connecting to local mcp server fails. 

* Use tmp file for stdio client
* Log the error to Kiln log archive
* Update error UI for local mcp, add trouble shooting steps. 

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Added a detailed error panel for local MCP connection failures with troubleshooting steps, displayed command/arguments, error details, and a "View Logs" action.

* **Bug Fixes**
  * Captures and surfaces MCP server stderr in error messages to provide fuller diagnostic details for local connection attempts (remote behavior unchanged).

* **Chores**
  * Centralized "View Logs" behavior so Application Logs consistently opens the logs.

* **Tests**
  * Updated tests to expect the centralized error instruction and new error output.

### DIFF
--- a/libs/core/kiln_ai/tools/mcp_session_manager.py
+++ b/libs/core/kiln_ai/tools/mcp_session_manager.py
@@ -172,11 +172,10 @@ class MCPSessionManager:
         )
 
         # Create temporary file to capture MCP server stderr
-        err_log_fd, err_log_path = tempfile.mkstemp(suffix="_mcp_stderr.log")
-        err_log_file = open(err_log_fd, "w+")
+        err_log = tempfile.TemporaryFile(mode="w+", encoding="utf-8")
 
         try:
-            async with stdio_client(server_params, errlog=err_log_file) as (
+            async with stdio_client(server_params, errlog=err_log) as (
                 read,
                 write,
             ):
@@ -187,7 +186,8 @@ class MCPSessionManager:
                     yield session
         except Exception as e:
             # Read stderr content from temporary file for debugging
-            stderr_content = self._read_tmp_file(err_log_fd, err_log_path)
+            err_log.seek(0)  # Read from the start of the file
+            stderr_content = err_log.read()
             if stderr_content:
                 logger.error(
                     f"MCP server '{tool_server.name}' stderr output: {stderr_content}"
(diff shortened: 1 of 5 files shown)


libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -170,29 +173,51 @@ async def _create_local_mcp_session(
             command=command, args=args, env=env_vars, cwd=cwd
         )
 
+        # Create temporary file to capture MCP server stderr
+        err_log = tempfile.TemporaryFile(mode="w+", encoding="utf-8")
```
leonardmq: You can probably use the temporary file context manager to avoid having to create and close the temp file manually:
```py
with tempfile.NamedTemporaryFile(delete=True) as err_log:
    # file writes / read

# file is closed and deleted automatically when the block exits
```

scosman:
Visual update. Error details truncates height for long messages.

Screenshot 2025-10-02 at 12 27 55 PM

libs/core/kiln_ai/tools/mcp_session_manager.py
aiyakitori: This is good suggestion but sadly we can't use this here since the file would be closed before the session was yield unless we refactor and place it at the top level. 

"with doesn't work here - the file would close when the with block exits, but you need it open during the yield (when the caller uses the session)"

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -170,29 +174,51 @@ async def _create_local_mcp_session(
...
             if mcp_error and isinstance(mcp_error, McpError):
-                self._raise_local_mcp_error(mcp_error)
+                self._raise_local_mcp_error(mcp_error, stderr_content)
 
             # Re-raise the original error but with a friendlier message
-            self._raise_local_mcp_error(e)
+            self._raise_local_mcp_error(e, stderr_content)
+        finally:
+            # Close and delete the temp file
+            err_log.close()
```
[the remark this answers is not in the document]
aiyakitori: Fixed

PR #680, opened 2025-10-02
feat: more doc extractors and pdf to image pipeline

leonardmq:
## What does this PR do?

Add new extractor models across different providers. As some of the new extractor models (Qwen VL) do not support PDF natively, they first need to be converted into images first - we use `pypdfium2` for this - we do the normal page-by-page pipeline, but convert the page into an image before sending it to the model based on a new model provider property named `multimodal_requires_pdf_as_image`.

Also add a template for all local extraction + embedding with Ollama.

## Checklists

## Summary by CodeRabbit

- **New Features**
  - Provider-based setup for search tool templates (OpenAI, Google Gemini, Ollama) and unified provider checks.
  - Ollama flow with a dedicated dialog showing required local models/commands and guided navigation.
  - Added “Local Qwen” template for Ollama/local usage.
  - PDFs can be processed as images when a model requires image input.
  - Expanded model catalog with multiple Qwen VL variants (including a non-thinking option).

- **Tests**
  - Added tests for PDF→image conversion, per-page image-based extraction, caching, and error scenarios.

- **Chores**
  - Added runtime dependencies for PDF rendering and image processing.

### DIFF
--- a/app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte
+++ b/app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte
@@ -11,7 +11,7 @@
   import { progress_ui_state } from "$lib/stores/progress_ui_store"
   import {
     rag_config_templates,
-    type RequiredApiKeysSets,
+    type RequiredProvider,
     type RagConfigTemplate,
   } from "./rag_config_templates"
   import Warning from "$lib/ui/warning.svelte"
@@ -19,7 +19,10 @@
 
   $: project_id = $page.params.project_id
 
+  let selected_template_id: string | null = null
+  let selected_template: RagConfigTemplate | null = null
   let requires_api_keys_dialog: Dialog | null = null
+  let requires_ollama_dialog: Dialog | null = null
 
   onMount(() => {
     load_settings()
@@ -31,18 +34,30 @@
       subtitle: template.preview_subtitle,
       description: template.preview_description,
       tooltip: template.preview_tooltip,
-      on_click: () => suggestion_selected(template, id),
-      required_api_keys: template.required_api_keys,
+      on_click: () => {
+        selected_template_id = id
+        selected_template = template
+        suggestion_selected(template, id)
+      },
     }),
   )
 
-  let missing_api_keys: RequiredApiKeysSets | null = null
-  $: missing_api_keys_string = missing_api_keys
+  let missing_provider: RequiredProvider | null = null
+  $: missing_provider_string = missing_provider
     ? {
         Openai: "OpenAI",
         Gemini: "Google Gemini",
-      }[missing_api_keys]
+        Ollama: "Ollama",
+      }[missing_provider]
     : null
+
+  function redirect_to_template(template_id: string) {
+    // Go to the create search tool page with the template id
+    goto(
+      `/docs/rag_configs/${project_id}/create_rag_config?template_id=${template_id}`,
+    )
+  }
+
   function suggestion_selected(
     suggestion: RagConfigTemplate,
     template_id: string,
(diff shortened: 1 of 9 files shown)


libs/core/kiln_ai/adapters/extractors/litellm_extractor.py
```diff
@@ -197,11 +219,6 @@ async def _extract_single_pdf_page(
             )
 
         content = response.choices[0].message.content
-        if not content:
```
leonardmq, line 200: The idea here originally was that if we were getting an empty string, it was probably due to an error, but in practice, it is not necessarily the case:
1. Small models (e.g. Qwen 2.5 VL 7B) sometimes just sends back an empty string when the page has almost nothing on it - it is legitimate model behavior.
2. Sometimes the page legitimately has nothing on it, and if we fail the page, the whole PDF fails - so that prevents ever processing a file that contains white pages in it.

So I removed this check and we keep `""` as a valid extraction output

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte:178]: testing comment outside the diff

libs/core/kiln_ai/utils/test_pdf_utils.py
```diff
@@ -71,3 +74,13 @@ async def test_split_pdf_into_pages_temporary_directory_creation(mock_file_facto
     # Verify the temporary directory is cleaned up
     for temp_dir in captured_temp_dirs:
         assert not temp_dir.exists()
+
+
+async def test_convert_pdf_to_images(mock_file_factory):
```
scosman, line 79: was worried about perf, but fast!

libs/core/kiln_ai/utils/pdf_utils.py
```diff
@@ -36,3 +37,22 @@ async def split_pdf_into_pages(pdf_path: Path) -> AsyncGenerator[list[Path], Non
                 page_paths.append(page_path)
 
         yield page_paths
+
+
+def convert_pdf_to_images(pdf_path: Path, output_dir: Path) -> list[Path]:
+    image_paths = []
+
+    # note: doing this in a thread causes a segfault - but this is slow and blocking
```
scosman, line 45: hmmm, agreed. Mind trying to find an option? Blocking main run loop is bad.

Option 1:
 - Can we try the `import pypdfium2` inside the threaded method. Might fix it, by giving each thread it's own library instance.
 - Try something like `multiprocessing.set_start_method("spawn", force=True)` ChatGPT suggestion on lib, my idea to just spawn process. We just need to pass a string path and get a string path back so should be easy.

libs/core/kiln_ai/utils/pdf_utils.py
```diff
@@ -36,3 +37,22 @@ async def split_pdf_into_pages(pdf_path: Path) -> AsyncGenerator[list[Path], Non
...
+    # so we should try to find a better way
+    pdf = pypdfium2.PdfDocument(pdf_path)
+    try:
+        for idx, page in enumerate(pdf):
```
scosman, line 49: add an `await sleep(0)` at minimum

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3587,6 +3825,34 @@ class KilnModel(BaseModel):
             ),
         ],
     ),
+    # Qwen 3 235B (22B Active) VL Non-Thinking
+    KilnModel(
+        family=ModelFamily.qwen,
+        name=ModelName.qwen_3_vl_235b_a22b_no_thinking,
+        friendly_name="Qwen 3 235B (22B Active) VL Non-Thinking",
```
scosman: "Qwen 3 VL Instruct (Vision-Language) - 235B (22B Active)"

Important to move the VL up, it gets lost after the size, and there are a bunch of other Qwen 3 models. "Instruct" to match their official name (prefer those when we can, "Non Thinking" more for a if a model has 2 modes and we want to differentiate)

app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte
```diff
@@ -62,35 +77,45 @@
...
+    // we do not have an easy way to check if Ollama and the required models are installed
+    // so we always show the dialog
+    if (suggestion.required_provider === "Ollama") {
+      missing_provider = "Ollama"
+      requires_ollama_dialog?.show()
+      posthog.capture("missing_api_keys_for_search_tool", {
+        template_id,
+      })
+      return
+    }
```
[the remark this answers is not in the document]
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/add_search_tool/+page.svelte:112]: For backward compatibility, probably best to keep the old one?

libs/core/kiln_ai/utils/pdf_utils.py
leonardmq: Added in https://github.com/Kiln-AI/Kiln/pull/680/commits/3e0e3e59f76cff13e3194a733cbdbb1f0125ce7a

Currently it does not have an impact because we only convert one page at a time. 

Later on I would also like combine the PDF splitting and the image into one context manager (with an arg to choose whether to yield each page as a PDF or as an image)

scosman (approved):

libs/core/kiln_ai/utils/pdf_utils.py
leonardmq: The import inside the method solves the segfault, but we get other issues when running a full batch of docs:
- multiprocessing -> `too many open files`
- thread -> at some point starts hanging

We probably need to semaphore the number of threads / processes we can spawn from this - `asyncio` thread seems to have a built-in default based on the number of cores, but maybe that is too high. 

I have added the `asyncio.sleep(0)` between each page (and we only ever convert one page at a time), it should not block too much and avoids the risk of exhausting resources - while processing a bunch of PDFs, the dashboard / server is still responsive. Will try to find a way later (might overlap with the broader task of improving the rate limit / concurrency management)

libs/core/kiln_ai/adapters/ml_model_list.py
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/680/commits/0ba238b7f1b7934587054993c49cc1f733641cce

PR #681, opened 2025-10-03
Add simple session caching for remote mcp server

aiyakitori:
## What does this PR do?

Add simple session caching for remote mcp server. This would prevent spawning new processes for every tool call. 

* Caching is only done for local mcp, remote mcp via http is ephemeral. 
* Currently we don't support TTL (time-to-live) cache or other self clean up mechanisms. Clients are required to manually clean up the sessions. It's done at the following places
   * app shutdown
   * delete / edit tool server

## Checklists

### DIFF
--- a/app/desktop/studio_server/tool_api.py
+++ b/app/desktop/studio_server/tool_api.py
@@ -356,6 +356,9 @@ def connect_tool_servers_api(app: FastAPI):
         # Validate the tool server connectivity
         await validate_tool_server_connectivity(existing_tool_server)
 
+        # Remove cached session if any
+        await MCPSessionManager.shared().invalidate_session(tool_server_id)
+
         # Save the tool to file
         existing_tool_server.save_to_file()
 
@@ -415,6 +418,9 @@ def connect_tool_servers_api(app: FastAPI):
         MCPSessionManager.shared().clear_shell_path_cache()
         await validate_tool_server_connectivity(tool_server)
 
+        # Remove cached session if any
+        await MCPSessionManager.shared().invalidate_session(tool_server_id)
+
         # Save the tool to file
         tool_server.save_to_file()
 
@@ -433,6 +439,8 @@ def connect_tool_servers_api(app: FastAPI):
     @app.delete("/api/projects/{project_id}/tool_servers/{tool_server_id}")
     async def delete_tool_server(project_id: str, tool_server_id: str):
         tool_server = tool_server_from_id(project_id, tool_server_id)
+        # Close and remove the cached session
+        await MCPSessionManager.shared().invalidate_session(tool_server_id)
         # Delete the secrets from the settings
         tool_server.delete_secrets()
         # Delete the tool server from the file system
--- a/libs/core/kiln_ai/tools/mcp_session_cache.py
+++ b/libs/core/kiln_ai/tools/mcp_session_cache.py
@@ -1,0 +1,63 @@
+import asyncio
+
+from mcp.client.session import ClientSession
+
+from kiln_ai.utils.logging import logging
+
+logger = logging.getLogger(__name__)
+
+
+class MCPSessionCache:
+    """
+    Simple cache for MCP ClientSession by ExternalToolServer ID.
+
+    This cache provides manual lifecycle management - sessions must be
+    explicitly closed via close_session() or close_all(). Sessions remain alive until explicitly
+    removed or cleared.
+    """
+
+    def __init__(self):
+        self._sessions: dict[str, ClientSession] = {}
+        self._lock = asyncio.Lock()
+
+    async def _cleanup_session(self, session: ClientSession) -> None:
+        """Helper to clean up a session"""
+        try:
+            await session.__aexit__(None, None, None)
+        except Exception as e:
+            # Ignore errors during cleanup so we don't cause MCPSessionManager to throw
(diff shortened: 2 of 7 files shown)


libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -50,9 +52,21 @@ async def mcp_client(
...
+                # Instead, the session is persisted in the MCPSessionCache and requires explicit invalidation via invalidate_session().
+                if tool_server.id is not None and (
+                    cached := await self._session_cache.get(tool_server.id)
+                ):
+                    yield cached
+                    return
+
                 async with self._create_local_mcp_session(tool_server) as session:
+                    if tool_server.id is not None:
+                        await self._session_cache.set(tool_server.id, session)
```
aiyakitori, line 69: @scosman i wonder if we should only cache the tool server when we are doing tool calls? i.e. there is really no point on keeping those sessions alive if all the app is doing is navigate through settings and browsing the available tools via `list_tools()`?

libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -50,10 +52,24 @@ async def mcp_client(
...
+                    yield cached
+                    return
+
+                # For new sessions that will be cached, manually enter context without auto-exiting
+                # This prevents the session from being closed when context manager exits
+                context = self._create_local_mcp_session(tool_server)
+                session = await context.__aenter__()
+                if tool_server.id is not None:
+                    cached_session = CachedSession(session=session, context=context)
+                    await self._session_cache.set(tool_server.id, cached_session)
```
aiyakitori, line 71: @scosman  i wonder if we should only cache the tool server when we are doing tool calls? i.e. there is really no point on keeping those sessions alive if all the app is doing is navigate through settings and browsing the available tools via list_tools()? Although we can now display the content instantly on 2nd visit, no spinners.

libs/core/kiln_ai/tools/mcp_session_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    async def set(self, server_id: str, session: ClientSession) -> None:
+        # Close old session if it exists to prevent memory leak
+        async with self._lock:
+            old_session = self._sessions.get(server_id)
+            self._sessions[server_id] = session
+
+        # Clean up old session, no lock needed
+        if old_session is not None:
+            await self._cleanup_session(old_session)
+
```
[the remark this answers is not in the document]
aiyakitori: How does this work with MCP SDK's studio_client? Read `_create_local_mcp_session`, 

```
                context = self._create_local_mcp_session(tool_server)
                session = await context.__aenter__()
```
When the cache calls context.__aexit__(), it unwinds these nested contexts, which:
Exits the ClientSession context
Exits the stdio_client context ← this closes the subprocess/transport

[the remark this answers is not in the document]
aiyakitori:
```
    async def _cleanup_session(self, session: CachedSession) -> None:
        """Helper to clean up a session"""
        try:
            await session.context.__aexit__(None, None, None)
        except Exception as e:
            # Ignore errors during cleanup so we don't cause MCPSessionManager to throw
            logger.error(f"Error cleaning up ClientSession: {e}")
            pass
```
This is how the session was cleaned up. We first __aexit__() on the "context", and remove the reference to MCP's ClientSession from the cache for garbage clean up. In fact, this is way we cache the "context", otherwise when we read the cache again, the context would've been garbage collected when mcp_client() exits, leaving a ClientSession object without the underlying context. 

Double check if my statements were correct.

libs/core/kiln_ai/tools/mcp_session_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    async def close_session(self, server_id: str) -> None:
+        """
+        Close and remove a specific session from the cache.
+        """
+        async with self._lock:
+            session = self._sessions.pop(server_id, None)
+
+        if session is not None:
+            await self._cleanup_session(session)
+
```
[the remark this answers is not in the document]
aiyakitori: https://github.com/Kiln-AI/Kiln/pull/681#discussion_r2401351380

Check other thread, not needed anymore. Update other comments in the PR if they are stale.

app/desktop/studio_server/tool_api.py
```diff
@@ -356,6 +356,9 @@ async def edit_remote_mcp(
         # Validate the tool server connectivity
         await validate_tool_server_connectivity(existing_tool_server)
 
+        # Remove cached session if any
+        await MCPSessionManager.shared().invalidate_session(tool_server_id)
```
aiyakitori, line 360: RemoteMCP currently doesn't use cache but I'm afraid in the future and forget to invalidate it? Or we could remove this....

sfierro: keep it

libs/core/kiln_ai/tools/mcp_session_manager.py
sfierro: im a big fan of no spinners

scosman:
@chiang-daniel let's close this? I made a "session based" version here: https://github.com/Kiln-AI/Kiln/pull/1015/changes

We want session isolation so general caching isn't ideal.

PR #700, opened 2025-10-06
fix: kiln tool interface run method change

leonardmq:
## What does this PR do?

Fix: broken `RagTool` implementation due to interface change that (silently) broke this class' `run` method. Details below - might want to figure out a way to trigger a typing error if the method prototype changes.

Compiler does not catch the invalid implementation / override because it is in `basic` mode (but `strict` causes a billion typing errors). The `"reportIncompatibleMethodOverride": "error",` option can be used, but causes some other errors elsewhere. Will do in another PR.

---

The `run` method in the `KilnToolInterface` changed from:
```py
@abstractmethod
async def run(self, **kwargs) -> Any:
    """Execute the tool with the given parameters."""
    pass
```

to:
```py
@abstractmethod
async def run(self, context: ToolCallContext | None = None, **kwargs) -> Any:
    """Execute the tool with the given parameters and calling context if provided."""
    pass
```

The `RagTool(KilnToolInterface)` implemented the interface as follows - and was not updated after the interface change:
```py
async def run(self, query: str) -> str:
    ...
```

This caused the function to be called with `RagTool.run(context=None, query=xxx)`, which then caused:
```
TypeError: RagTool.run() got multiple values for argument 'query'
```

## Checklists

## Summary by CodeRabbit

* **New Features**
  * RAG tool supports optional context-aware execution for richer, tailored results.
  * Returns consolidated plain-text search results for easier consumption.

* **Refactor**
  * RAG tool API updated: callers should provide query as a named argument; positional-only usage may be impacted.
  * Improved input validation with clearer errors when query is missing.

* **Tests**
  * Added tests covering context handling and query validation.

### DIFF
--- a/libs/core/kiln_ai/tools/rag_tools.py
+++ b/libs/core/kiln_ai/tools/rag_tools.py
@@ -1,5 +1,5 @@
 from functools import cached_property
-from typing import Any, Dict, List
+from typing import Any, Dict, List, TypedDict
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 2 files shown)


leonardmq:
@coderabbitai review

scosman (approved):

PR #701, opened 2025-10-06
fix: typing should catch invalid overrides

leonardmq:
## What does this PR do?

This forks off https://github.com/Kiln-AI/Kiln/pull/700 to add the permanent typechecking fix and fix resulting errors.

Change typechecking config to add:
```
# flag invalid implementation of base abstract methods
"reportIncompatibleMethodOverride": "error"
```

This typechecking flags the math tools as being invalid typing (they still work); I updated that code to pass. Not a big fan of needing to unwrap `kwargs` manually inside the method body, but that seems required to pass the typecheck (might be another way), and is better than risking silent interface drift.

## Checklists


leonardmq:
@coderabbitai review

PR #703, opened 2025-10-07
Remove eval run config default

sfierro:


app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_configs/+page.svelte
```diff
@@ -547,7 +511,7 @@
               {eval_id}
               {current_eval_config_id}
               run_all={true}
-              btn_primary={!focus_select_eval_config}
+              btn_primary={!focus_run_all}
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/compare_run_configs/+page.svelte]: naming seems backwards. we focus it when it's true, but named the inverse. Logic might be okay with ! but confusing. Also: maybe just inline the tiny bit of logic

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/+page.svelte
```diff
@@ -224,14 +191,14 @@
     "Create Eval Data",
     "Human Ratings",
     "Find the Best Judge",
-    "Find the Best Way to Run this Task",
+    "Evaluate Run Configurations",
```
Comment (scosman) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/+page.svelte]: Old title was better. Yours is "what" but 99.5% of users won't know what a "Run Configuration" is. Yes title is "why" in old format, but that's more compelling.

PR #708, opened 2025-10-08
fix: API binding check CI failing on 3.14

leonardmq:
## What does this PR do?

Python 3.14 was released yesterday - the current `Check API Bindings` CI job is not pinning the Python version. Dependency install fails on `3.14` (due to `shapely` being required in the `google-cloud-aiplatform`).

Tried installing dependencies with `3.14` locally and getting different dependency issues coming out of a different dependency chain:
```
`pyobjc-core` (v10.3.1) was included because `kiln-root` (v0.1.0) depends on `kiln-studio-desktop` (v0.1.0) which depends on `pystray`
        (v0.19.5) which depends on `pyobjc-framework-quartz` (v10.3.1) which depends on `pyobjc-core`
```

Easier to pin Python to `3.13` in CI for now - but looks like we need to upgrade all the CI jobs to use `3.14` and bump up a number of dependencies to support Python `3.14`.

## Checklists


.github/workflows/coverage.yml
```diff
@@ -1,19 +1,19 @@
 name: Coverage Report
 
 on:
-    pull_request:
-    push:
-        branches: [ main ]
+  pull_request:
```
leonardmq, line 4: Auto format with default `YAML` formatter in Cursor - indent was inconsistent here

.github/workflows/coverage.yml
```diff
@@ -1,19 +1,19 @@
...
     runs-on: ubuntu-latest
     permissions:
-      contents: read          # To checkout the code
-      pull-requests: write    # To comment on PRs
-      issues: write          # To comment on issues (PRs are issues)
-      actions: read          # To read workflow artifacts
+      contents: read # To checkout the code
+      pull-requests: write # To comment on PRs
+      issues: write # To comment on issues (PRs are issues)
+      actions: read # To read workflow artifacts
```
leonardmq, line 16: Auto format with default `YAML` formatter in Cursor

.github/workflows/coverage.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: Comment PR with coverage info
         if: github.event_name == 'pull_request'
         uses: actions/github-script@v7
-        continue-on-error: true  # Don't fail the workflow if we can't comment (e.g., from forks)
+        continue-on-error: true # Don't fail the workflow if we can't comment (e.g., from forks)
```
leonardmq, line 80: Auto format with default `YAML` formatter in Cursor

scosman (approved):

PR #711, opened 2025-10-08
refactor: filter docs by tags when traversing in rag loader

leonardmq:
## What does this PR do?

Changes:
- refactor: filter by tags in the vector store loader
- test: add tests for the filtering

The `VectorSearchLoader` is iterating over the entire subtree for each document, which is wasteful because only documents with the target tags can ever have relevant descendant chunks (extractions, chunks, embeddings). Filtering documents by tags on the outermost loop saves us a lot of traversing.

Previous behavior was yielding the correct chunks by coincidence / implication. We only yield the (docs x chunks x embeddings) that have all the relevant configs; and if a chunk has all the right configs then it must have been from the right RAG Config (or an identical copy of it). So correct, but implicit and wasteful.

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Document loading now respects tag filters: when no tags are configured all project documents are included; when tags are set only documents with matching tags are processed, supporting multiple tags for improved relevance.

* **Tests**
  * Added comprehensive tests covering no-tag, single-match, no-match, partial-match and multi-tag scenarios to validate filtering behavior.

### DIFF
--- a/libs/core/kiln_ai/adapters/vector_store_loaders/test_vector_store_loader.py
+++ b/libs/core/kiln_ai/adapters/vector_store_loaders/test_vector_store_loader.py
@@ -131,6 +131,34 @@ def rag_config_factory(mock_project):
     return fn
 
 
+@pytest.fixture
+def document_factory(mock_project, mock_attachment_factory):
+    def fn(
+        name: str = "Test Document",
+        description: str = "Test Document",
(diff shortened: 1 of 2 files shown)


leonardmq:
CI broken due to Python 3.14 - waiting on https://github.com/Kiln-AI/Kiln/pull/708 to be merged

PR #713, opened 2025-10-08
MCP server fix, and better typing for tools

scosman:
- make our tool interface typed. We should never have allowed "Any". It caused downstream bugs where Python objects were serialized into strings in the MCP server.
 - Update lots of code/tests to used new type
 - Fix the MCP server to work as Kiln Task as Tools
 - Remove ugly factory style in favour of Mocks

## Summary by CodeRabbit

- New Features
  - Automatic discovery of Kiln Task tools from external servers when collecting project tools.
  - Added a read-only option when accessing a project's tasks.

- Improvements
  - Standardized tool outputs across the platform, providing more consistent responses in chats and tool calls.
  - Running Kiln Task Tool without an explicit context now uses a safe default (no saving), reducing unexpected errors and improving reliability.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
@@ -625,9 +625,9 @@ class LiteLlmAdapter(BaseAdapter):
                     role="tool",
                     tool_call_id=tool_call.id,
                     content=result.output,
-                    kiln_task_tool_data=None
-                    if not isinstance(result, KilnTaskToolResult)
-                    else result.kiln_task_tool_data,
+                    kiln_task_tool_data=result.kiln_task_tool_data
+                    if isinstance(result, KilnTaskToolResult)
+                    else None,
                 )
             )
 


libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -29,14 +31,20 @@ class ToolCallContext:
...
+
 class KilnToolInterface(ABC):
     """
     Abstract interface defining the core API that all Kiln tools must implement.
     This ensures consistency across all tool implementations.
     """
 
     @abstractmethod
-    async def run(self, context: ToolCallContext | None = None, **kwargs) -> Any:
+    async def run(
```
leonardmq, line 45: The compiler is not enforcing this interface. The concrete implementations of `run` can take in any args and return anything regardless of what we declare here. 

Demo forking off this branch here - implementation of the `RagTool` violates this interface and no type error surfaces: 
https://github.com/Kiln-AI/Kiln/compare/scosman/mcp_server_tat_fix...leonard/show-typecheck-not-enforced?expand=1

We need to merge this to enforce the interface: 
https://github.com/Kiln-AI/Kiln/pull/701

PR #714, opened 2025-10-09
fix: warnings in rag tests

leonardmq:
## What does this PR do?

Fix warnings in RAG runner tests that were caused by breaking out of the `collect_records` async generator early. When breaking out of an async generator, we must close it explicitly.

Also add better handling for case where there are no upstream records to index (due to upstream stages failing); it used to show as an error in the RAG logging UI that `Vector dimensions are not set`, but now will simply show `INFO: No records to index` (after the upstream error logs).

## Checklists

## Summary by CodeRabbit

* **Bug Fixes**
  * No longer throws an error when there are no records to index; shows a clear progress message indicating “No records to index.”
  * Improves stability by gracefully stopping early and cleaning up resources when no records are found.
* **Tests**
  * Added tests covering early-return scenarios with no records, verifying progress messaging and proper cleanup.
  * Updated expectations to ensure no indexing actions are attempted when there are no records.


leonardmq:
@coderabbitai review

PR #718, opened 2025-10-09
Refactor AvailableModelsDropdown to take in a settings object

sfierro:
This will help with my next PR which needs to pass through model dropdown options from the new RunConfigComponent so we don't duplicate props.

## Summary by CodeRabbit

- New Features
  - Run configs now auto-select a default (saved or “custom”) when none is chosen.
  - Model picker now accepts a single settings object to control requirements and suggested mode, simplifying configuration.

- Refactor
  - Unified model dropdown logic to read from centralized settings for consistent filtering, warnings, and badges across the app.
  - Updated pages and dialogs to use the new settings-based model picker; behavior preserved while API surface changed.

### DIFF
--- a/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
@@ -34,7 +34,6 @@
   export let current_task: Task
   export let model_name: string = ""
   export let provider: string = ""
-  export let requires_structured_output: boolean = false
   export let model_dropdown_settings: Partial<ModelDropdownSettings> = {}
   export let selected_run_config_id: string | null = null
   export let save_config_error: KilnError | null = null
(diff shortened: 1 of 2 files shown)


app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
```diff
@@ -27,13 +27,14 @@
...
   import { load_task_prompts } from "$lib/stores/prompts_store"
+  import type { ModelDropdownSettings } from "./model_dropdown_settings"
 
   // Props
   export let project_id: string
   export let current_task: Task
   export let model_name: string = ""
   export let provider: string = ""
-  export let requires_structured_output: boolean = false
+  export let model_dropdown_settings: Partial<ModelDropdownSettings> = {}
```
leonardmq, line 37: Was the main intent to make managing this over here as a whole easier (instead of a bunch of independent values)?

sfierro: yeah basically didn't want to copy and paste the 7 variables or whatever from the child, and if the other children change and also require a bunch of settings props we can do this for all instead of copy and pasting them all. its safer for clients to know exactly what each setting does too instead of having to know the underlying implementation

PR #719, opened 2025-10-09
[KIL-145] Refactor SDG generate outputs dialog to use new RunConfigComponent

sfierro:
## Summary by CodeRabbit

- New Features
  - Unified run configuration panel on Generate so model, provider, prompt, tools, and structured output are selected together.
  - Added safeguards to prevent generation when run configuration is missing.

- Refactor
  - Streamlined generation and sample-saving flows to use a single run configuration object for output settings.

- Chores
  - Enhanced analytics to include tool-usage metrics during runs.

### DIFF
--- a/app/desktop/studio_server/eval_api.py
+++ b/app/desktop/studio_server/eval_api.py
@@ -154,8 +154,6 @@ class EvalProgress(BaseModel):
     golden_dataset_fully_rated_count: int
     # The current selected eval method
     current_eval_method: EvalConfig | None
-    # The current selected run method
-    current_run_method: TaskRunConfig | None
 
 
 class EvalResultSummary(BaseModel):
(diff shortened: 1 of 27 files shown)


app/web_ui/src/routes/(app)/run/+page.svelte
```diff
@@ -85,6 +85,9 @@
         mcp_tools: run_config_component
           .get_tools()
           .filter((tool) => tool.startsWith("mcp::")).length,
+        kiln_task_tools: run_config_component
```
sfierro, line 88: opportunistic fix :D

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -544,12 +528,14 @@
 
   async function generate_sample(
     sample: SampleData,
-    model_name: string,
-    provider: string,
-    prompt_method: string,
     topic_path: string[] | undefined,
   ): Promise<GenerateSampleResponse> {
+    const run_config_properties =
+      run_config_component?.run_options_as_run_config_properties()
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:534]: @scosman - do we allow accessing child components' methods from a parent component via `bind:this`?

Not familiar enough with Svelte to know if this is to avoid (due to possibly obscure reactivity / lifecycle interactions). Maybe writing to a store from the child and reading from the store here would be safer

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -544,12 +528,14 @@
...
     try {
+      if (!run_config_properties) {
+        throw new KilnError("Run config properties not found")
+      }
```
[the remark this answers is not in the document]
Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:538]: run_options_as_run_config_properties can't throw

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:534]: discussed on slack

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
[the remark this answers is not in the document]
Comment (leonardmq) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:538]: @coderabbitai - do you know why the component reference here could become `null`? 

It seems to become `null`, apparently (but not necessarily because of, I don't know) after a warning we see in devtools console showing: `Form submission canceled because the form is not connected`. What happens is a few calls succeed, and then after that, we get `Run config properties not found` on the subsequent calls and `run_config_component` is undefined / null

PR #723, opened 2025-10-12
[KIL-110] Add Tools to Dataset Run Properties

sfierro:
screenshot_2025-10-12_at_4 50 40___pm_720

## Summary by CodeRabbit

- New Features
  - Display “Available Tools” on run and tool settings pages with clickable badges and links to relevant tool details.
  - Property lists now support multiple values, rendering them as badges or linked items.

- Improvements
  - Smarter linking for various tool types, routing users to the appropriate settings or docs pages.
  - Faster page loads by fetching models and tools in parallel.
  - More consistent tool data refresh after operations.
  - Clearer states when tool data is loading or unavailable (e.g., “Loading...”, “None”).

### DIFF
--- a/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task/[tool_server_id]/+page.svelte
+++ b/app/web_ui/src/routes/(app)/settings/manage_tools/[project_id]/kiln_task/[tool_server_id]/+page.svelte
@@ -11,6 +11,8 @@
     Task,
     TaskRunConfig,
     KilnTaskServerProperties,
+    ProviderModels,
+    PromptResponse,
   } from "$lib/types"
   import {
     load_available_models,
@@ -34,6 +36,7 @@
     prompts_by_task_composite_id,
   } from "$lib/stores/prompts_store"
   import { get_tools_property_info } from "$lib/stores/tools_store"
+  import type { UiProperty } from "$lib/ui/property_list"
 
   $: project_id = $page.params.project_id
   $: tool_server_id = $page.params.tool_server_id
@@ -50,6 +53,7 @@
   let unarchive_error: KilnError | null = null
   let tools_property_value: string | string[] = "Loading..."
   let tool_links: (string | null)[] | undefined = undefined
+  let run_config_properties: UiProperty[] = []
 
   onMount(async () => {
     await fetch_tool_server()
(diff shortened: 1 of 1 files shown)


app/web_ui/src/lib/stores/tools_store.ts
```diff
@@ -9,3 +11,37 @@ export const { store: tools_store, initialized: tools_store_initialized } =
...
   })
+
+export function get_tools_property_info(
+  tool_ids: string[],
+  project_id: string,
+  available_tools: Record<string, ToolSetApiDescription[]>,
+): { value: string | string[]; links: (string | null)[] | undefined } {
+  const project_tools = available_tools[project_id]
+  if (!project_tools) {
+    return { value: "Loading...", links: undefined }
```
leonardmq, line 22: This might be unexpected behavior at the caller; it could be more linear if we checked this at the caller and handled loading of the whole property block over there? 

But I also get there is a tradeoff with null checks everywhere over there if we do that - and this is not particularly problematic

app/web_ui/src/lib/ui/property_list.svelte
```diff
@@ -32,7 +33,56 @@
...
+        {#if Array.isArray(property.value)}
+          {#if property.badge}
+            <div class="flex flex-wrap gap-1">
+              {#each property.value as value, i}
+                {@const link = property.links
+                  ? property.links.length == property.value.length
+                    ? property.links[i]
+                    : null
+                  : null}
+                <button
```
leonardmq, line 45: Might be worth extracting components out of this because it seems like we are doing the same thing in a loop here as we are doing for individual properties.

Maybe could do something like: 
- `PropertyListValue` -> handle rendering an individual value (badge, link, etc.)
- `PropertyListArray` -> rendering the multiple `PropertyListValue` in some kind of loop and the flex wrapper

Worth making a ticket in my opinion, but not blocking this PR

app/web_ui/src/lib/utils/link_builder.ts
```diff
@@ -34,3 +34,23 @@ export function dataset_item_link(
...
+  ) {
+    return `/settings/manage_tools/${project_id}/tool_servers/${tool_id.split("::")[2]}`
+  } else if (tool_id.startsWith("kiln_task::")) {
+    return `/settings/manage_tools/${project_id}/kiln_task/${tool_id.split("::")[1]}`
+  } else if (tool_id.startsWith("kiln_tool::rag::")) {
+    return `/docs/rag_configs/${project_id}/${tool_id.split("::")[2]}/rag_config`
+  } else if (tool_id.startsWith("kiln_tool::")) {
+    return `/settings/manage_tools/${project_id}`
+  } else {
+    return null
```
leonardmq, line 54: Might be worth warning logging what the value is here

PR #724, opened 2025-10-13
refactor: extract some icons into components

leonardmq:
## What does this PR do?

We should follow this pattern to make icons a little easier to manage and know which one is used for what. I have only done a handful of them, we can do the rest as we go along.

Note: 
- passing `class` around as prop does not work (something caused by the build process; not sure what) so we cannot declare a `class_override` prop or something. Because of that, each icon component should declare `class="w-full h-full"`, and the SVG should have `currentColor`.

We can apply styling with a parent `div` at the caller to override the style, like:
```ts
<div class="text-red-500 h-12 w-12">
  <SomeIconComponent />
</div>
```

Changes:
- extract some hardcoded SVGs into components
- move all icon components into `ui/icons/` (instead of `ui/`)
- replace references
- add `qna_icon`

## Checklists

## Summary by CodeRabbit

- New Features
  - Introduced reusable icons for Evals, Fine-tune, Q&A, and Documents.
  - Updated navigation, empty states, and dialogs to use the new icons.

- Refactor
  - Replaced inline SVGs with consistent icon components across the app.
  - Standardized the checkmark to a unified icon component.
  - Removed unused script blocks from icon components.

- Style
  - Unified icon sizing and spacing for a more consistent look.
  - Minor layout tweaks around icons to improve visual alignment.


leonardmq:
@coderabbitai review

PR #725, opened 2025-10-13
refactor: synth data page become routing entry point to different modes

leonardmq:
## What does this PR do?

Extracted from a bigger branch into this standalone PR to decrease the risk of mega conflicts due to moving giant files around for Q&A synthetic data.

The diff looks big, mostly because of the files moving around:
- https://github.com/Kiln-AI/Kiln/pull/725/commits/8bf11e0334b5a595506239ed4f515f93d1174cdc -> only moves files around
- https://github.com/Kiln-AI/Kiln/pull/725/commits/631e83c8d70b76656f15856e0c571020fa292c7a -> the actual changes to the logic (reviewing this one is much easier) 

We have a new synth data type (Docs Q&A) that works very differently from how the current synth data works. The new type uses docs from RAG to generate questions/answers based on the docs - the flow is different and the objects it deals with are different (no concept of topics / inputs / outputs; instead it has questions / answers).

The current synth data page is coupled to the current synth data modes and we need to move that specificity down to make room for the new type of synth data that will be living alongside the current ones.

Changes:
- refactor: move current main synth data page into a `synth/` path
- refactor: make the current main synth data page into a routing layer that passes on the params to the mode-specific page (`/synth`; and later we will have `/qna`)
- refactor: the logic for what the empty state components trigger is now passed via URL params (since it moves from being inside the same component to being a separate page under `synth/`)

## Checklists

## Summary by CodeRabbit

- New Features
  - Adds a comprehensive synthetic data generation page: multi-step workflow (add topics, generate inputs/outputs, save), parallel generation, per-sample status/error tracking, bulk save with progress, configurable run settings, and contextual guidance.

- Refactor
  - Replaces the previous heavy intro with a lightweight entry that routes eval/fine-tune flows to the new synth experience and removes legacy multi-step/stateful UI from the intro.

### DIFF
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
@@ -4,23 +4,84 @@
   import { page } from "$app/stores"
   import { goto } from "$app/navigation"
   import DataGenIntro from "./data_gen_intro.svelte"
+  import { indexedDBStore } from "$lib/stores/index_db_store"
+  import { writable, type Writable } from "svelte/store"
 
+  let loading = true
   $: project_id = $page.params.project_id
   $: task_id = $page.params.task_id
 
-  onMount(() => {
-    handle_routing()
+  // we only need gen_type to do the routing, the type-specific data is handled by the
+  // mode-specific pages we redirect to
+  type SavedDataGenState = {
+    gen_type?: "training" | "eval" | null
+  }
+
+  let saved_state: Writable<SavedDataGenState> = writable({
+    gen_type: null,
   })
 
-  function handle_routing() {
+  onMount(async () => {
+    await handle_routing()
+  })
+
+  async function handle_routing() {
+    loading = true
     const reason_param = $page.url.searchParams.get("reason")
 
-    // Eval/Fine-tuning modes redirect to synth page
+    // Eval/Fine-tuning modes redirect to synth page - this is when we explicitly link back
+    // to the synth page (e.g. via toast UI)
     if (reason_param === "eval" || reason_param === "fine_tune") {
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -0,0 +1,48 @@
...
+  $: project_id = $page.params.project_id
+  $: task_id = $page.params.task_id
+
+  onMount(() => {
+    handle_routing()
+  })
+
+  function handle_routing() {
+    const reason_param = $page.url.searchParams.get("reason")
+
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:31]: Here we are going to have another `reason_param === "qna"` that will route to a `qna/` page (sibling of the `synth/` page)

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth/+page.svelte
```diff
@@ -0,0 +1,1288 @@
...
+    // 1. There is no saved state and URL has state: setup the URL state
+    // 2. URL state matches saved state: load the saved state
+    // 3. Saved state and URL state are different: show an alert to the user asking if they want to replace the saved state with the URL state
+    // 4. There's no URL state and there is saved state: setup the saved state (with a UI option to clear it)
+    // 5. No state, don't setup, and wait for the user to setup via UI
+
+    // The URL params can specify a specific setup for the data gen.
+    const reason_param = $page.url.searchParams.get("reason")
+    // Map "fine_tune" (user facing string) to "training" (pre-existing internal value)
+    const gen_type = reason_param === "fine_tune" ? "training" : reason_param
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth/+page.svelte:184]: Here we map `fine_tune` to `training`. On the routing page, we pass on `reason=fine_tune` that gets read here

PR #737, opened 2025-10-21
Leverage run config options in SDG topic / input gen

sfierro:
## Summary by CodeRabbit

* **Refactor**
  * Unified model/provider settings into a single run-config object used across topic and sample generation; API payloads now use this consolidated run_config_properties.

* **New Features**
  * Replaced model dropdowns with a unified Run Config UI control and added structured-output support where applicable.
  * Run Config exposes new options to control prompt/tools display and supports task-less usage.

* **UX/Bug Fixes**
  * Safer handling of null/no-task states and improved tool loading/filtering to prevent errors.

### DIFF
--- a/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
@@ -178,7 +178,7 @@
   }
 
   export async function save_new_run_config(): Promise<TaskRunConfig | null> {
-    if (!project_id || !current_task?.id) {
+    if (!current_task?.id) {
       return null
     }
     try {
@@ -203,8 +203,11 @@
   async function get_selected_run_config(): Promise<
     TaskRunConfig | "custom" | null
   > {
+    if (!current_task?.id) {
+      return null
+    }
     // Make sure the task run configs are loaded, will be quick if they already are
-    await load_task_run_configs(project_id, current_task?.id ?? "")
+    await load_task_run_configs(project_id, current_task.id)
 
     // Map selected ID back to TaskRunConfig object
     if (!selected_run_config_id) {
@@ -215,7 +218,7 @@
       // Find the config by ID
       const all_configs =
         $run_configs_by_task_composite_id[
-          get_task_composite_id(project_id, current_task?.id ?? "")
+          get_task_composite_id(project_id, current_task.id)
         ] ?? []
       let run_config = all_configs.find(
         (config) => config.id === selected_run_config_id,
@@ -253,7 +256,7 @@
 
 <div class="w-full flex flex-col gap-4">
   <AvailableModelsDropdown
-    task_id={current_task?.id ?? ""}
+    task_id={current_task?.id ?? null}
     bind:model
     settings={updated_model_dropdown_settings}
     bind:error_message={model_dropdown_error_message}
--- a/app/web_ui/src/lib/ui/run_config_component/tools_selector.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/tools_selector.svelte
@@ -22,9 +22,7 @@
 
   async function load_tools(project_id: string, task_id: string | null) {
     // Load available tools
-    if (project_id) {
-      load_available_tools(project_id)
-    }
+    load_available_tools(project_id)
 
     if (!task_id) {
       tools = []


app/desktop/studio_server/data_gen_api.py
```diff
@@ -102,17 +94,12 @@ async def generate_categories(
...
+        run_config_properties = input.run_config_properties
+        # Override prompt id to simple just in case we change the default in the UI in the future.
+        run_config_properties.prompt_id = PromptGenerators.SIMPLE
         adapter = adapter_for_task(
             categories_task,
-            run_config_properties=RunConfigProperties(
-                model_name=input.model_name,
-                model_provider_name=model_provider_from_string(input.provider),
-                prompt_id=PromptGenerators.SIMPLE,
-                # We don't expose setting this manually in the UI, so pull a recommended mode from ml_model_list
```
sfierro, line 111: Now we do expose this setting in the UI!

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
```diff
@@ -31,20 +31,22 @@
...
   export let hide_create_kiln_task_tool_button: boolean = false
+  export let hide_prompt_selector: boolean = false
+  export let show_tools_selector_in_advanced: boolean = false
+  export let requires_structured_output: boolean = false
 
   let model: string = $ui_state.selected_model
   let prompt_method: string = "simple_prompt_builder"
   let tools: string[] = []
   let requires_tool_support: boolean = false
-  $: requires_structured_output = !!current_task.output_json_schema
```
sfierro, line 47: need parent to pass this in now since task is now nullable but we still need this setting even without a task

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte
```diff
@@ -268,17 +284,20 @@
...
+          model_dropdown_settings={{
             requires_data_gen: true,
             requires_uncensored_data_gen:
               guidance_data.suggest_uncensored($selected_template),
             suggested_mode: guidance_data.suggest_uncensored($selected_template)
               ? "uncensored_data_gen"
               : "data_gen",
           }}
-          bind:model
         />
```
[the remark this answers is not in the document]
Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/generate_samples_modal.svelte:301]: I know the task input json schema is always set for topic/inputs so ignoring.

app/web_ui/src/lib/ui/run_config_component/create_new_run_config_dialog.svelte
```diff
@@ -66,6 +66,7 @@
           bind:this={run_config_component}
           {project_id}
           current_task={task}
+          requires_structured_output={!!task.output_json_schema}
```
leonardmq: Caller might forget to pass this, and seems like we could derive it internally in the `RunConfigComponent` since it takes in the `current_task`? Or are there cases where this component does not expect to take in a task at all?

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
leonardmq: What are the scenarios / pages where this component does not need / expect a task?

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
```diff
@@ -202,7 +204,7 @@
     TaskRunConfig | "custom" | null
   > {
     // Make sure the task run configs are loaded, will be quick if they already are
-    await load_task_run_configs(project_id, current_task.id ?? "")
+    await load_task_run_configs(project_id, current_task?.id ?? "")
```
leonardmq: This causes a crash:
- On an empty Fine Tuning synth data session
- Click `Add Topics` or `Generate Model Inputs`
- The current task id is an empty string and the API calls goes to `run_configs_store.ts:59  GET http://localhost:8757/api/projects/842529506281/tasks//task_run_configs 404 (Not Found)` (notice the path segment where there should be the task id is empty)

leonardmq (changes requested):
I approved but then ran into a crash during manual clicking around

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
sfierro: for when we want transient run config options (synth data topic/input gen) vs. are running an actual Kiln Task

app/web_ui/src/lib/ui/run_config_component/run_config_component.svelte
sfierro: ah thanks, adding a task_id null check first for early return

PR #748, opened 2025-10-23
Use unique name for GCP bucket

chiang-daniel:
## What does this PR do?

Vertex fine-tune is broken because we are using the same `kiln-ai-data` as bucket name for all clients. This needs to be globally unique. Fix the bucket name logic to include the vertex project id which is also globally unique. 

Verified with a successful vertex fine-tune session.
image

## Checklists

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed storage bucket naming to incorporate project-specific identifiers, removing hard-coded bucket references for better isolation and flexibility.

* **Tests**
  * Added tests validating project-to-bucket name mapping and ensuring unique bucket name generation across projects.

### DIFF
--- a/libs/core/kiln_ai/adapters/fine_tune/vertex_finetune.py
+++ b/libs/core/kiln_ai/adapters/fine_tune/vertex_finetune.py
@@ -219,7 +219,7 @@ class VertexFinetune(BaseFinetuneAdapter):
 
     @classmethod
     def _unique_bucket_name(cls) -> str:
-        project_id = Config.shared().vertex_project_id
+        project_id, _ = cls.get_vertex_provider_location()
         # See https://cloud.google.com/storage/docs/buckets#naming
         # Bucket names must contain 3-63 characters
         # See https://cloud.google.com/resource-manager/docs/creating-managing-projects


sfierro (approved):

libs/core/kiln_ai/adapters/fine_tune/test_vertex_finetune.py
```diff
@@ -276,15 +276,15 @@ async def test_generate_and_upload_jsonl(
     )
 
     mock_path = Path("mock_path.jsonl")
-    expected_uri = "gs://kiln-ai-data/1234567890/mock_path.jsonl"
+    expected_uri = "gs://kiln-ai-data-test-project/1234567890/mock_path.jsonl"
```
scosman, line 279: why change this string? is it only going to work on your account now (and it only worked on mine before?)

chiang-daniel: The test is using mock and is expecting the id to be in the new format. `kiln-ai-data-{project_id}`, project_id in the test is "test-project"

PR #756, opened 2025-10-28
Tool call eval

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Tool definition API to fetch tool function, description, and params.
  * "Tool Call" / "Appropriate Tool Use" eval template and UI with full-trace support.
  * EvalDataType (final_answer | full_trace) and optional full-trace task_run_trace in eval runs.
  * Session_id and tool_id support for synthetic data generation; mandatory-tools & single-select tools UI; disabled-option keyboard/navigation.

* **Tests**
  * Expanded unit/integration tests covering eval trace formatting, eval runner paths, tool definition and UI behaviors.

### DIFF
--- a/app/web_ui/src/lib/ui/run_config_component/tools_selector.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/tools_selector.svelte
@@ -177,9 +177,10 @@
   {:else}
     <FormElement
       id="tools"
-      label="Tools & Search"
+      {label}
+      {description}
       inputType="multi_select"
-      info_description="Select the tools available to the model. The model may or may not choose to use them."
+      {info_description}
       bind:value={tools}
       fancy_select_options={get_tool_options($available_tools[project_id])}
       empty_state_message={$available_tools[project_id] === undefined


app/web_ui/src/lib/stores/tools_store.test.ts
```diff
@@ -0,0 +1,180 @@
...
+          params: {
+            path: {
+              project_id: "test_project_id",
+              task_id: "test_task_id",
+              tool_id: "test_tool_id",
+            },
+          },
+        },
+      )
+    })
```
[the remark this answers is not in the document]
sfierro: this is wrong regarding the data = null check its already verified by the error. regarding the type, it's always gonna be string if we hit a crash we did something wrong in our own code

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth/+page.svelte
```diff
@@ -194,28 +208,42 @@
...
+        setup(
+          gen_type,
+          template_id,
+          eval_id,
+          tool_id,
+          project_id,
+          task_id,
+          splits,
+          session_id,
+        )
```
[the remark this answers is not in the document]
Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth/+page.svelte]: good idea on the legacy session thing. but for the ID its ok if they collide its a risk we're willing to take for IDs to be human readable right now

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/select_eval_template.svelte
```diff
@@ -429,3 +599,55 @@
...
+    on:submit={create_tool_call_eval}
+    warn_before_unload={!!(
+      !tool_call_eval_create_complete &&
+      (tool_call_eval_name ||
+        should_call_tool_guidelines ||
+        should_not_call_tool_guidelines ||
+        selected_tool !== null)
+    )}
+  >
+    <div class="font-light text-sm">
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/select_eval_template.svelte]: Would be more standard to put this as the `subtitle` or `sub_subtitle` in the `Dialog` component instead

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/select_eval_template.svelte
```diff
@@ -429,3 +599,55 @@
...
+    />
+    <FormElement
+      label="Should Call Guidelines"
+      description="Guidelines or examples for when the tool should be called."
+      info_description="Include guidelines or examples to help the judge model understand when the tool should be called. The format is flexible (plain text). You can include a description or multiple examples if needed."
+      inputType="textarea"
+      id="should_call_example"
+      bind:value={should_call_tool_guidelines}
+    />
+    <FormElement
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/select_eval_template.svelte:645]: 1Password Chrome extension thinks this is a password input. We ran into this before, and we handle it in the `FormElement` for normal text and number input, but not for `textarea`.

Can you add the same flags in `FormElement` (https://github.com/Kiln-AI/Kiln/blob/9f0c1412c41d1bea3987a6bdc6d4953504f0ec37/app/web_ui/src/lib/utils/form_element.svelte#L169):
```ts
autocomplete="off"
data-op-ignore="true"
```

libs/core/kiln_ai/adapters/eval/eval_utils/eval_trace_formatter.py
```diff
@@ -0,0 +1,165 @@
...
+            or not isinstance(message["content"], str)
+        ):
+            return None
+
+        # For Kiln task tools, extract just the output field from the JSON response
+        if message["role"] == "tool":
+            try:
+                parsed = json.loads(message["content"])
+                if parsed and isinstance(parsed, dict) and "output" in parsed:
+                    return parsed["output"]
```
leonardmq, line 103: Maybe check that `parse["output"]` is a string (I guess it always is? But the type here is `unknown`)

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/+page.svelte
```diff
@@ -109,6 +111,9 @@
               eval_set_filter_id,
               eval_configs_filter_id,
               template_properties,
+              evaluation_data_type: (evaluation_data_type || "final_answer") as
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/create_evaluator/+page.svelte]: Can we define the `final_answer | full_trace` literal on the `EvalTemplateResult` type instead of `string` so here we would not need casting? Seems like we can also grab the type from the SDK (`export type EvalDataType = components["schemas"]["EvalDataType"]`) as it is exposed via the API

app/web_ui/src/lib/ui/fancy_select.svelte
```diff
@@ -480,14 +480,24 @@
...
-        )
+        const flatOptions = filteredOptions.flatMap((group) => group.options)
+        let nextIndex = focusedIndex + 1
+        while (
+          nextIndex < flatOptions.length &&
+          flatOptions[nextIndex].disabled
+        ) {
+          nextIndex++
+        }
+        focusedIndex = Math.min(nextIndex, flatOptions.length - 1)
```
leonardmq: Not sure if it matters, but the `flatOptions.length - 1` we fall back on may be disabled - maybe should keep a copy of the original `focusedIndex` and `Math.min(nextIndex, oldFocusedIndex)` if it matters

Maybe also good to extract into its own function (e.g. `nextEnabledOptionIndex(xxx)`) as this logic seems to be appearing twice in the file (here: https://github.com/Kiln-AI/Kiln/blob/9f0c1412c41d1bea3987a6bdc6d4953504f0ec37/app/web_ui/src/lib/ui/fancy_select.svelte#L541)

app/web_ui/src/lib/ui/fancy_select.svelte
```diff
@@ -480,14 +480,24 @@
...
         scrollToFocusedIndex()
       } else if (event.key === "ArrowUp") {
         event.preventDefault()
-        focusedIndex = Math.max(focusedIndex - 1, 0)
+        const flatOptions = filteredOptions.flatMap((group) => group.options)
+        let prevIndex = focusedIndex - 1
+        while (prevIndex >= 0 && flatOptions[prevIndex].disabled) {
+          prevIndex--
+        }
+        focusedIndex = Math.max(prevIndex, 0)
```
leonardmq: Same here with `0` possibly being disabled - not sure if it matters.

Also good to extract this stepping logic into a function as it happens again further down in this file

app/desktop/studio_server/tool_api.py
```diff
@@ -668,3 +669,41 @@ async def get_search_tools(project_id: str) -> list[SearchToolApiDescription]:
             for rag_config in project.rag_configs(readonly=True)
             if not rag_config.is_archived
         ]
+
+    @app.get("/api/projects/{project_id}/tasks/{task_id}/tools/{tool_id}/definition")
+    async def get_tool_definition(
+        project_id: str, task_id: str, tool_id: str
+    ) -> dict[str, Any]:
```
leonardmq: Can we return a `TypedDict` or `BaseModel` here instead of `dict[str, Any]` - we would get some typing on the known keys in the frontend and we would not need casting here: https://github.com/Kiln-AI/Kiln/blob/9f0c1412c41d1bea3987a6bdc6d4953504f0ec37/app/web_ui/src/lib/stores/tools_store.ts#L73

app/web_ui/src/lib/ui/run_config_component/tools_selector.svelte
```diff
@@ -9,10 +9,18 @@
...
   export let task_id: string | null = null
+  export let label: string = "Tools & Search"
+  export let description: string | undefined = undefined
+  export let info_description: string | undefined =
+    "Select the tools available to the model. The model may or may not choose to use them."
   export let tools: string[] = []
+  export let mandatory_tools: string[] | null = null
   export let hide_create_kiln_task_tool_button: boolean = false
+  export let single_select: boolean = false
+  export let single_select_selected_tool: string | null = null // Only used if single_select is true
```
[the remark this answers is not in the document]
sfierro: we don't want to use the store for the single select case

PR #758, opened 2025-10-28
Fine-tune with traces & tool calls

chiang-daniel:
## What does this PR do?

Previously fine-tune only supports simple message formats - system, user, assistant. Since the introduction of traces, we now have a full trace from a run including tool calls. This PR adds support to build datasets for fine-tuning with traces and tool calling. 

* Add new dataset formatter for traces
* Tested fine-tune with Fireworks and the download options. 

## Checklists

[sample fireworks dataset.txt](https://github.com/user-attachments/files/23219143/sample.fireworks.dataset.txt)
@tawnymanticore 

## Summary by CodeRabbit

* **New Features**
  * Trace-based fine-tuning formatter with multiple export formats and a public dataset-format enum; dataset export uses it when traces exist.
  * New chunker UI components and typed chunker/embedding property models; structured LanceDB/vector-store property support and new embedding-config GET API.
  * Expanded model catalog (vision-language variants) and deterministic provider ordering for OpenRouter. Added generate_model_id helper.

* **Tests**
  * Extensive unit tests for trace formatter, chunkers, LanceDB paths, model/provider behavior, and OpenRouter ordering.

* **Chores**
  * Minor UI/presentation tweaks and internal payload shape adjustments.


libs/core/kiln_ai/adapters/fine_tune/dataset_format.py
```diff
@@ -0,0 +1,25 @@
+from enum import Enum
+
+
+class DatasetFormat(str, Enum):
+    """Formats for dataset generation. Both for file format (like JSONL), and internal structure (like chat/toolcall)"""
+
+    """OpenAI chat format with plaintext response"""
+    OPENAI_CHAT_JSONL = "openai_chat_jsonl"
```
chiang-daniel, line 8: - [ ] Together AI doesn't support dataset with `Tool` role, it seems like it's not supporting fine-tune with tools in general. Existing system/user/assistant pattern still works.

libs/core/kiln_ai/adapters/fine_tune/trace_based_dataset_formatter.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+
+        return {"messages": messages}
+
+    def generate_vertex_gemini(
+        self,
+        trace: list[ChatCompletionMessageParam],
+    ) -> Dict[str, Any]:
+        """Generate vertex gemini message from trace"""
+        # See https://cloud.google.com/vertex-ai/generative-ai/docs/models/tune-function-calling
```
chiang-daniel, line 228: - [x] not working for parallel (multiple tool calls) yet. Need to find working examples.

libs/core/kiln_ai/adapters/fine_tune/trace_based_dataset_formatter.py
```diff
@@ -0,0 +1,313 @@
...
+        # Get content of last message
+        last_message_content = last_message.get("content", None)
+
+        messages.append(
+            {
+                "role": "assistant",
+                "content": None,
+                "tool_calls": [
+                    {
+                        "id": "call_1",
```
tawnymanticore, line 166: is `call_1` here static? I thought the tool ID is a hash or something in your test cases

chiang-daniel: This was followed by the same change we did prior. @scosman any reason why this was static instead of dynamic?

libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py
```diff
@@ -370,13 +351,25 @@ def dump_to_file(
...
+                    formatter = TraceBasedDatasetFormatter(
+                        system_message=self.system_message
+                    )
+                    example = formatter.build_training_chat_from_trace(
+                        task_run=task_run,
+                        data_format=format_type,
+                    )
+                else:
+                    # Training_chat is a list of ChatMessage
+                    training_chat = build_training_chat(
```
leonardmq, line 365: Is `build_training_chat` legacy / backward compatibility or will there be scenarios where task runs going forward could have no trace?

libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py
```diff
@@ -370,13 +351,25 @@ def dump_to_file(
...
-                training_chat = build_training_chat(
-                    task_run=task_run,
-                    system_message=self.system_message,
-                    data_strategy=data_strategy,
-                    thinking_instructions=self.thinking_instructions,
-                )
-                example = generator(training_chat)
+                # Create example if trace is available and not empty
+                if task_run.trace is not None and len(task_run.trace) > 0:
+                    # Call build_training_chat_from_trace
```
leonardmq: These two comments seem redundant - a comment on why we have task runs with / without trace could be helpful though

libs/core/kiln_ai/adapters/fine_tune/dataset_formatter.py
chiang-daniel: My understanding is that all runs will have traces! @sfierro this is true right? 
@leonardmq @scosman also i think there might be scenarios where we want to force the non-trace version (in the UI) so we could match exact fine-tune data to match a previous dataset.

libs/core/kiln_ai/adapters/fine_tune/trace_based_dataset_formatter.py
```diff
@@ -0,0 +1,420 @@
...
+                        "function": {
+                            "name": "task_response",
+                            "arguments": last_message_content,  # pass a string directly
+                        },
+                    }
+                ],
+            },
+        )
+
+        return {"messages": messages}
```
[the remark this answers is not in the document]
chiang-daniel: This was following the logics from dataset_formatter.py

chiang-daniel:
this PR is no longer relevant. All changes in https://github.com/Kiln-AI/Kiln/pull/781

PR #767, opened 2025-10-31
Openrouter determinism with provider routing, adding a preferred provider list

tawnymanticore:
## What does this PR do?

Adding an order to the provider preferences taken from https://x.com/ArtificialAnlys/status/1955102409044398415 (simply used the list in the charts, roughly averaging rank of both equally).

here are the [docs](https://openrouter.ai/docs/features/provider-routing#ordering-specific-providers), which hint we should be using the slug (lowercase) rather than name (uppercase). which would imply a bug in previous routing preferences in this file (made that change in this PR as well)

and this is the endpoint with names/slugs: https://openrouter.ai/api/v1/providers
eg.
```
{"name":"Fireworks","slug":"fireworks","privacy_policy_url":"https://fireworks.ai/privacy-policy","terms_of_service_url":"https://fireworks.ai/terms-of-service","status_page_url":"https://status.fireworks.ai/"},{"name":"InferenceNet","slug":"inference-net","privacy_policy_url":"https://inference.net/privacy-policy","terms_of_service_url":"https://inference.net/terms-of-service","status_page_url":null},{"name":"DeepInfra","slug":"deepinfra","privacy_policy_url":"https://deepinfra.com/privacy","terms_of_service_url":"https://deepinfra.com/terms","status_page_url":"https://status.deepinfra.com/"},
```

## Checklists

- [X] New tests have been added to any work in /lib

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized provider routing preferences for OpenRouter model configurations.
  * Normalized provider ordering and exclusion settings across R1-related models for consistent fallback behavior.

* **Tests**
  * Added test coverage for OpenRouter provider routing configurations and fallback scenarios.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
@@ -435,9 +435,9 @@ class LiteLlmAdapter(BaseAdapter):
             # Require providers that support the reasoning parameter
             provider_options["require_parameters"] = True
             # Prefer R1 providers with reasonable perf/quants
-            provider_options["order"] = ["Fireworks", "Together"]
+            provider_options["order"] = ["fireworks", "together"]
             # R1 providers with unreasonable quants
-            provider_options["ignore"] = ["DeepInfra"]
+            provider_options["ignore"] = ["deepinfra"]
 
         # Only set of this request is to get logprobs.
         if (
@@ -447,7 +447,7 @@ class LiteLlmAdapter(BaseAdapter):
             # Don't let OpenRouter choose a provider that doesn't support logprobs.
             provider_options["require_parameters"] = True
             # DeepInfra silently fails to return logprobs consistently.
-            provider_options["ignore"] = ["DeepInfra"]
+            provider_options["ignore"] = ["deepinfra"]
 
         if provider.openrouter_skip_required_parameters:
             # Oddball case, R1 14/8/1.5B fail with this param, even though they support thinking params.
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -412,16 +412,32 @@ def build_extra_body(self, provider: KilnModelProvider) -> dict[str, Any]:
...
 
         if provider.r1_openrouter_options:
             # Require providers that support the reasoning parameter
             provider_options["require_parameters"] = True
             # Prefer R1 providers with reasonable perf/quants
-            provider_options["order"] = ["Fireworks", "Together"]
+            provider_options["order"] = ["fireworks", "together"]
             # R1 providers with unreasonable quants
-            provider_options["ignore"] = ["DeepInfra"]
+            provider_options["ignore"] = ["deepinfra"]
```
chiang-daniel, line 440: just curious why previous it wasn't lowercase?

tawnymanticore: I think a bug. API asks for `plug` rather than `name`. Names are uppercase, plugs are lowercase

leonardmq: Did you find a list of provider slugs in their docs by the way? I was looking at their docs the other day and could not find an actual list @tawnymanticore

leonardmq (approved):

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
tawnymanticore: @leonardmq yep, its right here:  https://openrouter.ai/api/v1/providers
its mega buried in this endpoint JSON

PR #782, opened 2025-11-07
ml_model_list update: Add MiniMax M2 and Kimi K2 Thinking

tawnymanticore:
## What does this PR do?

Adding the two latest thinking models.

`test_structured_input_cot_prompt_builder` Test failure

I do get the same failing test for both of these thinking models. not sure if it's a test issue actually.

K2 or M2  returns 3 messages instead of 5. With reasoning_capable=True, it uses SingleTurnR1ThinkingFormatter (3 messages) instead of TwoMessageCotFormatter (5 messages). 

 is the fix to add tuned_chat_strategy as a parameter to these models, or to change the test? I don't see any other models using that parameter so not sure if its deprecated or discouraged.

## Checklists

- [X] New tests have been added to any work in /lib

## Summary by CodeRabbit

* **New Features**
  * Kimi K2 Thinking model now available via Fireworks AI and OpenRouter providers
  * Minimax M2 model now available via OpenRouter and SiliconFlow CN providers
  * Both models support advanced reasoning capabilities and structured data generation

### DIFF
--- a/libs/core/kiln_ai/adapters/ml_model_list.py
+++ b/libs/core/kiln_ai/adapters/ml_model_list.py
@@ -4653,7 +4653,6 @@ built_in_models: List[KilnModel] = [
                 model_id="z-ai/glm-4.6:exacto",
                 structured_output_mode=StructuredOutputMode.json_instructions,
                 reasoning_capable=True,
-                reasoning_optional_for_structured_output=True,
             ),
             KilnModelProvider(
                 name=ModelProviderName.siliconflow_cn,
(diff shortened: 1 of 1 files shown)


tawnymanticore:
Test failure:

The test test_structured_input_cot_prompt_builder[kimi_k2_thinking-openrouter] fails because the trace has 3 messages instead of the expected 5.

Error:
assert len(trace) == 5E           assert 3 == 5

Root Cause:
Reasoning-capable models (reasoning_capable=True) with COT prompts use SingleTurnR1ThinkingFormatter, which produces 3 messages (system, user, assistant with reasoning + answer). The test expects TwoMessageCotFormatter, which produces 5 messages (system, user, assistant thinking, user follow-up, assistant final answer).

Context:
Model: kimi_k2_thinking with openrouter
Test: test_structured_input_cot_prompt_builder
Expected: 5 messages (two-message COT format)
Actual: 3 messages (single-turn R1 thinking format)

This is expected for reasoning-capable models; they use the single-turn format by design. The test may need to be updated to handle reasoning models, or this failure may be acceptable for these models. (???)

PR #786, opened 2025-11-10
refactor: off-loop and throttle PDF image conversion

leonardmq:
## What does this PR do?

Issue: some PDF manipulation operations are blocking the event loop. Splitting PDFs by page, converting PDFs into images. 

Changes:
- refactor: do PDF-to-image conversion in a process, and throttle to one process max with a semaphore

The semaphore is because we cannot spawn infinite processes or we run into other resource issues - e.g. `Too many open files`. The operation is relatively fast, so a queue of 1 is fine.

## Checklists

## Summary by CodeRabbit

* **Improvements**
  * PDF-to-image conversion now runs in separate worker processes to improve concurrency, throughput, and stability for multiple simultaneous requests.

* **Tests**
  * Added comprehensive concurrency and cleanup/error-handling tests to validate consistent results under varying loads and ensure proper resource cleanup.

### DIFF
--- a/libs/core/kiln_ai/utils/test_pdf_utils.py
+++ b/libs/core/kiln_ai/utils/test_pdf_utils.py
@@ -6,7 +6,11 @@ import pytest
 from pypdf import PdfReader
 
 from conftest import MockFileFactoryMimeType
-from kiln_ai.utils.pdf_utils import convert_pdf_to_images, split_pdf_into_pages
+from kiln_ai.utils.pdf_utils import (
+    _convert_pdf_to_images_sync,
+    convert_pdf_to_images,
+    split_pdf_into_pages,
+)
 
 
 async def test_split_pdf_into_pages_success(mock_file_factory):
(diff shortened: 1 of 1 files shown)


libs/core/kiln_ai/utils/test_pdf_utils.py
```diff
@@ -84,3 +89,45 @@ async def test_convert_pdf_to_images(mock_file_factory):
...
+    """Test running convert_pdf_to_images concurrently from multiple tasks."""
+    await run_convert_pdf_concurrently(mock_file_factory, concurrency=1)
+
+
+async def test_convert_pdf_to_images_concurrent_access_3(mock_file_factory):
+    """Test running convert_pdf_to_images concurrently from multiple tasks."""
+    await run_convert_pdf_concurrently(mock_file_factory, concurrency=3)
+
+
+def test__convert_pdf_to_images_sync(mock_file_factory):
```
scosman, line 120: nit: how slow are these?

Might want a new flag for slow tests. Still run them in CI, but not in dev_env? Not blocking

scosman (approved):

libs/core/kiln_ai/utils/test_pdf_utils.py
leonardmq: These are fast, but can add that in another PR - probably a good idea to have that as an option

scosman: We're actually pretty good. 1 1s test, but pretty reasonable. Not urgent.

`uv run python3 -m pytest --benchmark-quiet -q --durations=25 .`

PR #790, opened 2025-11-11
Infer OpenAI compatible provider from legacy model IDs

Helweg:
## What does this PR do?

- infer the OpenAI-compatible provider name when legacy runs only persist a model id
- update the adapter registry to use the inferred provider while keeping the canonical model id
- extend provider tool tests to cover the inference logic and align with suite conventions

I, @Helweg, confirm that I have read and agree to the [Contributors License Agreement](https://github.com/Kiln-AI/Kiln/blob/main/CLA.md).

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Enhanced OpenAI-compatible model ID parsing with support for multiple format specifications
  * Improved automatic provider name inference for simplified model configuration


leonardmq:
@Helweg - beautiful - thanks! We are going to check it out as soon as we can

PR #793, opened 2025-11-12
Improve form validators

scosman:
- Validate on value change, not simulated keyboard events
- Use a context to register form elements with form
- Nicer error placement, so it works visually when there isn’t a subtitle

Benefit: Validators now work on bool & enums

## Summary by CodeRabbit

* **New Features**
  * Centralized form validation now coordinates validation across form fields for more consistent feedback.
  * Error state styling extended to select controls (visible error outline).
  * Info tooltip gains an alternate symbol option (exclaim) to change its visual cue.

* **Bug Fixes / UX**
  * Error messages are now shown as a label-level badge in addition to inline errors for clearer visibility.

### DIFF
--- a/app/web_ui/src/lib/utils/form_element.svelte
+++ b/app/web_ui/src/lib/utils/form_element.svelte
@@ -95,8 +95,11 @@
   }
 
   // run validator after value change
+  let initialized = false
   function run_validator_on_change(_: unknown) {
-    run_validator()
+    if (initialized) {
+      run_validator()
+    }
   }
   $: run_validator_on_change(value)
 
@@ -109,6 +112,7 @@
   let unregister: (() => void) | null = null
 
   onMount(() => {
+    initialized = true
     if (formContainer) {
       unregister = formContainer.registerFormElement({ run_validator })
     }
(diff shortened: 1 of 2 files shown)


leonardmq:
It seems like the initial state on mount is in error state - on first load (both after internal nav or hard refresh), all the forms I see are in error state:

image

image

image

app/web_ui/src/lib/utils/form_element.svelte
```diff
@@ -88,11 +89,37 @@
...
 
-  function run_validator() {
+  export function run_validator() {
     const error = validator(value)
     error_message = error
   }
 
+  // run validator after value change
+  function run_validator_on_change(_: unknown) {
+    run_validator()
```
leonardmq: This seems to be running on mount - other comment with screenshots: https://github.com/Kiln-AI/Kiln/pull/793#issuecomment-3520180409

scosman: Fixed in 4d736fb144d85ca0e3c9511ff47d42dcf09f1352

PR #802, opened 2025-11-13
fix: generate Q&A in action dropdown menu

leonardmq:
## What does this PR do?

1. Hide `Generate Q&A Pairs` in the `...` menu until after extraction has been run; also hide it on docs that failed to extract
2. Show warning dialog when clicking `Generate Q&A Pairs` on a Document node that has chunks - to warn the user that the chunks down that document are going to be replaced and all associated Q&A pairs lost

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Introduced a confirmation warning dialog when generating Q&A pairs for entire documents, alerting users about potential data loss and chunk replacement impacts.
  * Updated the Generate Q&A menu item visibility to display only when appropriate document configuration is present and the document is in a valid state.

### DIFF
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte
@@ -4,6 +4,7 @@
   import Output from "../../../../run/output.svelte"
   import { createEventDispatcher } from "svelte"
   import type { QnaStore } from "./qna_ui_store"
+  import Warning from "$lib/ui/warning.svelte"
 
   type QnAPair = {
     id: string
@@ -49,6 +50,7 @@
   let part_output_dialog: Dialog | null = null
   let selected_part_text: string | null = null
   let document_parts_dialog: Dialog | null = null
+  let generate_document_warning_dialog: Dialog | null = null
 
   function toggleQAPairExpand(qaId: string) {
     expandedQAPairs[qaId] = !expandedQAPairs[qaId]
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte
```diff
@@ -131,14 +145,13 @@
...
         </li>
-        <li>
-          <button
-            on:click|stopPropagation={() =>
-              dispatch("generate_for_document", { document_id: document.id })}
-          >
-            Generate Q&A Pairs
-          </button>
-        </li>
+        {#if $qnaMaxStep && $qnaMaxStep > 2 && !document.extraction_failed}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte:148]: It would have been much nicer to name these steps instead of using the numbers - this works though

Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte:148]: is it possible to be on the 3rd step but had extraction failed? how do you get out of that state right now?

Comment (leonardmq) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte:148]: It should not - if you have one or more extractions that fail:
- the docs that failed get a badge that says `Not extracted`
- next step is blocked
- you get a nudge at the top saying "Some extractions failed, either remove the documents that failed (which they can do with the `...` menu) or retry" (and the "Run extraction" button becomes "Retry")

Then once all the docs that are there have been extracted, we unlock then next step

Comment (sfierro) [reply app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/qna/qna_document_node.svelte:148]: okie

PR #807, opened 2025-11-13
fix: semantic chunker and concurrency

leonardmq:
## What does this PR do?

Issue: we pass our own implementation of the embedding model to LlamaIndex in the Semantic Chunker, but LlamaIndex makes a very high number of concurrent calls to the embedding API. Causes meltdown and `Too many open files`.

Changes:
- add a semaphore in the LlamaIndex embedding adapter that we inject into it, to cap the number of concurrent calls to our embedding adapter at `10` max concurrent

## Checklists

## Summary by CodeRabbit

* **Performance**
  * Improved concurrency handling for embedding operations to enhance stability when processing multiple simultaneous requests, preventing resource contention and optimizing system throughput.

### DIFF
--- a/libs/core/kiln_ai/adapters/chunkers/embedding_wrapper.py
+++ b/libs/core/kiln_ai/adapters/chunkers/embedding_wrapper.py
@@ -1,3 +1,4 @@
+import asyncio
 from typing import List
 
 from llama_index.core.embeddings import BaseEmbedding
@@ -10,6 +11,7 @@ class KilnEmbeddingWrapper(BaseEmbedding):
 
     def __init__(self, embedding_adapter: BaseEmbeddingAdapter):
         super().__init__()
+        self._semaphore = asyncio.Semaphore(10)
         self._embedding_adapter = embedding_adapter
 
     def _get_query_embedding(self, query: str) -> List[float]:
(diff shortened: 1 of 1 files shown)


libs/core/kiln_ai/adapters/chunkers/embedding_wrapper.py
leonardmq, line 9: This class is only used by the LlamaIndex semantic chunker

libs/core/kiln_ai/adapters/chunkers/embedding_wrapper.py
```diff
@@ -10,6 +11,7 @@ class KilnEmbeddingWrapper(BaseEmbedding):
 
     def __init__(self, embedding_adapter: BaseEmbeddingAdapter):
         super().__init__()
+        self._semaphore = asyncio.Semaphore(10)
```
leonardmq, line 14: Specific number is arbitrary, but this is low enough that it will stay below rate limits for sure even with long documents where we may hit the API a lot continuously over a long period of time

chiang-daniel: probably worth a comment saying this is arbitrary?

PR #823, opened 2025-11-17
Show reference answer in eval run UI

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Added support for reference answer evaluations, allowing reference answers to be captured and displayed alongside outputs in the run results interface.

* **Tests**
  * Added validation tests for reference answer handling across different evaluation types.

### DIFF
--- a/libs/core/kiln_ai/datamodel/eval.py
+++ b/libs/core/kiln_ai/datamodel/eval.py
@@ -240,7 +240,7 @@ class EvalRun(KilnParentedModel):
             and evaluation_data_type != EvalDataType.reference_answer
         ):
             raise ValueError(
-                f"reference_answer is only valid for reference answer evals. Got: {evaluation_data_type}"
+                f"reference_answer is only valid for reference answer evals. Got: {evaluation_data_type.value}"
             )
         return self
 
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte
```diff
@@ -198,6 +198,12 @@
                 <div>
                   {result.input}
                 </div>
+                {#if result.reference_answer}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte:201]: If we know the type of eval in this page, we could show `null` if there is none - could possibly help catch bugs (e.g. wrong dataset with no reference answers at all if that is possible)

Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte:201]: I was thinking that but was worried about people that created ref ans evals before this change they'll have null too

Comment (sfierro) [reply app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte:201]: I think I just thought that if it doesn't show up that's the same as null with regards to bug catching (and no it isn't possible to point to the wrong dataset, they're just normal data points so we'd take the output of the saved run and treat it as reference)

PR #827, opened 2025-11-17
Spec data model class

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Add Spec entities (name, definition, type, priority, status, tags, optional eval link), Task→Specs relation, CRUD API endpoints, server wiring, and UI/type exports.
* **Bug Fixes / Validation**
  * Enforce tag rules: no empty tags and no spaces within tags.
* **Tests**
  * Comprehensive unit and API tests for creation, defaults, validation, listing, retrieval, and updates.
* **Chores**
  * Run config fields now use enum types for stricter typing.

### DIFF
--- a/libs/server/kiln_server/test_spec_api.py
+++ b/libs/server/kiln_server/test_spec_api.py
@@ -757,6 +757,65 @@ def test_create_spec_invalid_tags_type(client, project_and_task):
     assert any(error["loc"] == ["body", "tags"] for error in res["source_errors"])
 
 
+def test_create_spec_empty_string_in_tags(client, project_and_task):
+    project, task = project_and_task
+
+    spec_data = {
+        "name": "Test Spec",
+        "definition": "The system should always respond politely",
+        "type": SpecType.desired_behaviour.value,
+        "priority": Priority.p1,
+        "status": SpecStatus.active.value,
+        "tags": [""],
+        "eval_id": None,
+    }
+
+    with patch("kiln_server.spec_api.task_from_id") as mock_task_from_id:
+        mock_task_from_id.return_value = task
+        response = client.post(
+            f"/api/projects/{project.id}/tasks/{task.id}/spec", json=spec_data
+        )
+
+    assert response.status_code == 422
+    res = response.json()
+    assert "source_errors" in res
+    assert any(
+        "empty strings" in error.get("msg", "").lower()
+        for error in res["source_errors"]
+    )
+
+
+def test_create_spec_tag_with_space(client, project_and_task):
+    project, task = project_and_task
+
+    spec_data = {
+        "name": "Test Spec",
+        "definition": "The system should always respond politely",
+        "type": SpecType.desired_behaviour.value,
+        "priority": Priority.p1,
+        "status": SpecStatus.active.value,
+        "tags": ["tag with space"],
+        "eval_id": None,
+    }
+
+    with patch("kiln_server.spec_api.task_from_id") as mock_task_from_id:
+        mock_task_from_id.return_value = task
+        response = client.post(
+            f"/api/projects/{project.id}/tasks/{task.id}/spec", json=spec_data
+        )
+
+    assert response.status_code == 422
+    res = response.json()
+    assert "source_errors" in res
+    assert any(
+        (
+            "spaces" in error.get("msg", "").lower()
+            or "underscores" in error.get("msg", "").lower()
+        )
+        for error in res["source_errors"]
+    )
+
+
 def test_update_spec_missing_required_fields(client, project_and_task):
     project, task = project_and_task
 


libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+    description: str | None = None
+    definition: str | None = None
+    priority: SpecPriority | None = None
+    status: SpecStatus | None = None
+    tags: List[str] | None = None
+    eval_id: str | None = None
+
+
+def connect_spec_api(app: FastAPI):
+    @app.post("/api/projects/{project_id}/tasks/{task_id}/spec")
```
leonardmq, line 35: Why not `/specs` (with an `s`) for consistency with the ones below?

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+            spec_kwargs["eval_id"] = spec_data.eval_id
+
+        spec = Spec(**spec_kwargs)
+        spec.save_to_file()
+        return spec
+
+    @app.get("/api/projects/{project_id}/tasks/{task_id}/specs")
+    async def get_specs(project_id: str, task_id: str) -> List[Spec]:
+        parent_task = task_from_id(project_id, task_id)
+        return parent_task.specs()
```
leonardmq: Can do `readonly=True` - this prevents fetching from FS if it was already loaded (we use readonly for reads, but not for updates to ensure we have the latest)

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+
+    @app.patch("/api/projects/{project_id}/tasks/{task_id}/specs/{spec_id}")
+    async def update_spec(
+        project_id: str, task_id: str, spec_id: str, spec_updates: SpecUpdateRequest
+    ) -> Spec:
+        spec = spec_from_id(project_id, task_id, spec_id)
+
+        update_dict = spec_updates.model_dump(exclude_unset=True)
+        for key, value in update_dict.items():
+            setattr(spec, key, value)
```
leonardmq: This has no typechecking in case of request schema and datamodel drift so double checked how this behaves - seems like risk is limited:
- request drops unknown fields
- setattr still runs Pydantic validation
- if request defines an extra field (e.g. `c: str`) and that field is not in the datamodel, `setattr` will drop it silently; maybe worth throwing explicitly in here if `key not in Spec.model_fields` to explicitly catch drift, but not convinced it would be that much better than silently dropping the wrong field (and catching with tests)

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+        return spec
+
+    raise HTTPException(
+        status_code=404,
+        detail=f"Spec not found. ID: {spec_id}",
+    )
+
+
+class SpecCreateRequest(BaseModel):
+    name: str
```
leonardmq: Could use `FilenameString` here to directly reject? If `str` here, it will still be rejected but at the `setattr` for that key

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+    definition: str
+    type: SpecType
+    priority: Optional[SpecPriority] = None
+    status: Optional[SpecStatus] = None
+    tags: Optional[List[str]] = None
+    eval_id: Optional[str] = None
+
+
+class SpecUpdateRequest(BaseModel):
+    name: str | None = None
```
leonardmq: Could use `FilenameString` here to directly reject? If `str` here, it will still be rejected but at the `setattr` for that key

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def connect_spec_api(app: FastAPI):
+    @app.post("/api/projects/{project_id}/tasks/{task_id}/spec")
+    async def create_spec(
+        project_id: str, task_id: str, spec_data: SpecCreateRequest
+    ) -> Spec:
+        parent_task = task_from_id(project_id, task_id)
+
+        spec_kwargs = {
```
leonardmq: You probably do not need this intermediate dict - you can instantiate Spec and then set the optional fields. This will also avoid having this untyped dict that won't get caught by typechecking if it drifts from what `Spec(**kwargs)` expects). 

Something Like:
```py
@app.post("/api/projects/{project_id}/tasks/{task_id}/spec")
async def create_spec(
    project_id: str, task_id: str, spec_data: SpecCreateRequest
) -> Spec:
    parent_task = task_from_id(project_id, task_id)

    spec = Spec(
        parent=parent_task,
        name=spec_data.name,
        description=spec_data.description,
        definition=spec_data.definition,
        type=spec_data.type,
        eval_id=spec_data.eval_id,
    )

    if spec_data.priority is not None:
        spec.priority = spec_data.priority
    if spec_data.status is not None:
        spec.status = spec_data.status
    if spec_data.tags is not None:
        spec.tags = spec_data.tags

    spec.save_to_file()
    return spec
```

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,90 @@
...
+            "type": spec_data.type,
+        }
+        if spec_data.priority is not None:
+            spec_kwargs["priority"] = spec_data.priority
+        if spec_data.status is not None:
+            spec_kwargs["status"] = spec_data.status
+        if spec_data.tags is not None:
+            spec_kwargs["tags"] = spec_data.tags
+        if spec_data.eval_id is not None:
+            spec_kwargs["eval_id"] = spec_data.eval_id
```
leonardmq: Seems like `eval_id` can be `None` on the datamodel so could be passed in with the non-optional properties

libs/core/kiln_ai/datamodel/spec.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    status: SpecStatus = Field(
+        default=SpecStatus.not_started,
+        description="The status of the spec.",
+    )
+    tags: List[str] = Field(
+        default=[],
+        description="The tags of the spec.",
+    )
+    eval_id: ID_TYPE | None = Field(
```
leonardmq, line 75: Not clear on this myself, but Steve seems to be saying a `Spec` should be able to have many `Eval` and not 1:1

libs/server/kiln_server/spec_api.py
sfierro: its creating a singular spec though, we do this convention elsewhere too

libs/server/kiln_server/spec_api.py
sfierro: oops yeah this was Claude, changing to how we do it elsewhere

libs/server/kiln_server/spec_api.py
sfierro: oops yeah once again changing to how we do it elsewhere

libs/server/kiln_server/spec_api.py
sfierro: fixed by changing the creation request

libs/core/kiln_ai/datamodel/spec.py
sfierro: I think he confirmed over slack its just 1:1 after all (just many judges)

libs/server/kiln_server/spec_api.py
```diff
@@ -0,0 +1,77 @@
...
+    ) -> Spec:
+        spec = spec_from_id(project_id, task_id, spec_id)
+
+        spec.name = spec_data.name
+        spec.definition = spec_data.definition
+        spec.type = spec_data.type
+        spec.priority = spec_data.priority
+        spec.status = spec_data.status
+        spec.tags = spec_data.tags
+        spec.eval_id = spec_data.eval_id
```
leonardmq, line 74: Is there a use case for replacing the eval id you think?

libs/server/kiln_server/spec_api.py
leonardmq: Not sure this is really a convention since we are not doing it consistently - e.g. `POST /finetunes` or `POST /create_xyz` RPC style naming.

We should probably settle for one scheme and use it everywhere - maybe separate later PR doing just renaming. I would suggest going with standard REST style names where 1:m is plural even if the endpoint creates only one resource in that collection

libs/server/kiln_server/spec_api.py
sfierro: not sure yet, think its fine to keep like this for now though

PR #831, opened 2025-11-18
Fixed link nav from compare screen

sfierro:
This link was going to the compare judges screen which isn't supported for RAG. For RAG evals make the link go to the compare run configs screen which handles judge creation there!

image_720

## Summary by CodeRabbit

* **New Features**
  * Smarter "Manage Eval Configs" routing with dynamic navigation that shows loading or error states and routes based on template type (e.g., RAG vs. eval), improving workflow clarity.

* **Improvements**
  * Evaluation template caching and loading indicators to reduce redundant fetches, speed up responsiveness, and drive dynamic statuses in the eval comparison view.

### DIFF
--- a/app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
+++ b/app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
@@ -704,14 +704,24 @@
                   <div class="px-6 py-4 text-center">
                     {#if section.has_default_eval_config === false}
                       <div>Select a default eval config to compare scores.</div>
-                      <a
-                        href={eval_templates_cache[section.eval_id] === "rag"
-                          ? `/evals/${project_id}/${task_id}/${section.eval_id}/compare_run_configs`
-                          : `/evals/${project_id}/${task_id}/${section.eval_id}/eval_configs`}
-                        class="btn btn-xs rounded-full"
-                      >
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -228,12 +233,62 @@
...
   $: comparisonFeatures = generateComparisonFeatures(
     selectedModels,
     eval_scores_cache,
   )
 
+  // Reactively fetch eval templates for sections
+  $: {
+    comparisonFeatures.forEach((section) => {
+      if (
+        section.eval_id &&
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:282]: Typing seems to say `section.eval_id` is never nullish (and probably never null string?) so possibly don't need this check?

app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -228,12 +233,62 @@
...
+        section.eval_id !== "kiln_cost_section" &&
+        eval_templates_cache[section.eval_id] === undefined &&
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/evals/[project_id]/[task_id]/compare/+page.svelte:284]: Could possibly remove the duplicative cache / loading checks here since we do it in `fetch_eval_template` - would make this a bit lighter to reason about

PR #834, opened 2025-11-18
Sortable spec columns

sfierro:
https://github.com/user-attachments/assets/7d272615-68f1-44f7-a14b-97b7465d219b

## Summary by CodeRabbit

* **New Features**
  * Added interactive sorting to the specs table. Click any column header (Name, Type, Priority, Status, Created At) to sort; active column shows ▲/▼; click again reverses direction.

* **UI Improvements**
  * Improved display formatting for spec Type and Priority for clearer, more readable labels.


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -48,6 +59,61 @@
   function formatPriority(priority: number): string {
     return `P${priority}`
   }
+
+  function getStatusSortOrder(status: string): number {
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte]: Why not use `SpecStatus` type here? You already have it in the $lib/types.ts

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -48,6 +59,61 @@
...
+    if (status === "active") return 0
+    if (status === "future") return 1
+    if (status === "deprecated") return 2
+    return 3 // should never happen, but just in case
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte]: This already seems to happen if the status is `archived`. 

You can catch unhandled cases like this with the typechecker doing something like:
```ts
// where SpecStatus is type: 'active' | 'future' | 'deprecated' | 'archived'
function getStatusSortOrder(status: SpecStatus): number {
    switch (status) {
        case "active":
            return 0
        case "future":
            return 1
        case "deprecated":
            return 2
        case "archived":
            return 3
        default: {
            // if we ever add new possibilities that are missing above, the `never` will be wrong
            // and typechecker will show us
            const _: never = status
            return 4
        }
    }
}   
```

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -91,21 +157,81 @@
...
+                on:click={() => handleSort("name")}
+                class="hover:bg-base-200 cursor-pointer"
+              >
+                Name
+                <span class="inline-block w-3 text-center">
+                  {sortColumn === "name"
+                    ? sortDirection === "asc"
+                      ? "▲"
+                      : "▼"
+                    : "\u200B"}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:186]: What does the `\u200B` do (as opposed to an empty string)?

leonardmq (approved):

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte]: oops yes I should!

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte]: thank you!

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:186]: from Claude:

Difference between "\u200B" and an empty string "":
"\u200B" (zero-width space):
Invisible, but still content
The browser treats it as content, so the element maintains its dimensions
With inline-block w-3, the element keeps its width because there's content (even if invisible)
More reliable for maintaining layout structure
"" (empty string):
No content at all
The element might collapse or not render consistently
Even with w-3, some browsers might not render the element properly without content
Can cause layout shifts or inconsistent spacing
In your case, with the inline-block w-3 class, the width should be enforced, but using \u200B ensures:
The element is treated as having content
Consistent rendering across browsers
The layout structure is maintained
If you used "", the span might still work due to the w-3 class, but \u200B is more reliable for maintaining the intended layout, especially when the sort arrow isn't shown. It's a common pattern for maintaining consistent spacing in UI layouts.

PR #841, opened 2025-11-21
Added spec properties property and changed definition to description

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Specs now use a structured properties object with many property types (behaviour, tone, formatting, localization, safety, tooling, etc.) for richer guidance.

* **Changes**
  * "Definition" renamed to "Description" across UI and API; create/edit flows now submit properties instead of a flat type field.
  * Type display and sorting now derive from properties.spec_type.
  * Some spec templates consolidated/renamed (behaviour) and others removed.

* **Tests**
  * Expanded tests for properties validation, creation, update, and error cases.


libs/core/kiln_ai/datamodel/spec.py
```diff
@@ -56,8 +57,10 @@ class Spec(KilnParentedModel):
...
 
     name: FilenameString = Field(description="The name of the spec.", min_length=1)
-    definition: str = Field(
-        description="A detailed definition of the spec.", min_length=1
+    description: str = Field(description="A description of the spec.", min_length=1)
+    properties: SpecProperties | None = Field(
+        default=None,
+        description="The properties of the spec.",
     )
     type: SpecType = Field(
```
leonardmq: If you need a union of `properties`, you can remove the `type` here in the root. And just keep `spec_type` inside the `properties`. 

To narrow down the types, you need to discriminate on `model.properties["type"] == "xyz"` for the typechecker to know (checking `if model.type == xyz` does not make it narrow down the types) - see the other comment with examples.

In some of the models where I used discriminated union, I have it in both the root model and inside `properties`, but only for backward compatibility (because existing users may already have files on disk where the type is in the root)

libs/core/kiln_ai/datamodel/spec.py
```diff
@@ -56,8 +57,10 @@ class Spec(KilnParentedModel):
     """A spec for a task."""
 
     name: FilenameString = Field(description="The name of the spec.", min_length=1)
-    definition: str = Field(
-        description="A detailed definition of the spec.", min_length=1
+    description: str = Field(description="A description of the spec.", min_length=1)
+    properties: SpecProperties | None = Field(
```
leonardmq: You need to pass in the `discriminator="spec_type"` into the `Field` (see the `ChunkerConfig` as an example).

The idea is to tell Pydantic that the type of this field (`properties`) is conditioned on the value of `spec_type` (inside it).

Then when you do this:
```py
if model.properties["spec_type"] == "appropriate_tool_use":
    # here you get the correct type out of the box, without needing to cast or validate again
```

Pydantic also uses that info when you create / write an object; if you do something like:
```py
model = Spec(
    name = "x",
    # ...
    properties = {
        "spec_type": "appropriate_tool_use",
        # ...
    }
)
```

Pydantic will automatically use the associate type of `properties` and will validate according to that specific dict. So this really helps get rid of a lot of manual validation.

If I recall, you also get discriminated union in the frontend. So in the UI you also get automatic type inference when doing:
```ts
const model = fetch("/api/xxx/somemodel")
if (model.properties.spec_type === "xyz) {
    // here typescript narrows it down to the associated type
}
```

If you need to validate within a subtype of `properties` (e.g.  if `AppropriateToolUseProperties` has some internal logic like the `tool_id` must match some arbitrary logic); you can use the `Annotated` and `AfterValidator` and then use that as your type; like here:
https://github.com/Kiln-AI/Kiln/blob/4cdddfcabd0323dde03c8063760665914aa8ea74/libs/core/kiln_ai/datamodel/chunk.py#L81

This creates a wrapped type that runs the validator (in the same way a `@model_validator("after")` would do, on the `properties` `TypedDict`); and then you define your properties to be of this type; like here:
https://github.com/Kiln-AI/Kiln/blob/4cdddfcabd0323dde03c8063760665914aa8ea74/libs/core/kiln_ai/datamodel/chunk.py#L103

libs/core/kiln_ai/datamodel/spec.py
```diff
@@ -79,6 +82,50 @@ class Spec(KilnParentedModel):
...
+        """Validate that properties match the spec type when provided."""
+        if self.properties is None:
+            return self
+
+        if self.type.value != self.properties["spec_type"]:
+            raise ValueError(
+                f"Spec type mismatch: {self.type.value} != {self.properties['spec_type']}."
+            )
+
+        if self.properties["spec_type"] == "appropriate_tool_use":
```
leonardmq: This is the kind of thing that can go away if you do discriminated union. Pydantic does essentially this internally and automatically (according to the corresponding `TypedDict`)

libs/core/kiln_ai/datamodel/spec_properties.py
```diff
@@ -0,0 +1,19 @@
...
+    inappropriate_tool_use_guidelines: str | None
+
+
+class UndesiredBehaviourProperties(TypedDict, total=True):
+    spec_type: Literal["undesired_behaviour"]
+    undesired_behaviour_guidelines: str
+    examples: str
+
+
+SpecProperties = AppropriateToolUseProperties | UndesiredBehaviourProperties
```
leonardmq: Confirming this file is already ok wrt discriminated union - just need to wire up the pieces on the `Spec` model

libs/server/kiln_server/spec_api.py
```diff
@@ -23,7 +24,8 @@ def spec_from_id(project_id: str, task_id: str, spec_id: str) -> Spec:
 
 class SpecUpsertRequest(BaseModel):
     name: FilenameString
-    definition: str
+    description: str
+    properties: SpecProperties | None
```
leonardmq: You could also do discriminated union here (would only need to add the `Field(discriminator="spec_type")` in here) - I think that will make your request in the frontend also get typechecked (i.e. not allow passing in the wrong properties given the `spec_type` in `properties`)

In the chunker API, I did this - don't really remember if that is any good or why, but probably ok pattern:
https://github.com/Kiln-AI/Kiln/blob/4cdddfcabd0323dde03c8063760665914aa8ea74/libs/server/kiln_server/document_api.py#L360

Essentially just having a getter that returns the union of possible types (I think it is because the typechecker in Python does not consider valid to try to shove an object of type `A` into an `A | B`, it wants an explicitly `A | B` union type; or something along these lines)

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -464,12 +465,13 @@
...
               name: spec.name,
-              definition: spec.definition,
+              description: spec.description,
+              properties: spec.properties ?? null,
               type: spec.type,
               priority: spec.priority,
               status: new_status,
               tags: spec.tags,
-              eval_id: spec.eval_id || null,
+              eval_id: spec.eval_id ?? null,
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:472]: Not about this line:
- line 523: empty state links to the `/create_spec` page (not the one where you get to pick a template)

https://github.com/Kiln-AI/Kiln/blob/4cdddfcabd0323dde03c8063760665914aa8ea74/app/web_ui/src/routes/(app)/specs/%5Bproject_id%5D/%5Btask_id%5D/%2Bpage.svelte#L533

Not sure if intentional

Comment (leonardmq) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:472]: Just saw you fixed the empty state in the next one: https://github.com/Kiln-AI/Kiln/pull/842

libs/core/kiln_ai/datamodel/spec.py
sfierro: I was thinking not all spec types need properties though, but let me look at your other comments and maybe this will make more sense :D

libs/server/kiln_server/spec_api.py
sfierro: don't think I need the get_properties_for_chunker_type as the all properties will be defined, it looks like you had that to put some default values in for the client. but did change this to the discriminator one

libs/core/kiln_ai/datamodel/spec.py
```diff
@@ -135,3 +57,13 @@ def validate_tags(self) -> Self:
                 raise ValueError("tags cannot contain spaces. Try underscores.")
 
         return self
+
+    @model_validator(mode="after")
+    def validate_type_matches_properties(self) -> Self:
+        if self.properties is not None:
+            properties_type = self.properties.get("spec_type")
+            if properties_type != self.type:
```
leonardmq: Why not have `properties: { type: xyz }` and remove root `type` instead of letting the properties be null? 

Having `properties` optional will force null checks in many places, break some of the automatic type discrimination in some places (don't remember how, but I remember optional on `properties` was breaking some typing and validation), and forces you to maintain that root type that could just be in `properties` and get you the automatic validation.

If you remove the root `type: SpecType` and just have `properties: { type }`, this check will be done automatically by Pydantic. Only problem is backward compatibility, but this model has not been used publicly so you can just delete the local datamodels you might have

leonardmq (approved):

sfierro:
Closing for now as the spec model object is in flux we might not do properties at all

PR #842, opened 2025-11-21
Sfierro/spec forms

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Introduced type-specific forms for creating different specification types (Undesired Behaviour, Tool Use, Reference Answer Accuracy) for improved clarity and guidance.

* **Bug Fixes**
  * Fixed form validation to properly detect and display errors for dropdown/select fields.

* **Improvements**
  * Updated spec creation workflow navigation paths.
  * Clarified table column header labeling.

### DIFF
--- a/app/web_ui/src/lib/api_schema.d.ts
+++ b/app/web_ui/src/lib/api_schema.d.ts
@@ -2176,8 +2176,8 @@ export interface components {
         /** AppropriateToolUseProperties */
         AppropriateToolUseProperties: {
             /**
-             * Spec Type
-             * @constant
+             * @description discriminator enum property added by openapi-typescript
+             * @enum {string}
              */
             spec_type: "appropriate_tool_use";
             /** Tool Id */
@@ -5198,13 +5198,13 @@ export interface components {
              * @description A description of the spec.
              */
             description: string;
+            /** @description The type of spec. */
+            type: components["schemas"]["SpecType"];
             /**
              * Properties
              * @description The properties of the spec.
              */
-            properties?: components["schemas"]["AppropriateToolUseProperties"] | components["schemas"]["UndesiredBehaviourProperties"] | null;
-            /** @description The type of spec. */
-            type: components["schemas"]["SpecType"];
+            properties?: (components["schemas"]["AppropriateToolUseProperties"] | components["schemas"]["UndesiredBehaviourProperties"]) | null;
             /**
              * @description The priority of the spec.
              * @default 1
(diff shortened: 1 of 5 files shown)


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/undesired_behaviour_form.svelte
```diff
@@ -0,0 +1,34 @@
...
+  <svelte:fragment slot="form_elements">
+    <FormElement
+      label="Undesired Behaviour"
+      description="A detailed description of the undesired behaviour you want to catch. This will be used by AI to understand the spec."
+      id="undesired_behaviour"
+      inputType="textarea"
+      bind:value={undesired_behaviour}
+    />
+    <FormElement
+      label="Examples"
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/undesired_behaviour_form.svelte:27]: Not about this PR specifically:

For this kind of fields, I am never really sure if a certain format is expected. I guess the answer is no; but as a user you wonder what style of writing you should even go with. Like should I describe the problem in abstract terms, or give something more narrow like `input: xyz; output: zzz; I don't like it because xxx`, should I be precise or not, should I say "You should do XYZ" or "I am a model and should not do XYZ", etc.

Vague idea - something like a placeholder with different genres of inputs (more or less in the idea of when you open ChatGPT and it shows you examples of questions you might ask it like: "How do I buy a car?")

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/create_spec_form.svelte
```diff
@@ -0,0 +1,87 @@
...
+  import { createKilnError, KilnError } from "$lib/utils/error_handlers"
+  import type { SpecProperties, SpecType } from "$lib/types"
+  import { goto } from "$app/navigation"
+  import { page } from "$app/stores"
+  import { client } from "$lib/api_client"
+  import FormElement from "$lib/utils/form_element.svelte"
+
+  export let spec_type: SpecType
+  export let form_has_unsaved_changes: boolean = false
+  export let get_properties: (() => SpecProperties) | undefined = undefined
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/create_spec_form.svelte:12]: Why not pass the `SpecProperties` as props instead of a callback?

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/create_spec_form.svelte:12]: because until this is called, the properties won't be valid (all fields filled out) so the form needs to know when to create it finally

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/undesired_behaviour_form.svelte
Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/create_spec/forms/undesired_behaviour_form.svelte:27]: I agree, we will polish later!

sfierro:
Closing because the plan is changing.

PR #861, opened 2025-11-27
KIL-106: Add subtasks cost to Usage property list

sfierro:
## Summary
Adds "Subtasks Cost" to the Usage property list when viewing runs that used Kiln task tools. This shows the total cost of all subtask runs invoked via Kiln Task Tools.

Screenshot 2025-11-27 at 3 41 04 AM

## Changes
- Added helper function to extract `kiln_task_tool_data` from trace messages
- Added recursive function to fetch subtask runs and calculate total cost (handles nested subtasks)
- Display "Subtasks Cost" in Usage section when subtask costs exist (> $0)
- Show loading state while fetching subtask run data

## How it works
When a task uses a Kiln Task Tool, the tool response message contains `kiln_task_tool_data` with the format `{project_id}:::{tool_id}:::{task_id}:::{run_id}`. This PR parses these references from the trace, fetches the subtask runs, and sums their costs recursively (to handle nested task tools calling other task tools).

## UI Result
The Usage property list will now show (when applicable):
- Cost: $0.000528
- **Subtasks Cost: $0.001234** ← New line
- Tokens: 1234

Closes KIL-106

## Summary by CodeRabbit

* **New Features**
  * Subtask costs (including nested subtasks) are now calculated and shown in the run view.
  * Usage display extended to include subtask cost and subtask token totals.
  * Loading indicators and request deduplication prevent duplicate calculations and show progress while subtask usage is computed.

### DIFF
--- a/app/web_ui/src/routes/(app)/run/run.svelte
+++ b/app/web_ui/src/routes/(app)/run/run.svelte
@@ -53,14 +53,15 @@
     return references
   }
 
-  async function calculate_subtask_cost(
+  async function calculate_subtask_usage(
     trace: Trace | null | undefined,
     visited: Set<string> = new Set(),
-  ): Promise<number> {
-    if (!trace) return 0
+  ): Promise<{ cost: number; tokens: number }> {
+    if (!trace) return { cost: 0, tokens: 0 }
 
     const references = extract_subtask_references(trace)
     let total_cost = 0
+    let total_tokens = 0
 
     for (const ref of references) {
       const key = `${ref.project_id}:${ref.task_id}:${ref.run_id}`
@@ -83,14 +84,17 @@
 
         if (!response.error && response.data) {
           total_cost += response.data.usage?.cost ?? 0
-          total_cost += await calculate_subtask_cost(
+          total_tokens += response.data.usage?.total_tokens ?? 0
+          const subtask_usage = await calculate_subtask_usage(
             response.data.trace,
             visited,
           )
+          total_cost += subtask_usage.cost
+          total_tokens += subtask_usage.tokens
         }
       } catch (error) {
         console.warn(
-          "Failed to fetch subtask cost, continuing on error as subtask run may have been deleted.",
+          "Failed to fetch subtask usage, continuing on error as subtask run may have been deleted.",
           {
             project_id: ref.project_id,
             task_id: ref.task_id,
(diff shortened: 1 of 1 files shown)


app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -28,6 +29,92 @@
...
+            visited,
+          )
+        }
+      } catch {
+        // Continue on error - subtask run may have been deleted
+      }
+    }
+
+    return total_cost
+  }
```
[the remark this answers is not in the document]
sfierro: Good catch! Fixed in 49b3bcc3 - added a request-id guard to prevent stale async results from overwriting current state when the user switches runs quickly.

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -28,6 +29,105 @@
...
+
+        if (!response.error && response.data) {
+          total_cost += response.data.usage?.cost ?? 0
+          total_cost += await calculate_subtask_cost(
+            response.data.trace,
+            visited,
+          )
+        }
+      } catch {
+        // Continue on error - subtask run may have been deleted
```
leonardmq: Probably still worth at least logging the error?

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -28,6 +29,105 @@
...
+      }
+    }
+
+    return total_cost
+  }
+
+  let subtask_cost: number | null = null
+  let subtask_cost_loading = false
+  let subtask_cost_request_id = 0
```
leonardmq: What is this for? When run changes? I don't follow the `request_id` and `subtask_cost_request_id`

app/web_ui/src/routes/(app)/run/run.svelte
```diff
@@ -28,6 +29,105 @@
...
+                project_id: ref.project_id,
+                task_id: ref.task_id,
+                run_id: ref.run_id,
+              },
+            },
+          },
+        )
+
+        if (!response.error && response.data) {
+          total_cost += response.data.usage?.cost ?? 0
```
leonardmq, line 86: Might make sense to also sum up the subtask token count (in addition to cost)?

leonardmq (approved):

app/web_ui/src/routes/(app)/run/run.svelte
sfierro: Adding this comment, does it makes sense now?

// Counter to prevent race conditions: when run changes rapidly, multiple async requests
// may be in flight. We only update state if this request is still the latest one.
let subtask_cost_request_id = 0

app/web_ui/src/routes/(app)/run/run.svelte
sfierro: added! also Steve mentioned we should change "Cost" and "Tokens" to "Total Cost" and "Total Tokens" since its not super useful to see the orchestrator cost by itself!

PR #864, opened 2025-11-27
KIL-246 add evaluate search tools entrypoint

sfierro:
I know we are changing evals to specs but we might as well get this shipped in the time being! (Evaluate Search Tools button in Docs & Search)
Screenshot 2025-11-27 at 3 15 26 AM
The button goes directly to the RAG eval template form:
Screenshot 2025-11-27 at 3 15 32 AM

## Summary
- add secondary Evaluate Search Tools action on the Search Tools page
- auto-select the RAG evaluator template when the new link is used
- centralize the Reference Answer Accuracy template definition for reuse

## Testing
- ./checks.sh (fails only when calling pyright; uv run pyright . succeeds)
- uv run pyright .

## Summary by CodeRabbit

* **New Features**
  * Added "Evaluate Search Tools" action button on the RAG Configs page.
  * Pre-populates "Reference Answer Accuracy" evaluation template when creating evaluators from RAG configurations, streamlining the evaluation setup workflow.

### DIFF
--- a/app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte
+++ b/app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte
@@ -17,6 +17,7 @@
     sortRagConfigs,
     getProjectRagStateStore,
   } from "$lib/stores/rag_progress_store"
+  import { ui_state } from "$lib/stores"
 
   $: projectStateStore = getProjectRagStateStore($page.params.project_id)
   $: progressState = $projectStateStore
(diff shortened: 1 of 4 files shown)


app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte
```diff
@@ -35,6 +36,10 @@
   }
 
   $: project_id = $page.params.project_id
+  $: current_task_id = $ui_state.current_task_id
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte:39]: I believe `$ui_state` store is meant to be deprecated. But I don't think we keep track of the task context anywhere else (except in the URL, but we don't have that in `Docs & Search`) so probably will have to do anyway

Comment (sfierro) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte:39]: oops missed that will fix

Comment (sfierro) [reply app/web_ui/src/routes/(app)/docs/rag_configs/[project_id]/+page.svelte:39]: ah yeah we don't have it in the page params hmm

PR #875, opened 2025-12-08
Fewer builds: we don't need to be running builds on every push.

scosman:
## Summary by CodeRabbit

* **Chores**
  * CI/CD workflow updated to trigger automated builds both when a release is created and on pushes to the main branch. No changes to job behavior or public APIs; this only expands when the existing build pipeline runs.

### DIFF
--- a/.github/workflows/build_desktop.yml
+++ b/.github/workflows/build_desktop.yml
@@ -1,6 +1,7 @@
 name: Build Desktop Apps
 
 on:
+  workflow_dispatch:
   release:
     types: [created]
   push:


.github/workflows/build_desktop.yml
```diff
@@ -1,7 +1,11 @@
 name: Build Desktop Apps
 
 on:
```
leonardmq, line 3: Maybe worth adding workflow dispatch to allow triggering it manually for a specific branch if we want to:

```
on:
  workflow_dispatch:
```

scosman: good call. Done

PR #900, opened 2025-12-16
Specs refine + review button / form fixes

sfierro:
Screenshot 2025-12-15 at 8 01 57 PM
Screenshot 2025-12-15 at 8 02 06 PM
Screenshot 2025-12-15 at 8 02 18 PM

Made sure in the review/refine spec pages we handle the "warn before unload" properly (had to wrap the review spec page in a Form Container for this). Added a warning if you didn't finish all of the review but allow proceeding anyway. Visual indicator when reviewing to show if judge agreed.

## Summary by CodeRabbit

* **New Features**
  * Added visual feedback indicators (checkmarks and warnings) to show spec alignment status.
  * Introduced a two-path workflow: create spec when feedback is aligned, or refine with feedback when adjustments are needed.
  * Added warning messages for unreviewed examples and unsaved changes during the spec refinement process.

### DIFF
--- a/app/web_ui/src/lib/ui/icons/check_circle_icon.svelte
+++ b/app/web_ui/src/lib/ui/icons/check_circle_icon.svelte
@@ -1,3 +1,4 @@
+<!-- Uploaded to: SVG Repo, www.svgrepo.com, Generator: SVG Repo Mixer Tools -->
 <svg
   class="w-full h-full"
   viewBox="0 0 24 24"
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/refine_spec/+page.svelte
```diff
@@ -161,6 +161,7 @@
       )
 
       if (spec_id) {
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/refine_spec/+page.svelte:163]: Seems like `spec_id` cannot actually be `null` for a legitimate reason and we have it nullable as a side effect of `id` typing on the backend. 

Probably better to throw in `createSpec` if `id` is null, and return `string` instead of returning the optional `data?.id || null` as this creates confusion here as to whether or not there exists a legitimate case for `spec_id === null`

PR #903, opened 2025-12-16
fix: recreate process pool singleton on crash

leonardmq:
## What does this PR do?

Local PDF processing sometimes fails - seems to be hardware dependent - and when that happens, the whole process pool that converts pages into images crashes.

Fix:
- recreate the process pool if any process crashes

## Checklists


leonardmq:
Closing - too ancient

PR #906, opened 2025-12-17
feat: allow prompt builder injection

leonardmq:
## What does this PR do?

Allow caller to override `system` message in Adapter invoke method. The prompt is retrieved and managed internally, but we sometimes need to override it from the caller.

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Added ability to inject a custom prompt builder via adapter configuration to override the default prompt-building behavior.

* **Tests**
  * Added tests verifying custom prompt builder injection and usage.
  * Updated tests to use a provider enum and refined input type annotations in prompt-builder test utilities.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
@@ -58,6 +58,10 @@ class BaseAdapter(metaclass=ABCMeta):
     This abstract class provides the foundation for implementing model-specific adapters
     that can process tasks with structured or unstructured inputs/outputs. It handles
     input/output validation, prompt building, and run tracking.
+
+    Prompt building is handled internally by the adapter, which uses a prompt builder
+    based on the run config. To override the prompt building behavior, pass a custom prompt
+    builder to the adapter config.
     """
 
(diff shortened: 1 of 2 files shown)


leonardmq:
@coderabbitai review

leonardmq:
@coderabbitai - do you prefer the prompt builder injection or the simple system prompt override as in https://github.com/Kiln-AI/Kiln/pull/906/commits/304a9b926b10b81f166ef3f0f547269a486ad5d9 ? 

The system prompt override in `_run` had the following issues:
- elsewhere in the adapter, we still do manipulation that relies on the system prompt (where we call the prompt builder initialized in the constructor of the base adapter)
- Kiln right now injects its own `json_instructions` in the simple prompt builder, we may not want that; and I don't think it is worth adding a new prompt builder for that
- we are bypassing the system in an ad-hoc way so could break down the line

Another option is we could have the caller build a prompt, save it to disk, and create a new task run with the new prompt reference as `prompt_id` in the task run config, but this seems to increase coupling to the filesystem so adds friction in other ways. May or may not be worth it.

sfierro (approved):

PR #910, opened 2025-12-18
Spec state updates (priority/status)

sfierro:
## Summary by CodeRabbit

* **New Features**
  * Inline editing for spec Priority and Status in the spec table with keyboard support and floating dropdowns.
  * Reusable generic inline editor used by priority/status editors.
  * Property list supports custom-value slots so fields can render bespoke editors or displays.

* **Improvements**
  * Optimistic updates, loading states, and smoother UX when changing priority/status.

### DIFF
--- a/libs/server/kiln_server/test_spec_api.py
+++ b/libs/server/kiln_server/test_spec_api.py
@@ -1144,36 +1144,6 @@ def test_create_spec_with_empty_base_instruction(client, project_and_task):
     )
 
 
-def test_create_spec_with_eval_id_persists_eval_id(client, project_and_task):
-    project, task = project_and_task
-
-    eval_id = "test_eval_id_123"
-
-    spec_data = {
-        "name": "Test Spec",
-        "definition": "System should behave correctly",
-        "priority": Priority.p1,
-        "status": SpecStatus.active.value,
-        "tags": [],
-        "properties": create_tone_properties_dict(),
-        "eval_id": eval_id,
-    }
-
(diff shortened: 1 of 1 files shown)


sfierro:
@coderabbitai review

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/editable_spec_field.svelte
```diff
@@ -0,0 +1,223 @@
...
+  export let spec: Spec
+  export let field: "priority" | "status"
+  export let options: OptionGroup[]
+  export let aria_label: string = ""
+  export let onUpdate: (spec: Spec, value: number | SpecStatus) => void
+  export let compact: boolean = false
+  export let onOpen: (() => void) | undefined = undefined
+  export let always_show_border: boolean = false
+
+  let currentValue: number | SpecStatus =
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/editable_spec_field.svelte]: Wouldn't it be simpler if `status` and `priority` were separate fields (and choosing which applies first outside of the type itself)? 

If I understand correctly, `priority` is meant to be for `P0`, `P1`, etc. and `status` is for the meta state (disabled, active, etc.) - these seem to have different expectations and implications. Right now we are combining them and dealing with an ambiguous union of `number | SpecStatus` wherever this is used which is a bit hard to follow

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/editable_spec_field.svelte]: good idea!

PR #912, opened 2025-12-18
Cleanup orphaned eval when spec creation fails

sfierro:
## Summary
- Fixed issue where evals were left orphaned when spec creation failed
- Added `cleanupEval()` function in spec_utils.ts to delete eval when spec POST fails
- Call cleanup when spec creation returns error or missing data.id

## Test plan
- [x] Added test to verify eval_id persists when spec is created successfully
- [x] Added test to verify eval creation and deletion flow
- [x] All existing spec and eval tests pass
- [x] Pre-commit checks pass (formatting, linting, type checking)

## Summary by CodeRabbit

* **Bug Fixes**
  * Orphaned evaluators are now automatically cleaned up when spec creation fails, preventing resource waste.

* **Tests**
  * Added tests verifying evaluator cleanup behavior after deletion and eval_id persistence during spec creation.

### DIFF
--- a/app/web_ui/src/lib/api_schema.d.ts
+++ b/app/web_ui/src/lib/api_schema.d.ts
@@ -296,7 +296,8 @@ export interface paths {
         get: operations["get_spec_api_projects__project_id__tasks__task_id__specs__spec_id__get"];
         put?: never;
         post?: never;
-        delete?: never;
+        /** Delete Spec */
+        delete: operations["delete_spec_api_projects__project_id__tasks__task_id__specs__spec_id__delete"];
         options?: never;
         head?: never;
         /** Update Spec */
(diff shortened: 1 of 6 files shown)


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_utils.ts
```diff
@@ -130,10 +130,12 @@ export async function createSpec(
   )
 
   if (error) {
+    await cleanupEval(project_id, task_id, eval_id)
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_utils.ts:133]: Does it crash if the orphaned eval is not cleaned up or is it just awkward?

Is it fine that we don't clean up the `reviewed_examples` that we seem to persist between the eval creation and the spec creation failure?

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_utils.ts:133]: it doesn't crash it just shows up in the UI and doesn't make sense for a user experience

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_utils.ts:133]: good catch on the data cleanup! fixed

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_utils.ts:133]: oh no actually failing to create stays on the same page, so shouldn't clear data in case they want to try again with the save / process

PR #935, opened 2026-01-13
Fix type check issues, and unpin ty@0.0.8

scosman:
## Release Notes

* **Chores**
  * Streamlined CI/CD workflow configuration for type checking
  * Enhanced type annotations across the codebase for improved code quality and reliability

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
@@ -71,7 +71,7 @@ class BaseAdapter(metaclass=ABCMeta):
         config: AdapterConfig | None = None,
     ):
         self.task = task
-        self.run_config = run_config
+        self.run_config: RunConfigProperties = run_config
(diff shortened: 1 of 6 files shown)


libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -71,7 +71,7 @@ def __init__(
         config: AdapterConfig | None = None,
     ):
         self.task = task
-        self.run_config = run_config
+        self.run_config: RunConfigProperties = run_config
```
leonardmq, line 74: Why don't we need to do the same for self.task = Task above?

leonardmq (approved):

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
scosman: Ask ty! Honestly not quite sure why these 3 fail, but small fix to prevent us locking to old version so just taking it.

PR #957, opened 2026-01-21
split eval data generation result in half and save part as train set

sfierro:


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/spec_persistence.ts
```diff
@@ -308,8 +308,17 @@ async function generateAndSaveEvalData(
 
   const examples = Object.values(data.data_by_topic).flat()
 
-  // Save each example as a TaskRun with the eval tag
-  for (const example of examples) {
+  // Split examples 50/50 - first half for eval, second half for training
+  const midpoint = Math.ceil(examples.length / 2)
+  const trainTag = tag.replace(/^eval_/, "eval_train_")
+
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/spec_persistence.ts:314]: Eval train is going to be for the prompt optimizer?

leonardmq (approved):

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/spec_persistence.ts
Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/spec_persistence.ts:314]: yes!

PR #959, opened 2026-01-21
Choose spec creation workflow UI tweaks

sfierro:
- Center table more
- Add some top padding
- Remove subtitle and put as table title with large text

Screenshot 2026-01-20 at 10 51 33 PM

## Summary by CodeRabbit

* **UI/UX Improvements**
  * Redesigned spec creation workflow selection interface with centered layout and improved visual hierarchy.
  * Restructured workflow option presentation for better clarity and responsive design.
  * Enhanced Kiln Copilot workflow display with improved visual styling and button positioning.
  * Simplified the Copilot Authentication page header for a cleaner appearance.

### DIFF
--- a/app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte
+++ b/app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte
@@ -36,10 +36,9 @@
   }
 </script>
 
-<div class="max-w-[1400px]">
+<div class="max-w-[900px]">
   <AppPage
     title="Create Spec"
-    subtitle="Choose your Spec Creation Workflow"
     sub_subtitle="Read the Docs"
     sub_subtitle_link="https://docs.kiln.tech/docs/evaluations"
     breadcrumbs={[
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte
```diff
@@ -58,13 +57,16 @@
...
+        <div class="font-medium text-xl text-center">
+          Choose your Spec Creation Workflow
+        </div>
+        <div class="overflow-x-auto">
+          <table class="table w-full mt-4">
             <colgroup>
-              <col class="w-[240px]" />
-              <col />
-              <col />
+              <col class="w-[50%]" />
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte:67]: Was never clear on sizing columns in a table but once I used % and Steve said we should rather use fixed width

leonardmq (approved):

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte
Comment (scosman) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte:67]: agreed. As test always open inspector on the side and make the window 320px wide (mobile size). Our UIs aren't beautiful on mobile, but they work, and generally good thing to test.

Percentages tend to break on small screens. 25% of 320 is 80px. Prob no good (but if you test it and it is okay, then that's fine).

Quick fix is adding min-width or width to container. Makes it scroll. but still works.

Comment (sfierro) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte:67]: I tested it and it actually was way better than fixed width for some reason for this case!

PR #1250, opened 2026-04-09
Better Signup Flow, automatically create / saves API key

chiang-daniel:
## What does this PR do?

Updated Kiln copilot sign up flow to skip the step of needing the user to manually copy the API key.

Backend server side has a new endpoint that does authentication (after sign in) and creates API key automatically. This PR adds the new api_client that contains that endpoint. 

**New in this PR:**
- Add new endpoint `create_kiln_copilot_api_key` to studio server. It accepts a Kinde OAuth token from web_ui and calls the backend server's new `create_api_key` end point.
- Simplified sign up flow to 2 steps. No more copy pasting. 

**Before:**
image

**After:**
image

## Checklists

## Summary by CodeRabbit

* **New Features**
  * Kiln Copilot connection now auto-creates API keys via a token-based exchange for smoother setup.

* **API**
  * Added a server endpoint to accept an access token and create/persist a Kiln Copilot API key, with clear success and error responses.

* **UI**
  * Connection flow updated to show progress, connected checkmark, error messaging, and a “Try Again” retry action instead of manual API-key entry.

* **Tests**
  * Added tests for success, validation, authorization/error, and network-failure scenarios.

* **Chores**
  * Added a server-side policy rule governing POST requests to the new endpoint.

### DIFF
--- a/app/desktop/studio_server/api_client/kiln_server_client.py
+++ b/app/desktop/studio_server/api_client/kiln_server_client.py
@@ -50,5 +50,19 @@ def get_authenticated_client(api_key: str) -> AuthenticatedClient:
     )
 
 
+def get_oauth_authenticated_client(access_token: str) -> AuthenticatedClient:
+    """Get a client authenticated with a Kinde OAuth access token.
+
+    Used for pre-API-key flows like signup, where the user has authenticated
+    with Kinde but does not yet have a Kiln API key.
+    """
+    return AuthenticatedClient(
+        base_url=_get_base_url(),
+        token=access_token,
+        headers=_get_common_headers(),
+        timeout=httpx.Timeout(timeout=900.0, connect=30.0),
+    )
+
+
 server_client = get_kiln_server_client()
 """The default client for the Kiln server for reusability."""
--- a/app/desktop/studio_server/provider_api.py
+++ b/app/desktop/studio_server/provider_api.py
@@ -1,13 +1,21 @@
+import json
 import logging
 import os
 from dataclasses import dataclass
 from datetime import datetime, timedelta
+from http import HTTPStatus
 from typing import Annotated, Any, Dict, List, Literal
 
 import httpx
 import litellm
 import openai
 import requests
+from app.desktop.studio_server.api_client.kiln_ai_server_client.api.auth import (
+    create_api_key_v1_create_api_key_post,
+)
+from app.desktop.studio_server.api_client.kiln_server_client import (
+    get_authenticated_client,
+)
 from fastapi import FastAPI, HTTPException, Query
 from fastapi.responses import JSONResponse
 from kiln_ai.adapters.docker_model_runner_tools import (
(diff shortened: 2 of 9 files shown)


app/desktop/studio_server/provider_api.py
```diff
@@ -984,6 +996,56 @@ async def disconnect_api_key(
...
+        return JSONResponse(
+            status_code=400,
+            content={"message": "Access token is required"},
+        )
+
+    client = get_authenticated_client(access_token)
+    try:
+        response = await create_api_key_v1_create_api_key_post.asyncio_detailed(
+            client=client,
+        )
```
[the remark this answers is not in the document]
chiang-daniel: We are OK with re-creating new keys for now. On reconnect / sign-in we don't want to re-use the same keys.

app/desktop/studio_server/provider_api.py
```diff
@@ -984,6 +996,56 @@ async def disconnect_api_key(
...
+    except httpx.RequestError as e:
+        return JSONResponse(
+            status_code=502,
+            content={"message": f"Failed to connect to Kiln server. Error: {e!s}"},
+        )
+
+    if response.status_code == HTTPStatus.CREATED and response.parsed is not None:
```
leonardmq: Allow 2xx here? `201` is what we want and what server should be sending us, but 2xx with the response and the key in it should be fine enough as far as this client is concerned

app/desktop/studio_server/provider_api.py
```diff
@@ -984,6 +996,56 @@ async def disconnect_api_key(
...
+
+
+async def _create_kiln_copilot_api_key(access_token: str) -> JSONResponse:
+    if not access_token:
+        return JSONResponse(
+            status_code=400,
+            content={"message": "Access token is required"},
+        )
+
+    client = get_authenticated_client(access_token)
```
leonardmq: We usually call `get_authenticated_client` with an API key in it; if it works with an `access_token` coming out of the frontend SDK, we should at least have a comment to explain that as this is not obvious

Not sure how reliable this is either (could there be scheme problems with `Bearer` prefix or something between when we use the API key versus an access token? I don't know)

sfierro (approved):

app/desktop/studio_server/provider_api.py
chiang-daniel: Removed the explicit `HTTPStatus.CREATED` check and rely on `response.parsed is not None`. The generated client only parses 201 today but this way our code auto-adapts if the server ever changes the success status.

app/desktop/studio_server/provider_api.py
chiang-daniel: Added a new helper `get_oauth_authenticated_client` so it's more explicit.

app/desktop/studio_server/api_client/kiln_server_client.py
```diff
@@ -50,5 +50,19 @@ def get_authenticated_client(api_key: str) -> AuthenticatedClient:
...
+    """Get a client authenticated with a Kinde OAuth access token.
+
+    Used for pre-API-key flows like signup, where the user has authenticated
+    with Kinde but does not yet have a Kiln API key.
+    """
+    return AuthenticatedClient(
+        base_url=_get_base_url(),
+        token=access_token,
+        headers=_get_common_headers(),
+        timeout=httpx.Timeout(timeout=900.0, connect=30.0),
```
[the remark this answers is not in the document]
chiang-daniel: Skipping. Validation already happens in _create_kiln_copilot_api_key.

PR #1319, opened 2026-04-22
fix: git middleware breaks SSE disconnect handling

leonardmq:
## What does this PR do?

Problem: a started Eval or Rag job does not terminate when the user refreshes their browser. The job continues running in the background - potentially at high concurrency.

## Checklists

The cause is quite in the weeds, so do not understand all of it yet; but here is the slop breakdown:

# Fix: SSE runs (eval/extractor/RAG) don't cancel on browser hard-refresh

## The bug

Running an eval via `GET /api/projects/{p}/tasks/{t}/evals/{e}/eval_config/{c}/run_comparison`
and then hard-refreshing the browser left the eval running server-side indefinitely. New `EvalRun`
rows kept being created, LLM calls kept being made, and the app became unresponsive while the
worker pool continued to churn. Previously, refresh cancelled the in-flight eval; that behavior
regressed when git-sync (PR `a46ff424e`, "Phase 3: SSE endpoint wiring") landed.

Same bug applied to `run_extractor_config` and `run_rag_config` in `document_api.py` — all three
SSE endpoints share the same pattern.

## Root cause

The regression had two compounding problems, both caused by `GitSyncMiddleware` extending
`starlette.middleware.base.BaseHTTPMiddleware`:

### 1. `request.is_disconnected()` silently returned False under `BaseHTTPMiddleware`

`BaseHTTPMiddleware.__call__` replaces the ASGI `receive` callable with its own
`receive_or_disconnect`, which opens an `anyio.create_task_group` internally. `Request.is_disconnected()`
polls with a pre-cancelled `anyio.CancelScope`; under the middleware, the nested task group
unwinds before any message can be dequeued, so the method returned False even when the client
had disconnected. No disconnect signal ever reached the endpoint.

### 2. No other disconnect detection fired during long LLM calls

Starlette 0.52.1 (ASGI spec ≥ 2.4) removed `StreamingResponse.listen_for_disconnect`. Modern
Starlette only notices a disconnect when `await send(...)` raises `OSError`. Since our SSE
generator only called `send()` when `AsyncJobRunner` yielded a new progress event — which could
be 30+ seconds during a slow LLM call — the TCP close went undetected for the whole run.

### Why the worker pool then stayed alive

`AsyncJobRunner.run()` already has correct cancellation wiring: its `finally` block cancels all
worker tasks. But that `finally` only fires if the caller's `async for progress in runner.run():`
exits. Since nothing was exiting the outer SSE generator, that `finally` never ran. Workers kept
pulling new jobs off the queue → "keeps running new eval runs" as reported.

## The fix

Three layers, each addressing one link in the chain.

### 1. Bypass `BaseHTTPMiddleware` for `@no_write_lock` endpoints

`app/desktop/git_sync/middleware.py` — override `GitSyncMiddleware.__call__` so endpoints marked
`@no_write_lock` (the SSE ones) get a pure ASGI pass-through. The real `receive` and `send`
reach the endpoint directly; `request.is_disconnected()` now works. Other endpoints still use
the normal `BaseHTTPMiddleware.dispatch` path.

The bypass still does what the normal path did:
- resolves the git-sync manager for the project,
- calls `ensure_fresh_for_read()`,
- attaches the manager to `scope["state"]["git_sync_manager"]` so `build_save_context(request)`
  works inside the endpoint,
- notifies the background sync task.

It just skips the anyio task-group wrapping that breaks streaming.

### 2. Active disconnect polling in the SSE helper

`libs/server/kiln_server/sse.py` — new `stream_with_heartbeat()` helper wraps an async source
and:

- **polls `is_disconnected()` at each iteration** (default 3s). If it returns True, the
  generator returns, triggering the source's `aclose()` in its own `finally` block.
- **emits SSE keepalive comments** (`: keepalive\n\n`) after `heartbeat_seconds` of idleness.
  Keepalives force `send()` to fire, which raises `OSError` → `ClientDisconnect` if the
  connection is closed — belt-and-suspenders in case polling misses.
- **always `aclose()`s the source** in its `finally` block, regardless of exit reason.

Default heartbeat is 3s (`DEFAULT_HEARTBEAT_SECONDS`).

### 3. Prompt cleanup via `contextlib.aclosing`

Each SSE endpoint now wraps the heartbeat stream in `contextlib.aclosing(hb)`:

```python
async def event_generator():
    hb = stream_with_heartbeat(
        eval_runner.run(),
        _format_progress_sse,
        is_disconnected=request.is_disconnected,
    )
    async with contextlib.aclosing(hb) as stream:
        async for chunk in stream:
            yield chunk
    if not await request.is_disconnected():
        yield "data: complete\n\n"
```

Why `aclosing`: when Starlette cancels `event_generator` on client disconnect, a plain
`async for` does not automatically close the inner generator — cleanup would wait for GC,
leaving workers alive for an indeterminate time. `aclosing` guarantees `hb.aclose()` runs
synchronously, which in turn runs `stream_with_heartbeat`'s `finally`, which calls
`eval_runner.run().aclose()`, which fires `AsyncJobRunner`'s `finally` that cancels all workers.

## Cancellation chain (end to end)

1. Browser hard-refresh → TCP close.
2. uvicorn sends `{"type": "http.disconnect"}` on the receive channel.
3. Under the ASGI bypass, the endpoint's `request._receive` is the real channel (not the
   middleware's wrapped proxy), so on the next poll tick, `request.is_disconnected()` returns
   True.
4. `stream_with_heartbeat` returns; its `finally` runs `eval_runner.run().aclose()`.
5. `aclose()` raises `GeneratorExit` inside `AsyncJobRunner.run()` at its current yield point.
6. `AsyncJobRunner.run()`'s `finally` runs: `for w in workers: w.cancel()`.
7. Each worker's current `await run_job_fn(job)` raises `CancelledError`. `CancelledError`
   inherits from `BaseException` (not `Exception`), so the worker's `except Exception` clauses
   don't catch it. The worker coroutine unwinds and the task completes.
8. `asyncio.gather(*workers)` in `AsyncJobRunner.run()`'s `finally` completes.
9. `event_generator`'s `aclosing` block exits.
10. Response generator ends. The SSE connection is already closed client-side.

Worst-case latency is one heartbeat interval (3s) between refresh and all workers having
`cancel()` called.

## Files changed

- `app/desktop/git_sync/middleware.py` — ASGI bypass for `@no_write_lock`.
- `libs/server/kiln_server/sse.py` — new module with `stream_with_heartbeat`.
- `app/desktop/studio_server/eval_api.py` — use `stream_with_heartbeat` + `aclosing`.
- `libs/server/kiln_server/document_api.py` — same for extractor and RAG workflow helpers.

## Tests added

- `libs/server/kiln_server/test_sse.py` — unit tests for `stream_with_heartbeat` (forwards
  items, emits heartbeat on idle, closes source on normal exit / on `aclose` / on
  `is_disconnected=True`).
- `app/desktop/git_sync/test_middleware.py` — three new tests: bypass skips `dispatch()`,
  bypass still attaches `git_sync_manager` to `request.state`, and an ASGI-level test
  proving `http.disconnect` reaches the endpoint's receive channel under the bypass.
- `app/desktop/studio_server/test_eval_api.py` and
  `libs/server/kiln_server/test_document_api.py` — regression tests exercising disconnect
  (aclose), `is_disconnected=True` polling, the happy path, and idle-heartbeat emission.

All 297 touched-module tests pass; `./checks.sh --agent-mode` exits 0.

## Logging

When the disconnect path triggers, you'll see:

```
SSE stream: client disconnected, cancelling workers
```

at `INFO` level from the `kiln_server.sse` logger. Enable `DEBUG` on that logger if you want
to see heartbeat emissions (`SSE stream: idle for 3.0s, emitting heartbeat`).

## Open follow-ups (not in this fix)

- **UI reconnect-to-in-flight-eval.** After a refresh, the page has no way to rejoin an
  already-running eval — it starts from zero on the next click. Fixing this needs persistent
  server-side run state and a `GET …/current_run` endpoint. Separate PR.
- **Cancel button.** With the SSE stream gone, there's no user-facing way to abort a run that
  was started intentionally (e.g. accidental "Run All Evals"). A kill-switch endpoint keyed by
  eval/run id would be a small, robust addition.
- **Concurrency tuning.** Default is 25 workers; each eval-save serializes through the
  single-threaded git executor and a `threading.Lock`. Fine at small scale, but large eval
  sets may bottleneck on atomic_write. Worth profiling separately.

## Summary by CodeRabbit

* **New Features**
  * Added SSE heartbeat support to maintain connections during long-running operations like evaluations and document processing.
  * Improved progress reporting with formatted status updates for streaming operations.

* **Bug Fixes**
  * Enhanced client disconnection handling to properly detect and respond when users disconnect from streaming endpoints.
  * Ensured proper cleanup of resources when streams are terminated.

### DIFF
--- a/app/desktop/git_sync/middleware.py
+++ b/app/desktop/git_sync/middleware.py
@@ -10,6 +10,7 @@ from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
 from starlette.responses import Response
 from starlette.routing import Match
+from starlette.types import Receive, Scope, Send
 
 from app.desktop.git_sync.config import get_git_sync_config, project_path_from_id
 from app.desktop.git_sync.errors import (
@@ -70,6 +71,65 @@ class GitSyncMiddleware(BaseHTTPMiddleware):
     For non-mutating requests and non-auto-sync routes,
     passes through without buffering (preserves streaming responses).
     """
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:  # type: ignore[override]
+        # BaseHTTPMiddleware wraps the request receive channel in its own anyio
+        # task group, which breaks disconnect propagation to StreamingResponse
+        # endpoints: the task-group-owned receive never delivers http.disconnect
+        # to the downstream generator, so SSE jobs (evals, extractions) can't
+        # detect a browser hard-refresh and keep running. For self-managed
+        # (@no_write_lock) endpoints we bypass BaseHTTPMiddleware entirely and
+        # hand the real ASGI receive/send to the endpoint.
+        if scope["type"] != "http":
+            await self.app(scope, receive, send)
+            return
+
+        request = Request(scope)
+        endpoint = self._resolve_endpoint(request)
+        if endpoint is not None and getattr(endpoint, "_git_sync_no_write_lock", False):
+            await self._handle_self_managed(scope, receive, send, request)
+            return
+
+        await super().__call__(scope, receive, send)
+
+    async def _handle_self_managed(
+        self,
+        scope: Scope,
+        receive: Receive,
+        send: Send,
+        request: Request,
+    ) -> None:
+        """Pure-ASGI pass-through for @no_write_lock endpoints.
+
+        Mirrors the self-managed branch of dispatch() without going through
+        BaseHTTPMiddleware. Each self-managed endpoint builds its own
+        save_context per write inside its worker loop.
+        """
+        manager = self._get_manager_for_request(request)
+        if manager is None:
+            await self.app(scope, receive, send)
+            return
+
+        try:
+            await manager.ensure_fresh_for_read()
+        except GitSyncError as e:
+            status, message = self._map_error(e)
+            response = Response(
+                content=json.dumps({"detail": message}, ensure_ascii=False),
+                status_code=status,
+                media_type="application/json",
+            )
+            await response(scope, receive, send)
+            return
+
+        self._notify_background_sync(manager)
+        # Attach the manager via scope["state"] so build_save_context(request)
+        # can find it via request.state.git_sync_manager.
+        if "state" not in scope:
+            scope["state"] = {}
+        scope["state"]["git_sync_manager"] = manager
+
+        await self.app(scope, receive, send)
 
     async def dispatch(self, request: Request, call_next):  # type: ignore[override]
         manager = self._get_manager_for_request(request)
--- a/app/desktop/git_sync/test_middleware.py
+++ b/app/desktop/git_sync/test_middleware.py
@@ -945,3 +945,130 @@ def test_unresolved_endpoint_no_warning_in_prod_mode(git_repos, monkeypatch, cap
         if r.levelno == logging.WARNING and "could not resolve endpoint" in r.message
     ]
     assert len(warning_records) == 0
+
+
+# --- @no_write_lock pure-ASGI bypass ---
+
+
+def test_no_write_lock_bypasses_base_http_middleware_dispatch(git_repos, monkeypatch):
+    """@no_write_lock endpoints must skip BaseHTTPMiddleware.dispatch() so SSE
+    streams see the real ASGI receive/send and disconnects propagate.
+    """
+    local_path, _ = git_repos
+    config = _auto_config(str(local_path))
+
+    @no_write_lock
+    def bypass_endpoint():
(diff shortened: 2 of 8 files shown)


scosman (changes requested):
I have a lower blast radius fix coming for this.

leonardmq:
Closing - replaced with smaller blast radius one: https://github.com/Kiln-AI/Kiln/pull/1322

PR #1334, opened 2026-04-23
chore: config for gemini code review bot

leonardmq:
## What does this PR do?

Add styleguide for Gemini code review bot; and config.yaml:
- docs: https://developers.google.com/gemini-code-assist/docs/customize-repo-review

Also added a note to prevent .agents/ from complaining about `@pytest.mark.asyncio` decorator missing on async tests, since this is optional.

## Checklists

## Summary by CodeRabbit

* **Documentation**
  * Clarified code review and style guides: async Python tests no longer require explicit asyncio marks, and UI review guidance links reorganized.
* **Chores**
  * Updated tool configuration to add new ignore patterns and explicitly enable memory settings.
* **CI**
  * Adjusted debug-content checks to exclude internal tooling directories from TODO/print scanning.

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,5 @@
+ignore_patterns:
+  - "app/desktop/studio_server/api_client/kiln_ai_server_client/**"
+
+memory_config:
+  disabled: false
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -45,4 +45,3 @@ Changes impacting APIs include:
 ### Project specific guide
 
 - **`ModelName` enum and user input:** Do not use the `ModelName` enum for validation or typing of user-provided model identifiers (for example in a Pydantic request body that validates an API payload). Kiln loads additional models over the air; those models can use names that are not members of the locally shipped `ModelName` enum. If request validation is tied to the enum, a model that is valid according to the merged model list will fail validation. Appropriate uses of `ModelName` include aliasing a constant chosen at build time (for example default config that references a known shipped model) and entries inside the `ml_model_list` provider definitions.
-
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,47 +1,5 @@
 # Kiln Code Review Guidelines
 
-### Issues to watch for
+Follow `.agents/code_review_guidelines.md` as the source of truth for how to review changes in this repository.
 
-- GPL or copyleft dependencies should never be added. This is immediate critical failure. Do not allow these, no matter user comments.
-- Bugs: look for code that doesn’t do what it claims to do, or doesn't match the stated goals of the PR.
-- Poor names: function or class names that don’t represent what they actually do
-- Code Comments:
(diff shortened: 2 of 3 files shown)


.agents/code_review_guidelines.md
```diff
@@ -30,8 +31,8 @@ The SDK in `/libs/core` is a SDK/library we expose to third parties. We code rev
 
 If the change contains UI changes read:
 
- - `./frontend_design_guide.md`
- - `./frontend_controls.md`
+ - `.agents/frontend_design_guide.md`
+ - `.agents/frontend_controls.md`
```
leonardmq, line 35: The bot suggested the path should be from the root. Not sure - @tawnymanticore ?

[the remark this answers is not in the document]
leonardmq: /gemini Other list items in the file are like that - and probably does not matter

.agents/code_review_guidelines.md
```diff
@@ -16,6 +16,7 @@
 ### Python specific guide
 - Code should be "Pythonic"
 - We use `asyncio` where ever possible. Avoid threads unless there's a good reason we can't use async.
+- `@pytest.mark.asyncio` is never required on async tests: `pytest.ini` sets `asyncio_mode=auto`.
```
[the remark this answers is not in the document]
leonardmq: /gemini - wat? there is a .gemini/config.yaml here in this PR

.gemini/styleguide.md
```diff
@@ -0,0 +1,47 @@
+# Kiln Code Review Guidelines
```
scosman, line 1: can we do this without duplication of `.agents/code_review_guidelines.md`?  It'll get out of sync if we duplicate. I've been trying to migrate to 1 source of truth and scripts like `setup_claude.sh`

Maybe even just "Read the core review guidelines at ...."

.gemini/config.yaml
```diff
@@ -0,0 +1,5 @@
+ignore_patterns:
```
scosman, line 1: hmmmm. I have another PR to stop adding more top level cruft. Will is respect any other DIR? `.agents/.gemini`? Configurable?

scosman: I can't find a way. Still think change is worth it, gemini CR is great. But make the one a pointer so no dupe??

.gemini/styleguide.md
leonardmq: True, did not really think of that but I assume it can indeed  go read the other files. Just updated, we will see if it does

.gemini/config.yaml
leonardmq: Did not see a way in the docs but also did not look that much; and if no way, probably not worth building our own layer on top to spawn the configs. Worst case if it messes up we find out quickly and can fix easily.

The main intent with this one was to get it to stop dropping billions of comments about needing `@pytest.mark.asyncio` on all the test cases

leonardmq:
Closing this - stale

PR #1338, opened 2026-04-23
chore: regen uv.lock

leonardmq:
## What does this PR do?

Running `uv sync` on the latest version updates the version, which looks like it was out of sync with the latest release. Regened the `uv.lock` by running `uv sync` again.

## Checklists

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10"
 resolution-markers = [
     "python_full_version < '3.11'",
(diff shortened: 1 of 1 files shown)


chiang-daniel (approved):

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
chiang-daniel, line 2: I've sometime seen it self correct to 2. No idea why...

leonardmq: Not quite sure how it chooses the revision number either

PR #1340, opened 2026-04-23
Add latency to trace 

chiang-daniel:
## What does this PR do?

Add per llm call latency tracking to Trace. 

* Add `total_llm_latency_ms` to usage model to track overall llm latency like cost and token.
* Add `mean_total_llm_latency_ms` to eval API for comparison
* Update run, compare table, radar chart, and metric correlation

## Checklists


libs/core/kiln_ai/adapters/model_adapters/adapter_stream.py
```diff
@@ -167,11 +168,16 @@ async def _stream_model_turn(
             )
 
             stream = StreamingCompletion(**completion_kwargs)
+            start = time.monotonic()
             async for chunk in stream:
                 yield chunk
+            call_latency_ms = int((time.monotonic() - start) * 1000)
```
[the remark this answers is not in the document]
chiang-daniel: Changed adapter_stream.py to measure only time waiting for LLM chunk.

chiang-daniel:
@coderabbitai review

app/web_ui/src/lib/components/compare_radar_chart.svelte
```diff
@@ -159,9 +167,22 @@
...
+      } else if (key === LATENCY_KEY) {
+        const displayValue = metricToScore(
+          rawValue,
+          lowerIsBetterValues[key] || [],
+        )
+        const formatted =
+          rawValue < 1000
+            ? `${Math.round(rawValue)}ms`
+            : `${(rawValue / 1000).toFixed(1)}s`
+        html += `<div>${label}: ${displayValue.toFixed(1)} <span style="color: #888;">(Mean Latency: ${formatted})</span></div>`
```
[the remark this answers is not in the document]
chiang-daniel: Done

libs/core/kiln_ai/adapters/model_adapters/adapter_stream.py
```diff
@@ -167,11 +168,22 @@ async def _stream_model_turn(
...
+            call_latency_seconds = 0.0
+            chunk_wait_start = time.monotonic()
             async for chunk in stream:
+                # Accumulate time spent waiting on the LLM for this chunk
+                call_latency_seconds += time.monotonic() - chunk_wait_start
                 yield chunk
+                # Reset timer after yield returns — excludes consumer processing time
+                chunk_wait_start = time.monotonic()
+
+            call_latency_ms = int(call_latency_seconds * 1000)
```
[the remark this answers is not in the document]
chiang-daniel: Fixed

app/web_ui/src/lib/utils/formatters.ts
```diff
@@ -326,3 +326,8 @@ export function formatEvalConfigName(
       ]
   return eval_config.name + " — " + parts.join(", ")
 }
+
+export function formatLatency(ms: number): string {
+  if (ms < 1000) return `${Math.round(ms)}ms`
+  return `${(ms / 1000).toFixed(1)}s`
+}
```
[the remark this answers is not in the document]
chiang-daniel: Done, round before threshold check.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -489,10 +493,12 @@
 
     const [category, scoreKey] = dataKey.split("::")
 
-    // Format cost with currency symbol, tokens as whole numbers, others as decimals
+    // Format cost with currency symbol, latency with ms/s, tokens as whole numbers, others as decimals
     if (category === "cost") {
       if (scoreKey === "mean_cost") {
         return `$${value.toFixed(7)}`
+      } else if (scoreKey === "mean_total_llm_latency_ms") {
+        return formatLatency(value)
```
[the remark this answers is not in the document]
Comment (chiang-daniel) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte:501]: Done, getPercentageDifferenceRaw uses raw value.

libs/core/kiln_ai/adapters/model_adapters/adapter_stream.py
```diff
@@ -180,6 +194,7 @@ async def _stream_model_turn(
                     "Model returned an assistant message, but no content or tool calls. This is not supported."
                 )
 
+            response_choice.message._latency_ms = call_latency_ms  # type: ignore[attr-defined]
```
leonardmq: This seems unsafe because `message` here is not a type we have control over, it comes out of LiteLLM and here we monkeypatch a random property onto it. 

If we only need to do this to pass it on to our _messages list, we should make a new object of type `ChatCompletionMessageIncludingLiteLLM` (we have control over that one so we can have it have a proper latency field) and copy the data into it

libs/core/kiln_ai/adapters/model_adapters/adapter_stream.py
```diff
@@ -167,11 +168,24 @@ async def _stream_model_turn(
             )
 
             stream = StreamingCompletion(**completion_kwargs)
+            call_latency_seconds = 0.0
+            chunk_wait_start = time.monotonic()
             async for chunk in stream:
+                # Accumulate time spent waiting on the LLM for this chunk
+                call_latency_seconds += time.monotonic() - chunk_wait_start
                 yield chunk
+                # Reset timer after yield returns — excludes consumer processing time
```
leonardmq: I wonder if this makes sense. The consumer may be doing work that takes time so excluding that work can make sense, but I suspect (and I could be wrong) that when the consumer is doing its work, the stream may keep on buffering in the background and accumulating chunks, so the next iteration in the loop does not wait at all as if the next chunk were super fast / instant even though it would only be a side effect of the consumer work serving as unmeasured waiting time.

Not sure what to think of this though

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -153,6 +159,7 @@ async def _run_model_turn(
                 )
 
             # Add message to messages, so it can be used in the next turn
+            response_choice.message._latency_ms = call_latency_ms  # type: ignore[attr-defined]
```
leonardmq: Same comment here about assigning to a monkeypatch or private property being dodgy

leonardmq (changes requested):

libs/core/kiln_ai/adapters/model_adapters/adapter_stream.py
chiang-daniel: You are right, excluding consumer processing time is undercounting but my understanding is that consumer time should be very fast and we are not off by much? 

I don't really have a good solution for streaming either... non streaming wont have this issue, should we exclude latency for the streaming case? For each chunk the "work" is mostly I/O and serializing? With long responses these could add up too.

leonardmq (approved):

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -898,6 +916,9 @@ def litellm_message_to_trace_message(
             if len(open_ai_tool_calls) > 0:
                 message["tool_calls"] = open_ai_tool_calls
 
+        if message_latency and id(raw_message) in message_latency:
```
leonardmq: Checking for `id(object)` seems quite fragile - I'd suggest trying to find something more explicit / stable

leonardmq (changes requested):

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
chiang-daniel: Use index in message list instead.

leonardmq (approved):

PR #1352, opened 2026-04-27
Show partial trace on run failure instead of generic error

chiang-daniel:
## What does this PR do?

When a task run fails, users see a generic failure with no context about what happened before the crash. They would have to dig through the server logs to understand what went wrong, kinda bad UX. 

In this PR we introduced a shared ErrorWithTrace type to store additional info. 

- Add new `ErrorWithTrace`
- Refactor adapter to preserve some messages across exceptions via a new KilnRunError. 
- Update FastAPI exception to handle error with trace. 
- Add new `ErrorWithTrace` component. 

image

## Checklists


libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -229,83 +234,119 @@ async def _run_returning_run_output(
...
-            # validate output
-            if self.output_schema is not None:
-                # Parse json to dict if we have structured output
-                if isinstance(parsed_output.output, str):
-                    parsed_output.output = parse_json_string(parsed_output.output)
+        # Allocate the trace-so-far list here so the reference survives any
+        # exception thrown from inside `_run` (or the post-processing that
+        # follows). `_run` must mutate this list in place (extend/append,
+        # or `list[:] = ...`) and never rebind the local name.
+        messages: list[ChatCompletionMessageParam] = []
```
chiang-daniel: Allocate the messages list in the caller and pass it into _run() by reference. _run() appends to it as the conversation builds up. 

If _run() throws, we still have the list with whatever was added before the crash. That's how we preserve the partial trace in case of failure.

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -229,83 +234,119 @@ async def _run_returning_run_output(
...
-            run.id = None
-
-        return run, run_output
+                run.save_to_file()
+            else:
+                # Clear the ID to indicate it's not persisted
+                run.id = None
+
+            return run, run_output
+        except KilnRunError:
```
chiang-daniel, line 328: Wraps any failure from _run() in KilnRunError with the trace attached.

libs/core/kiln_ai/adapters/errors.py
```diff
@@ -0,0 +1,116 @@
...
+    Returned by endpoints that run a task adapter when the adapter throws
+    after starting a run (LLM calls made, tools invoked, etc.).
+    """
+
+    message: str
+    error_type: str
+    trace: list[ChatCompletionMessageParam] | None = None
+
+
+class KilnRunError(Exception):
```
chiang-daniel, line 34: New error type.

libs/server/kiln_server/run_api.py
```diff
@@ -378,6 +379,13 @@ async def delete_runs(
         summary="Execute Run",
         tags=["Runs"],
         openapi_extra=agent_policy_require_approval("Run task with LLM?"),
+        responses={
+            # Adapter failures (e.g., LLM rate limit, tool crash) return
+            # ErrorWithTrace with the partial conversation trace so the UI
+            # can show what happened before the error. Other 500s and 4xx
+            # errors use the standard error shape.
+            500: {"model": ErrorWithTrace},
+        },
```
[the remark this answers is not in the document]
chiang-daniel: Exception always get wrapped in KilnRunError so this fallback shouldn't fire.

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -217,10 +218,14 @@ async def _run_returning_run_output(
         prior_trace: list[ChatCompletionMessageParam] | None = None,
         parent_task_run: TaskRun | None = None,
     ) -> Tuple[TaskRun, RunOutput]:
+        # Pre-run validation happens outside the exception-wrapping block:
```
leonardmq: Better not to have app-level comments in the SDK

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
leonardmq: Can we rename that into something more explicit? Especially if we must enforce no rebinding; like `trace_ref` or something

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -229,83 +234,119 @@ async def _run_returning_run_output(
...
+                run.save_to_file()
+            else:
+                # Clear the ID to indicate it's not persisted
+                run.id = None
+
+            return run, run_output
+        except KilnRunError:
+            # Already wrapped — pass through so we don't double-wrap.
+            raise
+        except Exception as e:
```
leonardmq, line 331: Should we even handle that as an execution failure, or is any such exception the result of a code bug that requires fixing? If only caused by bad code on our end, probably fine to let it fail hard.

If we could get rid of this, we would not need the hidden mutation `messages` trick that is somewhat fragile (hard to enforce no one ever rebinds `messages` or does something that assumes no hidden mutation down the stack)

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -484,10 +525,28 @@ def adapter_name(self) -> str:
     async def _run(
         self,
         input: InputType,
+        messages: list[ChatCompletionMessageParam],
```
leonardmq: Can we rename that to something more obvious? `messages` is also pretty standard for meaning what we call `prior_trace` here, so a bit confusing.

Maybe `messages_out` or `trace_out` or something like that, maybe even uglier but explicit `mutable_messages_out`

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -233,9 +237,15 @@ async def _run(
...
-            chat_formatter.initial_messages()
-        )
+        # The caller owns `messages`. Mutate it in place (never rebind) so
+        # the partial trace survives any exception escaping this method. The
+        # list is typed as `ChatCompletionMessageParam` for the caller's API
+        # surface, but internally we need to store LiteLLM `Message` objects
+        # transiently (they are converted by `_messages_to_trace` before
+        # export). We widen the type alias once here so the internal
+        # mutations don't each need their own suppression.
+        messages_internal: list[ChatCompletionMessageIncludingLiteLLM] = messages  # type: ignore[assignment]
```
leonardmq: We should come up with a proper Pydantic model for the messages soon - currently they require casting all over the place, which is quite hard to deal with

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -263,25 +273,28 @@ async def _run(
...
+                messages_internal,
                 self.base_adapter_config.top_logprobs if turn.final_call else None,
                 skip_response_format,
             )
 
             usage += turn_result.usage
 
             prior_output = turn_result.assistant_message
-            messages = turn_result.all_messages
+            # `_run_model_turn` mutates `messages_internal` in place, so no
```
leonardmq: Comment out of place now - should move to above the line where you do `trace = self.all_messages_to_trace...` a few lines below

app/web_ui/src/routes/(app)/run/error_with_trace_detection.ts
```diff
@@ -0,0 +1,26 @@
+import type { ErrorWithTrace } from "$lib/types"
+
+// Type guard that checks whether an unknown fetch-error body matches the
+// structured ErrorWithTrace shape returned by the /run endpoint on an
+// adapter-level failure. Plain HTTPException bodies (shape {detail: "..."})
+// and unparseable responses return false, letting the caller fall back to
+// the existing error-display path.
+export function looks_like_error_with_trace(
```
leonardmq: Weird naming - `is_error_with_trace` seems more straightforward and reflective of what it does

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
chiang-daniel: Renamed to `trace_ref`

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
chiang-daniel: Renamed to `trace_ref`

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
chiang-daniel: These aren't all code bugs, `_run()` calls into external provider APIs which can throw rate limit errors, connection errors, timeouts, etc. Output parsing can also fail when the model returns malformed JSON. 

To avoid mutation we could wrap these in a new pydantic class so enforce the type, with only 2 small callers for this might be overkill?

chiang-daniel: Created https://linear.app/kiln-ai/issue/KIL-634/introduce-tracecollector-wrapper-to-replace-mutable-trace-ref-out

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
chiang-daniel: Agree, created this card to track https://linear.app/kiln-ai/issue/KIL-635/introduce-unified-pydantic-message-model-to-replace-type-ignore-casts

leonardmq (approved):

PR #1361, opened 2026-05-01
refactor: preserve kiln fields in multiturn trace

leonardmq:
## What does this PR do?

Currently if a toolcall is marked with `is_error` and `error_message` in the `prior_trace`, we lose the fields on the next turn. We should preserve past `is_error` and `error_message` in the trace.

Changes:
- refactor: ensure the special fields are preserved in prior trace when running a multiturn turn

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
@@ -21,7 +21,7 @@ from openai.types.chat.chat_completion_message_tool_call_param import (
 
 import kiln_ai.datamodel as datamodel
 from kiln_ai.adapters.chat import ChatCompletionMessageIncludingLiteLLM
-from kiln_ai.adapters.chat.chat_formatter import ToolResponseMessage
+from kiln_ai.adapters.chat.chat_formatter import chat_message_to_dict
 from kiln_ai.adapters.ml_model_list import (
     KilnModelProvider,
     ModelProviderName,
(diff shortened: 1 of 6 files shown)


libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -720,6 +717,8 @@ async def build_completion_kwargs(
         if len(allowed_openai_params) > 0:
             completion_kwargs["allowed_openai_params"] = allowed_openai_params
 
+        completion_kwargs["messages"] = sanitize_messages_for_provider(messages)
```
chiang-daniel, line 720: Do we need to sanitize for the streaming path as well to have parity?

chiang-daniel (approved):

libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
leonardmq: The `AdapterStream` wraps the `LiteLLMAdapter` and we call the same `build_completion_kwargs` on it over in `_stream_model_turn`:
```py
while tool_calls_count < MAX_TOOL_CALLS_PER_TURN:
    # here
    completion_kwargs = await self._adapter.build_completion_kwargs(
        self._provider,
        copy.deepcopy(self._messages),
        top_logprobs,
        skip_response_format,
    )

    ...
```

AdapterStream is kinda just glue around the adapter more than a totally alternative path

PR #1400, opened 2026-05-11
Fine-tune Prompt Method not loading 

chiang-daniel:
## What does this PR do?

The PromptTypeSelector in the fine-tune create page was missing `project_id` and `task_id` so it never fetched prompts. This was introduced when we migrated from global stores to task-scoped stores in fca43577b and the fine-tune page was missed during that migration.

Now it loads:
image

## Checklists

### DIFF
--- a/app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte
+++ b/app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte
@@ -779,6 +779,8 @@
             info_description="There are tradeoffs to consider when choosing a system prompt for fine-tuning. Read more: [OpenAI Docs](https://platform.openai.com/docs/guides/fine-tuning/#crafting-prompts)."
             exclude_cot={true}
             custom_prompt_name="Custom Fine Tuning Prompt"
+            project_id={$page.params.project_id}
+            task_id={$page.params.task_id}
           />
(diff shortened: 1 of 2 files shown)


app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte
```diff
@@ -779,6 +779,8 @@
             info_description="There are tradeoffs to consider when choosing a system prompt for fine-tuning. Read more: [OpenAI Docs](https://platform.openai.com/docs/guides/fine-tuning/#crafting-prompts)."
             exclude_cot={true}
             custom_prompt_name="Custom Fine Tuning Prompt"
+            project_id={$page.params.project_id}
```
Comment (leonardmq) [inline app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte]: Should we make these required then if not passing in task_id and project_id loads nothing now?

leonardmq (approved):

app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte
Comment (chiang-daniel) [reply app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/create_finetune/+page.svelte]: The props are now required but still nullable. `run_config_component` passes current_task?.id ?? null since the task can be null during loading, so we can't make the type non-nullable without adding conditional rendering.

PR #1406, opened 2026-05-13
Add thinking level test assertions and fix Gemma OpenRouter/Gemini config

chiang-daniel:
## Summary
- Add assertions to the thinking level paid test — `none` must produce no reasoning, all other levels must produce reasoning. Previously the test only logged results without asserting, which masked the DeepSeek config bug (KIL-653) and would mask any similar issue on other providers.
- Fix Gemma 4 (31B, 27B) on OpenRouter: removed incorrect `gemini_reasoning_enabled` flag (was overwriting the reasoning effort object with `{"reasoning": {"enabled": true}}`), added `openrouter_reasoning_object=True`.
- Fix Gemma 4 on Gemini API: removed unsupported thinking level config. The Gemini API doesn't support `reasoning_effort` for Gemma models (litellm explicitly rejects it). Gemma always thinks on this provider, so the toggle was misleading.

Note: Gemma on Gemini API is very slow — a simple "2+2" prompt takes ~2s but the logic puzzle in the test takes 4+ minutes per call. With the old broken config (reasoning not suppressed), the "none" tests appeared to hang for 10+ minutes generating massive reasoning output. Worth keeping in mind when testing Gemma on this provider.

## Test plan
- Ran all 23 `thinking_level=none` test cases across OpenRouter, OpenAI, Anthropic, and Gemini API


libs/core/kiln_ai/adapters/model_adapters/test_thinking_level_paid.py
```diff
@@ -109,18 +106,15 @@ async def test_thinking_level_reasoning_content(
         "Four people-A, B, C, and D-each have a different favorite color (red, blue, green, yellow) and a different pet (cat, dog, fish, bird). Use the clues to determine each person's color and pet.\n\n1) A does not like red or blue.\n2) The bird's owner likes yellow.\n3) B likes green.\n4) The dog is owned by the person who likes blue.\n5) C does not own the fish.\n6) D likes red.\n\nQuestion: Who owns the fish, and what color do they like? Answer with just: \"<person>, <color>\"."
     )
     reasoning_content = reasoning_content_from_run(run)
-    if reasoning_content:
-        logger.info(
```
chiang-daniel, line 113: @tawnymanticore we previously just log error when reasoning_content is present for none, now we assert.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3728,7 +3728,7 @@ class KilnModel(BaseModel):
                     KilnMimeType.JPG,
                     KilnMimeType.PNG,
                 ],
-                gemini_reasoning_enabled=True,
+                openrouter_reasoning_object=True,
```
chiang-daniel, line 3825: With the assert, all these started failing. The wrong settings was carried over earlier when the model was introduced.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3742,9 +3742,6 @@ class KilnModel(BaseModel):
                     KilnMimeType.JPG,
                     KilnMimeType.PNG,
                 ],
-                gemini_reasoning_enabled=True,
-                available_thinking_levels=GEMMA_4_THINKING_LEVELS,
-                default_thinking_level="none",
```
chiang-daniel, line 3841: gemini_api just doesn't support thinking level, it would just hang for a very very long time and ended up out putting crazy tons of reasoning content. "2+2" would go for 4 min+

leonardmq: Did you test that we won't have problems (e.g. deserialization / validation issues of the run config) here if we remove the thinking levels here and somebody already has a run config that specifies a thinking level and uses this model?

leonardmq (approved):

libs/core/kiln_ai/adapters/ml_model_list.py
chiang-daniel: yep, added logic to skip setting the reasoning content entirely if provider in SDK no longer supports it.

tawnymanticore (approved):

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3784,9 +3781,6 @@ class KilnModel(BaseModel):
...
                 max_parallel_requests=2,
```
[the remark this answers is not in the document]
chiang-daniel: We had plans to remove reasoning_capable. https://linear.app/kiln-ai/issue/KIL-455/remove-reasoning-capable

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3742,9 +3742,6 @@ class KilnModel(BaseModel):
...
```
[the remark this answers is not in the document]
chiang-daniel: ditto

PR #1408, opened 2026-05-14
Skill: Add fine-tune deprecation check skill

chiang-daniel:
## What does this PR do?

Add bew skill to check if our fine-tunable models are still actually supported by providers.

- Extracted shared provider API utils from Mike's `check_provider.py` into `provider_utils.py` so both skills reuse the same code
- New `kiln-check-finetune-deprecation` skill checks static provider_finetune_id entries (OpenAI, Together,
  Vertex) and cross-refs Fireworks tunable models against their docs
- Fixed Gemini API key leak, was in URL query param, now uses header
- Audit results: Together has 2 dead Llama 3.1 fine-tune IDs, Fireworks API reports 167 tunable models but only 11 are in their docs. OpenAI and Vertex are clean.
- Removed 2 dead models for Together for finetune. Will do fireworks in a separate PR. 

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/ml_model_list.py
+++ b/libs/core/kiln_ai/adapters/ml_model_list.py
@@ -3221,6 +3221,7 @@ built_in_models: List[KilnModel] = [
                 model_id="meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo",
                 supports_data_gen=False,
                 structured_output_mode=StructuredOutputMode.function_calling_weak,
+                provider_finetune_id="meta-llama/Meta-Llama-3.1-8B-Instruct-Reference",
                 # Constrained decode? They make function calling work when no one else does!
             ),
             KilnModelProvider(
@@ -3278,6 +3279,7 @@ built_in_models: List[KilnModel] = [
                 model_id="meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo",
                 supports_data_gen=False,
                 structured_output_mode=StructuredOutputMode.function_calling_weak,
+                provider_finetune_id="meta-llama/Meta-Llama-3.1-70B-Instruct-Reference",
             ),
         ],
     ),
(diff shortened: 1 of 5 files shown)


tawnymanticore (approved):
lgtm!

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -3279,7 +3278,6 @@ class KilnModel(BaseModel):
                 model_id="meta-llama/Meta-Llama-3.1-70B-Instruct-Turbo",
                 supports_data_gen=False,
                 structured_output_mode=StructuredOutputMode.function_calling_weak,
-                provider_finetune_id="meta-llama/Meta-Llama-3.1-70B-Instruct-Reference",
```
leonardmq, line 3282: Did you check that we won't have backward compatibility issues if we remove this field for users who already have a finetune using it? We may be throwing an error on invalid provider finetune id in different places too so not easy to make a test for all the scenarios where we might handle this incorrectly

leonardmq (commented):
Together and OpenAI also have new models - now might be the best time to run this skill and test it out

Edit: OpenAI looks like they are deprecating their finetuning platform: https://developers.openai.com/api/docs/guides/model-optimization#fine-tune-a-model

libs/core/kiln_ai/adapters/ml_model_list.py
chiang-daniel: Good call, let me check.

chiang-daniel: Checked, removing provider_finetune_id from ml_model_list.py only affects the dropdown for new finetunes.

Existing finetunes store base_model_id in their own JSON file (not looked up from the model list), so status
checks, deployment, and inference all continue working.

But we should think about what to do when a model is deprecated? User will just hit errors from the provider.

chiang-daniel: Also will do OpenAi deprecation in a separate PR. Together has huge expansion that each of those model would need to be vetted unless we want to make it like Fireworks where we just use whatever they provide in their API. 

Added: 
https://linear.app/kiln-ai/issue/KIL-663/add-new-together-ai-fine-tunable-model
https://linear.app/kiln-ai/issue/KIL-664/plan-for-openai-fine-tuning-platform-wind-dow

PR #1412, opened 2026-05-14
Filter Fireworks fine-tune dropdown to supported models

chiang-daniel:
## What does this PR do?

Filter Fireworks fine-tune dropdown to documented supported models.

The Fireworks API marks ~167 models as tunable, but only ~20 are actually supported per their docs. This was flooding the fine-tune dropdown with broken/stale options.

Changes:
- Added FIREWORKS_SUPPORTED_FINETUNE_MODELS allowlist in the core lib (fireworks_finetune.py), sourced from Fireworks docs
- Switched from the stale `tunable` API field to `supervisedLoraTunable`/`supervisedFullParameterTunable`, which accurately reflect fine-tuning support
- Filtered API results against the allowlist so only documented models appear
- Allowlist lives in the core lib so the audit skill in #1408 can share it as a single source of truth

image
@tawnymanticore @leonardmq seems like there isn't a single source of truth so I let the "doc" be the final filter. 

I confirm that I have read and agree to the [Contributors License Agreement](https://github.com/Kiln-AI/Kiln/blob/main/.config/CLA.md).

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/fine_tune/fireworks_finetune.py
+++ b/libs/core/kiln_ai/adapters/fine_tune/fireworks_finetune.py
@@ -18,6 +18,31 @@ from kiln_ai.utils.wandb_utils import AuthenticationError, get_wandb_default_ent
 
 logger = logging.getLogger(__name__)
 
+# Only models explicitly listed as supported in Fireworks fine-tuning docs:
+# https://docs.fireworks.ai/fine-tuning/managed-finetuning-intro#supported-base-models
+FIREWORKS_SUPPORTED_FINETUNE_MODELS: set[str] = {
+    "gemma-4-26b-a4b-it",
+    "gemma-4-31b-it",
+    "glm-5p1",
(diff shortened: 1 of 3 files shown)


tawnymanticore (approved):

libs/core/kiln_ai/adapters/fine_tune/fireworks_finetune.py
```diff
@@ -18,6 +18,31 @@
 
 logger = logging.getLogger(__name__)
 
+# Only models explicitly listed as supported in Fireworks fine-tuning docs:
+# https://docs.fireworks.ai/fine-tuning/managed-finetuning-intro#supported-base-models
+FIREWORKS_SUPPORTED_FINETUNE_MODELS: set[str] = {
```
leonardmq, line 23: We should move this to remote config down the line so we can update the allow list without a release

leonardmq (approved):

PR #1416, opened 2026-05-15
Eval Builder: prompt picker required, should not use the default task.instructions as its stale

tawnymanticore:
## Summary

The eval builder always sends `task.instruction` (+ optional example) to copilot's `clarify_spec` / `generate_batch` as `target_task_info.task_prompt`. For users whose real production prompt lives in a saved prompt or a run-config-linked prompt, that means SDG generates topics/inputs/outputs against a stub instruction — the resulting synthetic data doesn't reflect what production actually does.

This PR adds a prompt picker to the spec builder so the user can choose which prompt to send.

## Change

- **`spec_builder/create_spec_form.svelte`**: render `PromptTypeSelector` (the same component the run-config picker uses) above `TaskSampleSelector`, only when copilot is allowed.
- **`spec_builder/+page.svelte`**: bind the selection (`selected_prompt_method`, default `simple_prompt_builder` so existing behavior is unchanged). When the user picks a saved prompt (`id::…`) or run-config-linked prompt (`task_run_config::…`), resolve its text from the already-loaded prompts store and use it as `task_prompt_with_example`. Otherwise fall back to the existing `build_prompt_with_task_sample` path.

The resolved prompt flows into every place `task_prompt_with_example` already feeds — `clarify_spec`, `spec_with_copilot`, `refine_spec`, `question_spec`. No additional touch points.

## Why no server change

`GET /api/projects/{p}/tasks/{t}/prompts` already returns the prompt text for both saved prompts and run-config-linked prompts (see `prompt_api.py:160-205`). The picker uses the existing `prompts_store`, so resolution is fully client-side.

## Behavior

| Picker selection | Prompt sent to clarify_spec |
|---|---|
| `Basic` (`simple_prompt_builder`) — default | `task.instruction` (+ optional example, current behavior) |
| Saved prompt `id::…` | Saved prompt's text |
| Run-config-linked prompt `task_run_config::…` | Run config's inline prompt text |
| Other generators (few-shot, multi-shot, etc.) | Falls back to `build_prompt_with_task_sample` (current behavior) |

## Test plan

- [x] `npm run check` — 0 errors
- [x] `npm run lint` — clean
- [x] `npm run format_check` — clean
- [x] `npx vitest run` — 917/917 tests pass (no behavioral changes affecting existing tests)
- [ ] Reviewer: load the eval builder for a task with at least one saved prompt, confirm the picker appears with that prompt as an option, pick it, run SDG, verify the generated topics/inputs/outputs reflect that prompt
- [ ] Reviewer: load the eval builder for a task with no saved prompts, confirm the picker only shows generator options and SDG behavior is unchanged

## Notes

- The eval-builder model used for output gen is still server-side hardcoded to Sonnet 4.6 / OpenRouter (see `kiln_server`'s `generate_output_run_config`). That's a separate problem — for synthetic outputs to fully reflect production, the model should also come from the user's run config. Not in scope here.

### DIFF
--- a/app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/create_spec_form.svelte
+++ b/app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/create_spec_form.svelte
@@ -9,6 +9,7 @@
   import TaskSampleSelector from "$lib/utils/task_sample_selector.svelte"
   import type { TaskSampleExample } from "$lib/utils/task_sample_example"
   import type { Priority } from "$lib/types"
+  import PromptTypeSelector from "$lib/ui/run_config_component/prompt_type_selector.svelte"
 
   export let name: string
   export let property_values: Record<string, string | null>
@@ -27,6 +28,7 @@
   export let task_id: string
   export let task_sample_example: TaskSampleExample | null = null
   export let has_unsaved_manual_entry: boolean = false
+  export let selected_prompt_method: string = "simple_prompt_builder"
 
   let form_container: FormContainer
 
@@ -112,6 +114,12 @@
   {/each}
 
   {#if copilot_allowed}
+    <PromptTypeSelector
+      bind:prompt_method={selected_prompt_method}
+      {project_id}
+      {task_id}
+      description="Prompt sent to Kiln Pro for generating synthetic data. Pick a saved prompt to mirror your production prompt; otherwise the task instruction is used."
+    />
     <TaskSampleSelector
       {project_id}
       {task_id}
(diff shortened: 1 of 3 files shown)


chiang-daniel (commented):
The disabled button doesn't tell the user why it's disabled. FormContainer has a submit_data_tip prop that would work here, something like "Select a prompt to continue".

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/create_spec_form.svelte
```diff
@@ -79,6 +89,7 @@
   on:submit={handle_submit}
   bind:error
   bind:submitting
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/create_spec_form.svelte:91]: The disabled button doesn't tell the user why it's disabled. FormContainer has a submit_data_tip prop that would work here, something like "Select a prompt to continue".

PR #1427, opened 2026-05-22
feat: enable Kiln API tool to call SSE endpoints

leonardmq:
## What does this PR do?

Allow GET query params in Kiln API tool.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -41,7 +41,7 @@ class KilnApiCallTool(KilnTool):
 
 Endpoint paths, request schemas, response fields, and jq filters are defined in per-endpoint documentation — not here. Load the endpoint doc before calling.
 
-For SSE endpoints (text/event-stream), the tool consumes the stream until it closes (or a `data: complete` sentinel) and returns body = {"event_count": N, "complete": bool}. Individual event payloads are not returned."""
+For SSE endpoints (text/event-stream), the tool consumes the stream until it closes (or a `data: complete` sentinel) and returns body = {"event_count": N, "stream_complete": bool, "message": str}. stream_complete only means the stream ended — not that the underlying job succeeded; check the flow's status afterward. Individual event payloads are not returned."""
 
     @staticmethod
     def _build_parameters_schema() -> dict[str, Any]:
@@ -151,8 +151,10 @@ For SSE endpoints (text/event-stream), the tool consumes the stream until it clo
                         response_text = json.dumps(
                             {
                                 "event_count": event_count,
-                                "complete": complete,
-                            }
+                                "stream_complete": complete,
+                                "message": "The stream has completed. Note this is not a guarantee that the job has completed with zero errors - check the status of the flow you were triggering after this.",
+                            },
+                            ensure_ascii=False,
                         )
                     else:
                         raw = await response.aread()
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -97,9 +97,10 @@ For SSE endpoints (text/event-stream), the tool collects emitted events until th
         if not url_path.startswith("/"):
             raise ValueError(f"url_path must start with '/', got: {url_path}")
 
-        if "?" in url_path:
+        if "?" in url_path or "#" in url_path:
             raise ValueError(
-                "url_path must not contain a query string. Pass query args via query_params."
+                "url_path must not contain a query string or fragment ('?' or '#'). "
+                "Pass query args via query_params."
             )
 
         if body_str is not None and method in {"GET", "DELETE"}:
@@ -148,9 +149,15 @@ For SSE endpoints (text/event-stream), the tool collects emitted events until th
                     else:
                         raw = await response.aread()
                         response_text = raw.decode("utf-8", errors="replace")
-            except httpx.TimeoutException:
+            except httpx.TimeoutException as e:
+                # Read timeouts use the long SSE bound; connect/write/pool use
+                # the short one. Report whichever actually fired.
+                if isinstance(e, httpx.ReadTimeout):
+                    timeout_seconds = SSE_READ_TIMEOUT_SECONDS
+                else:
+                    timeout_seconds = SSE_CONNECT_TIMEOUT_SECONDS
                 raise TimeoutError(
-                    f"Request to {url_path} timed out after {SSE_READ_TIMEOUT_SECONDS}s"
+                    f"Request to {url_path} timed out after {timeout_seconds}s"
                 )
             except httpx.ConnectError:
                 raise ConnectionError(f"Could not connect to server for {url_path}")
@@ -188,9 +195,12 @@ async def _consume_sse(response: httpx.Response) -> tuple[list[Any], bool]:
 
     Each event's ``data:`` lines are joined with newlines per the SSE spec,
     JSON-decoded when possible, and appended. Non-data lines (``id:``,
-    ``event:``, ``retry:``, comments) are ignored. ``complete`` is True when
-    we observe the Kiln-specific ``data: complete`` sentinel or the stream
-    ends cleanly.
+    ``event:``, ``retry:``, comments) are ignored. ``complete`` is True only
+    when we observe an explicit ``data: complete`` sentinel — a clean stream
+    end does NOT set it. That sentinel is emitted by some Kiln endpoints
+    (eval runs, RAG indexing) but not others (chat, which has no such marker),
+    so ``complete=False`` is normal for streams that don't use it and does not
+    by itself imply truncation.
     """
     events: list[Any] = []
     data_lines: list[str] = []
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -9,6 +9,19 @@
...
 
+# httpx read timeout is per-read (idle), not a wall-clock cap: it resets every
+# time a chunk arrives. So this bounds the gap *between* reads, not total
+# request duration. For SSE that's the gap between events; for regular
+# responses it's the gap between body chunks. A multi-hour eval that streams
+# progress regularly never trips it — only this long of total silence does.
+# Kiln's SSE endpoints are chatty (and emit keepalive pings), so a low bound is
+# safe for them while still letting a genuinely stalled request fail reasonably
+# fast.
+READ_TIMEOUT_SECONDS = 120.0
```
tawnymanticore: _🤖 Auto-generated review comment (Claude Opus 4.7) — please verify before acting._

The justification here — "Kiln's SSE endpoints are chatty (and emit keepalive pings)" — doesn't hold up against the server code. None of the SSE endpoints emit keepalive pings: `CancellableStreamingResponse` doesn't, and the `eval_api` / `document_api` generators yield only per-progress `data:` events plus the final `data: complete`.

Progress is event-driven, not timed — `AsyncJobRunner.run()` yields a `Progress` only when a job *completes* (after one initial yield at startup). So the gap between events equals the duration of a single unit of work. With `concurrency=25`, one slow eval item (reasoning model / large generation), or a single large document's LLM extraction in RAG, can exceed 120s of silence — and the per-read timeout will then raise `TimeoutError` against a server that's still working fine. The initial yield protects the connect→first-event window, but not subsequent gaps.

Suggest one of: raise the read bound, add a real server-side keepalive/heartbeat to the SSE generators, or soften this comment to match reality (drop the keepalive-pings claim).

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -142,3 +196,49 @@ async def run(  # type: ignore[override]
...
+
+    Event payloads are intentionally not retained — a long stream (e.g. an eval
+    run) can emit thousands of events, and the caller only needs to know it
+    finished, not replay every progress update. We count events and detect the
+    explicit ``data: complete`` sentinel. ``complete`` is True only on that
+    sentinel — a clean stream end does NOT set it. The sentinel is emitted by
+    some Kiln endpoints (eval runs, RAG indexing) but not others (chat), so
+    ``complete=False`` is normal for streams that don't use it and does not by
+    itself imply truncation. Draining still blocks until the stream closes, so
+    the tool call returns only once the underlying operation has finished.
```
tawnymanticore: _🤖 Auto-generated review comment (Claude Opus 4.7) — please verify before acting._

`complete=True` reliably signals "the stream reached the sentinel," but a caller will likely read it as "the operation succeeded" — and on these endpoints those aren't the same thing. Both reach the sentinel on failure:

- **Eval**: per-item failures are caught in `AsyncJobRunner._run_worker`, tallied into `progress.errors`, and the run still finishes and emits `data: complete`. This tool keeps only `event_count`/`complete` and drops `errors`, so an eval where every item failed still returns `{event_count: N, complete: true}`.
- **RAG**: `run_rag_workflow_runner_with_status` catches `Exception`/`TimeoutError`, emits the error as a `data:` log event, then **still** yields `data: complete` — so `complete: true` even on catastrophic failure, and `_consume_sse` discards the error payload.

Net: the model can't distinguish a fully-successful run from a fully-failed one. If callers are meant to act on the outcome, consider surfacing error events — or at least an `errors`/`had_error` count — rather than discarding all payloads.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
tawnymanticore: I also agree with claude, vaguely. just wanted to zoom in how these timeouts actually work / how we can test this

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
tawnymanticore: so one very annoying thing about running Evals in Kiln today is the flakiness that sometimes happens. maybe for scoring 150 samples, we get a few errors due to provider issues. the caller needs to re-run the evals until all errors go away. so Kiln Assistant really needs a way to count the number of failures as the Compare UI will block results from being shown until all errors are gone

tawnymanticore (changes requested):

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: This is out of scope:
- what it means for a job to complete as far as the assistant is concerned is that the stream completes; this is intentional; we cannot collect thousands of events and pass them back to the LLM to determine if they are all good or not (that would flood the context window)
- source of truth is the files on filesystem, not the stream, so the model is supposed to determine progress (defined according to what it is about) by looking at the derived files
- every flow has its own logic (e.g. Eval, RAG, etc.); no unified shape for defining progress either so we cannot treat them generically

Streaming progress back to the bot is part of the bigger overhaul

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: 1. The timeout is about the amount of time elapsed without any message going through the transport; in the case of an eval or RAG, we have messages coming through every few seconds at most. That it hangs on client disconnects is also how SSE works throughout the app, by design.
2. Kind of true, but also need to find a balance because this timeout also applies to non-SSE endpoints where if we go too high (as I had it initially), a normal non-SSE endpoint that actually times out for real would hang until this timeout is hit (so if we have it at 5 minutes, it would hang for 5 minutes before giving up).

But maybe that is fine to crank up timeout since we have a few other slow endpoints

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -9,6 +9,19 @@
...
 
+# httpx read timeout is per-read (idle), not a wall-clock cap: it resets every
+# time a chunk arrives. So this bounds the gap *between* reads, not total
+# request duration. For SSE that's the gap between events; for regular
+# responses it's the gap between body chunks. A multi-hour eval that streams
+# progress regularly never trips it — only this long of total silence does.
+# Kiln's SSE endpoints are chatty (and emit keepalive pings), so a low bound is
+# safe for them while still letting a genuinely stalled request fail reasonably
+# fast.
+READ_TIMEOUT_SECONDS = 900.0
```
chiang-daniel, line 20: is 900s too long? If the SSE endpoints send keepalive pings (comment says they do), something like 120-300s would catch a stall after a few missed pings without the agent sitting idle for 15 min on a dead connection.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -142,3 +196,49 @@ async def run(  # type: ignore[override]
...
+                    saw_complete_sentinel = True
+                    break
+                event_count += 1
+            continue
+        if line.startswith(":"):
+            # SSE comment — ignore
+            continue
+        if line.startswith("data:"):
+            # SSE spec: strip exactly one leading space after "data:".
+            data_lines.append(line[5:].removeprefix(" "))
```
chiang-daniel, line 229: might be worth adding a comment for the implicit fallthrough here — lines like `event:`, `id:`, `retry:` just get silently dropped. that's correct per SSE spec but someone reading this later might wonder if we forgot to handle them. something like `# All other SSE fields (event, id, retry) ignored per spec` after the data block

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -79,44 +105,72 @@ async def run(  # type: ignore[override]
...
+                    is_sse = content_type.startswith("text/event-stream")
+
+                    if is_sse:
+                        event_count, complete = await _consume_sse(response)
+                        response_text = json.dumps(
+                            {
+                                "event_count": event_count,
+                                "complete": complete,
+                            }
+                        )
```
chiang-daniel, line 157: nit: add `ensure_ascii=False`

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: The timeout also applies to non-SSE endpoints (all GET / POST / etc.); if we do the Specs back and forth through the tool, that will time out at 120s

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: I added a `message` field now that explains what happened - i.e. stream completed and please check the status of [the thing | eval | rag | etc.] again after this. The bot should be seeing that and know what to do

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Cranked it up to 900s (15 min)

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Added

tawnymanticore (approved):

PR #1428, opened 2026-05-25
Deprecate Models

chiang-daniel:
Got email from Cerebras that 2 models will be retiring. 
```
This is a friendly reminder that effective May 27, 2026, the following public endpoints will be retired:

Llama 3.1 8B (llama3.1-8b)
Qwen 3 235B Instruct (qwen-3-235b-a22b-instruct-2507)
```

Run through Mike's `kiln-check-deprecation` skill. 

- Cerebras (1): llama3.1-8b (retiring May 27)
- Gemini API (2): Gemma 3n models removed
- OpenRouter (17): Claude 3.5/3.7/4.5, Grok 3/3-mini/4/4.1-fast, older Llama/Mistral/Qwen models
- Together AI (4): DeepSeek R1, Qwen3-235B-2507, Kimi K2.5, MiniMax M2.5
- Groq (3): Qwen3-32B, Kimi K2, Kimi K2-0905
- SiliconFlow CN (17): Qwen VL/Next/QwQ, DeepSeek R1 distills, GLM 4.6/Z1, Kimi K2, ERNIE 4.5, Pangu Pro MoE
- All clear: OpenAI, Anthropic, Fireworks AI, Vertex AI


libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -7239,22 +7276,23 @@ class KilnModel(BaseModel):
                 deprecated=True,
                 supports_data_gen=True,
                 structured_output_mode=StructuredOutputMode.json_instruction_and_object,
-                suggested_for_evals=True,
+                suggested_for_evals=False,
```
chiang-daniel, line 7279: @tawnymanticore @leonardmq going to set suggested_for_evals to false for models that were deprecated. Doesn't make sense to keep suggesting a deprecated model imo.

tawnymanticore: yep good catch. kind of a nightmare that we have suggested models that are deprecated, oof.

tawnymanticore (approved):

libs/core/kiln_ai/adapters/ml_model_list.py
leonardmq: I believe deprecated ones already get filtered out before making it to the dropdown selectors in the UI. 

Probably could add a test here that goes through every model and checks something like:

```py
if deprecated:
  assert suggested_anything = False
```

For every model -> provider in the model list

leonardmq (approved):

PR #1435, opened 2026-05-29
feat: background job system (tracked async jobs + jobs UI)

leonardmq:
## Summary

Adds a generic **background-job system** for the local Kiln app, plus the jobs UI that surfaces it. Jobs run as in-process asyncio tasks that are tracked, controllable (pause / resume / cancel / delete), observable over SSE, and **decoupled from any HTTP connection** — closing the UI or dropping the stream never stops a job.

---

## How it works — and the idea behind it

The defining property is that a **job record is transient, in-memory bookkeeping — never a source of truth.**

- **Transient / in-memory only.** The registry's entire store is an in-memory `{job_id → JobRecord}` index that starts empty on every process boot. Nothing is written to disk, there is no startup recovery scan, and no `interrupted` state. On restart, prior jobs are simply gone.
- **The real state lives in Kiln entities.** The authoritative state of any operation lives in the project entities a job touches — eval runs, task runs, etc. The job record is just a best-effort *view* over that truth, used for visibility and control.
- **Workers are idempotent and re-derive progress.** Every worker implements `compute_state(params)`, which reads the source-of-truth entities to figure out "what's already done." The registry reconciles its in-memory snapshot against that derived truth at every lifecycle transition (start / pause / resume) and on status reads — so the believed state self-corrects without persistence.
- **Recovery = re-trigger.** Because workers are idempotent and re-derive their progress, re-running after a crash, restart, or pause→resume converges to the same end state without duplicating side effects. There is nothing to rehydrate.
- **Decoupled from connections.** A job's supervising asyncio task is owned entirely by the registry — created at `create`/`resume`, ended only by completion or an explicit `cancel`/`pause`. The SSE stream and the `wait` endpoints *observe* a job; they never own it. Abandoning a `wait` or closing the browser leaves the job running.

The one deliberate exception to "no disk" is **per-run error messages**: diagnostic spillover that has no representation in Kiln entities spools to an ephemeral, best-effort JSON file in the OS temp dir (`{tempdir}/kiln_jobs/{run_id}.json`), keyed by a per-run UUID. It is non-authoritative — the OS may delete it, and every reader treats "missing" as "empty."

## Why we need it

This is a **prerequisite for Kiln Assistant auto mode.** For the assistant to run long, multi-step workflows on the local server side, those workflows can't be tied to the lifetime of an HTTP request or an open browser tab — the user must be able to close the UI while work continues, then come back and observe / control it.

The existing eval path illustrates the gap: the legacy blocking `run_comparison` SSE endpoint runs the eval *inside the HTTP request*, so closing the browser cancels it. A job in the new system runs independently of any connection. This layer gives us that connection-decoupled, trackable, controllable execution primitive, built generically (typed workers, opaque params/result, free-form `metadata`) so future consumers — starting with the assistant — adapt to it rather than the reverse.

## Implementation notes

**Positioning vs. `AsyncJobRunner`.** This layer sits *above* the existing `AsyncJobRunner` worker pool; it does not replace it. The composition for evals:

```
JobRegistry              (new — tracked lifecycle, in-memory, HTTP, SSE)
  └─ EvalJobWorker       (new — one tracked job per eval invocation)
       └─ EvalRunner     (existing — unchanged)
            └─ AsyncJobRunner   (existing — unchanged)
```

**Core layer** (`app/desktop/studio_server/jobs/`):
- `registry.py` — in-memory `JobRegistry` singleton: `{job_id → JobRecord}` index, global semaphore (default 10, FIFO by `created_at`, `KILN_JOBS_MAX_CONCURRENT`), one supervising task per job, `compute_state` reconciliation, per-run error-log routing.
- `models.py` — `JobRecord`, state machine, `JobContext`, and the `JobWorker` base contract (`run`, `compute_state`, `supports_pause`).
- `events.py` — in-process pub/sub bus feeding SSE.
- `error_log.py` — the one module that touches disk; append/read/delete per `run_id`, all best-effort.

**REST + SSE** (`api.py`): `create / list / get / result / errors / pause / resume / cancel / delete` under `/api/jobs`, a pure-observer SSE stream (`/api/jobs/events`), and a `wait` capability (`GET /api/jobs/{id}/wait`, `POST /api/jobs/{type}?wait=true`) that observes without owning.

**Workers**: `NoopJob` (validation/canary that exercises the full error-log path) and `EvalJob` (wraps `EvalRunner` unchanged; idempotent, so `supports_pause = True` — pause is a hard mid-run cancel, resume re-collects only unfinished items).

**Git-sync:** background eval writes come from a registry-owned task, *outside* the request-scoped `GitSyncMiddleware`. A new request-free `save_context_for_project(project_id, ...)` (in `app/desktop/git_sync/save_context.py`) mirrors the middleware's manager resolution so each `EvalRun` is committed/pushed for auto-sync projects, and is a no-op for the default non-auto-sync case. (Resolution logic is intentionally duplicated from the middleware — a delegating refactor would break the middleware's test patches; both copies carry a "keep in sync" note.)

**Frontend**: a global "Jobs" sidebar indicator (subtle spinner + count) that opens a wide jobs dialog; a shared jobs table reused by the dialog and the full `/jobs` page; a run-eval dialog to start eval jobs from the jobs panel.

**Specs**: full spec set + phase plans under `specs/projects/background_job_system/`.

## Out of scope / follow-ups

- The assistant / orchestration layer that consumes this system — separate, future work.
- Wiring the existing specs-page **"Run Eval"** button through the job system (with pause-on-disconnect) — a separate follow-up branch off this one.
- Cloud Run remoting / surviving the desktop-app process; full per-job log streaming; job DAGs/dependencies; plan-style multi-job approval.

## Test plan / how to try it

**Manual:**
1. Run the app and open the **Jobs** full page (`/jobs`), or click the **Jobs** indicator in the sidebar to open the jobs dialog.
2. Kick off a job — use the **run-eval dialog** to start an eval job, or trigger a `NoopJob` as a canary.
3. Watch it stream live, then exercise **pause / resume / cancel / delete** against the running job.
4. Confirm decoupling: close the tab (or drop the SSE stream) mid-run and re-open — the job is still running and reconciles to the right state.

**Automated:**
- [x] `uv run ./checks.sh --agent-mode` passes (ruff, ty, pytest, web lint/check/test/build, schema check)
- [x] Backend: registry lifecycle (pause/resume/cancel/delete, reconciliation, semaphore), SSE decoupling, wait (incl. abandon-doesn't-cancel), eval worker idempotency + `compute_state`, git-sync `save_context`
- [x] Frontend: jobs store/SSE, indicator states, clear-completed, run-eval dialog, dialog open-signal
- [x] Manual UI pass (visual): sidebar widget states, dialog layout, live pause/resume/cancel against a running server


app/desktop/studio_server/jobs/registry.py
```diff
@@ -0,0 +1,512 @@
...
+    # -- reconciliation ------------------------------------------------------
+
+    async def _reconcile(self, job: JobRecord, emit_on_change: bool) -> bool:
+        worker = self._workers.get(job.type)
+        if worker is None:
+            return False
+        params = worker.params_model.model_validate(job.params)
+        derived = await worker.compute_state(params)
+        if derived is None:
+            return False
```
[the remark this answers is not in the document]
leonardmq: Good catch — addressed in ad9c24116. `_reconcile` now wraps `worker.compute_state()` in `try/except`, logs via `logger.exception`, and falls back to the last known in-memory state (returns False) so a deleted/transiently-unavailable entity can no longer 500 `GET /api/jobs/{id}` (or crash `_supervise`'s initial reconcile).

app/desktop/studio_server/jobs/workers/eval.py
```diff
@@ -0,0 +1,136 @@
...
+            success = baseline_success + progress.complete
+            total = baseline_success + progress.total
+            error = progress.errors
+            await ctx.report_progress(
+                success=success,
+                error=error,
+                total=total,
+            )
+
+        return EvalJobResult(total=total, success=success, error=error)
```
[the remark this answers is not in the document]
leonardmq: Agreed — addressed in ad9c24116. `compute_state` now delegates to a synchronous `_compute_state_sync` run via `asyncio.to_thread`, so the blocking `runs(readonly=True)` scans (`os.scandir` + per-child `open`/`read`/`json.loads`) no longer execute on the event loop and SSE/progress updates stay responsive for large eval-run directories.

app/web_ui/src/routes/(app)/+layout.svelte
```diff
@@ -424,6 +428,20 @@
...
+            on:click={() => jobs_dialog.open()}
+            aria-label="Background jobs"
+          >
+            <div class="sidebar-icon opacity-60">
+              <JobsIcon />
+            </div>
+            In progress
+            <SidebarJobsIndicator variant="inline" />
+          </button>
+        </li>
```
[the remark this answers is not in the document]
leonardmq: Already resolved on the current branch — the button no longer carries `aria-label="Background jobs"` (it was reworked and the redundant label dropped), so its accessible name now comes from the visible label plus `SidebarJobsIndicator`'s own `aria-label` (the active-jobs count), which is exactly what you suggested.

app/desktop/git_sync/save_context.py
```diff
@@ -0,0 +1,66 @@
...
+        return None
+
+    bg_sync = GitSyncRegistry.get_background_sync(manager.repo_path)
+    if bg_sync is not None:
+        bg_sync.notify_request()
+
+    def factory():
+        return manager.atomic_write(context=context)
+
+    return factory
```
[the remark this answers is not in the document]
leonardmq: Not changing this. `GitSyncManager.atomic_write()` and the background poller both acquire the same per-clone-path write lock (`background_sync.py` does `async with self._manager.write_lock()`), so they're serialized — pausing the poller cannot postpone the job's own commit/push, as your own analysis concludes. The only effect of idle-pause is deferring remote→local polling, which resumes on the next request; adding a dedicated pause/resume API would be scope creep for a non-issue. The single `notify_request()` intentionally mirrors the HTTP middleware's per-request behavior.

app/desktop/studio_server/jobs/registry.py
```diff
@@ -0,0 +1,512 @@
...
+        task = self._tasks.get(job_id)
+        if task is None:
+            return
+        # cancel() returns True only if the request landed on a not-yet-done
+        # task — i.e. the cancellation is actually delivered to the worker. If
+        # it returns False the worker already finished naturally; we must not
+        # override that terminal result.
+        if task.cancel():
+            self._cancel_delivered.add(job_id)
+        try:
```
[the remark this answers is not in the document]
leonardmq: Intentional — this is the whole purpose of `_cancel_delivered` (see the comment at the call site and `_supervise`, per functional_spec §2). If our `cancel()` lands on a not-yet-done task, the job MUST end up cancelled/paused even if the worker catches `CancelledError` (or calls `uncancel()`) and returns normally. `cancel()` returning True means the request was delivered to a live task — exactly the signal we want; the registry deliberately does not trust the worker to re-raise. Recording delivery only on a confirmed `task.cancelled()` would let a cancel-swallowing worker be reported as SUCCEEDED, which violates the spec. The non-guarantee you cite is real but does not apply here, because we are not using the return value as proof of *termination* — only as proof the cancel request reached a running task.

app/desktop/studio_server/jobs/api.py
```diff
@@ -0,0 +1,362 @@
...
+            try:
+                event = await asyncio.wait_for(
+                    subscription.__anext__(), timeout=KEEPALIVE_SECONDS
+                )
+            except asyncio.TimeoutError:
+                yield ": ping\n\n"
+                continue
+            except StopAsyncIteration:
+                break
+            yield _format_sse(event)
```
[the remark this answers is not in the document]
leonardmq: Confirmed — this is a real bug. Reproduced it standalone: `asyncio.wait_for(agen.__anext__(), timeout)` *finalizes* the async generator when the timeout cancels `__anext__()`, so the very next `__anext__()` raises `StopAsyncIteration` and the SSE stream dies after the first 15s keepalive ping. Fixing it along exactly the lines you suggest (push the timeout into `subscribe()` and yield a `ping` event). Not landed in this push yet: the existing SSE tests in `test_api.py` were green only because the buggy stream self-terminated after one ping, so the now-correctly-infinite stream makes `test_sse_empty_snapshot` hang at teardown. Tracking the fix + test updates + a teardown/cancel verification as a follow-up before marking this resolved.

app/desktop/studio_server/jobs/events.py
```diff
@@ -0,0 +1,105 @@
...
+from pydantic import BaseModel
+
+from .models import JobRecord
+
+
+class JobEvent(BaseModel):
+    """A single bus event. Per-job events carry the full record (idempotent snapshot)."""
+
+    event: Literal["snapshot", "job", "deleted"]
+    data: dict[str, Any]
```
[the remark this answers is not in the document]
leonardmq: Part of the SSE keepalive fix tracked in the thread on `api.py` — `"ping"` will be added to the `JobEvent` Literal together with the `subscribe(timeout=...)` change. Leaving unresolved until that lands with updated tests.

app/desktop/studio_server/jobs/events.py
```diff
@@ -0,0 +1,105 @@
...
+        subscriber = _Subscriber(job_id, type_name, project_id)
+        self._subscribers.add(subscriber)
+        try:
+            snapshot = self._filtered_snapshot(subscriber)
+            yield JobEvent(
+                event="snapshot",
+                data={"jobs": [r.model_dump(mode="json") for r in snapshot]},
+            )
+            while True:
+                yield await subscriber.queue.get()
```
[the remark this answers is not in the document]
leonardmq: Agreed and this is the core of the fix — `subscribe()` will take an optional `timeout` and yield a `ping` JobEvent on `asyncio.wait_for(queue.get())` timeout, so the generator is never cancelled from the outside. Bundled with the `api.py`/`events.py` SSE work tracked above; leaving unresolved until it lands with updated tests.

app/desktop/studio_server/jobs/api.py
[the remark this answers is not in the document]
leonardmq: Fixed in 824edec78. Confirmed real (reproduced standalone: `asyncio.wait_for(agen.__anext__())` finalizes the generator on the first timeout, so the stream died after one 15s ping). Moved the timeout into `subscribe()` exactly as you suggested — it yields a `ping` `JobEvent` on `asyncio.wait_for(queue.get())` timeout, and `_event_stream` maps `ping` → `: ping` so the generator is never cancelled from outside. Tests had to be reworked: httpx `ASGITransport` runs the app to completion and buffers the whole body, so it can not drive a (now correctly) infinite stream — the old SSE tests only passed because the buggy stream self-terminated. They now drive `_event_stream`/`subscribe` directly, plus one HTTP-level test that ends the stream via the new `events.shutdown()`.

app/desktop/studio_server/jobs/events.py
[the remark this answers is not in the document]
leonardmq: Done in 824edec78 — `"ping"` added to the `JobEvent` Literal.

app/desktop/studio_server/jobs/events.py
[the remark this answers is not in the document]
leonardmq: Done in 824edec78 — `subscribe()` now takes `timeout` and yields a `ping` `JobEvent` on `asyncio.wait_for(queue.get())` timeout, so the generator owns its own keepalive and is never cancelled from outside. (Bonus: `subscribe()` also returns on a close sentinel from the new `shutdown()`, used to end open SSE streams on server shutdown / hot reload.)

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+  </div>
+{:else if $jobs.length === 0}
+  <div class="flex justify-center items-center min-h-[55vh]">
+    <Intro
+      title="No jobs yet"
+      description_paragraphs={[
+        "Long-running workloads show up here. Manage them from this page, or leave — they'll keep running in the background.",
+      ]}
+    >
+      <div slot="icon" class="w-12 h-12 text-gray-400" aria-hidden="true">
```
sfierro:
```suggestion
      <div slot="icon" class="w-12 h-12 aria-hidden="true">
```

sfierro: Intro icons are always black not gray

app/web_ui/src/lib/components/jobs_dialog.svelte
```diff
@@ -0,0 +1,30 @@
...
+    last_signal = $jobs_dialog_open_signal
+    dialog?.show()
+  }
+</script>
+
+<Dialog
+  bind:this={dialog}
+  title="Jobs"
+  width="wide"
+  subtitle="Background work for the current project. Jobs keep running even if you leave this page."
```
sfierro: "Jobs keep running even if you leave" is a little redundant maybe since it already specifies "background work" and they can just see that when they leave and come back it's still there.

```suggestion
  subtitle="Track background work for the current project."
```

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+    } finally {
+      result_loading = false
+    }
+  }
+
+  // The row's overflow menu: view actions first, then lifecycle actions.
+  function row_menu_items(job: JobRecord): FloatingMenuItem[] {
+    const items: FloatingMenuItem[] = []
+    if (has_result(job)) {
+      items.push({ label: "View results", onclick: () => open_result(job) })
```
sfierro:
```suggestion
      items.push({ label: "View Results", onclick: () => open_result(job) })
```

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+    }
+    if (has_errors(job)) {
+      items.push({ label: "View errors", onclick: () => open_errors(job) })
```
sfierro:
```suggestion
      items.push({ label: "View Errors", onclick: () => open_errors(job) })
```

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+{:else if !$synced}
+  <div class="w-full min-h-[50vh] flex justify-center items-center">
+    <div class="loading loading-spinner loading-lg"></div>
+  </div>
+{:else if $jobs.length === 0}
+  <div class="flex justify-center items-center min-h-[55vh]">
+    <Intro
+      title="No jobs yet"
+      description_paragraphs={[
+        "Long-running workloads show up here. Manage them from this page, or leave — they'll keep running in the background.",
```
sfierro: Two things here: this isn't showing up in a page it's a modal (Dialog), and usually we have an action button o the <Intro/> elements too, for now maybe just a "Docs & Guide" secondary button? Or don't add that yet but track in a follow up when we actually enable and have docs for it.

Proposed strings:

Intro title = "This project has no tracked jobs"
Intro description = "Long-running workloads will show up here for you to view status and results."

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+      </thead>
+      <tbody>
+        {#each $jobs as job (job.id)}
+          <tr>
+            <td class="whitespace-nowrap">
+              <div class="flex flex-col gap-1">
+                <span class="font-medium">{job_type_display(job.type)}</span>
+                <span class="font-mono text-xs text-gray-500">{job.id}</span>
+                <span class="text-xs text-gray-500"
+                  >{formatDate(job.created_at)}</span
```
sfierro, line 234: why not have a dedicated created at date column at the end like we do for other tables?

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+          <th class="text-right"></th>
+        </tr>
+      </thead>
+      <tbody>
+        {#each $jobs as job (job.id)}
+          <tr>
+            <td class="whitespace-nowrap">
+              <div class="flex flex-col gap-1">
+                <span class="font-medium">{job_type_display(job.type)}</span>
+                <span class="font-mono text-xs text-gray-500">{job.id}</span>
```
sfierro:
```suggestion
                <span class="font-mono text-xs text-gray-500">ID: {job.id}</span>
```

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+  <div class="flex flex-row justify-end mb-3">
+    <button
+      class="btn btn-xs btn-ghost"
+      disabled={clearing_completed || completed.length === 0}
+      on:click={clear_completed}
+    >
+      {#if clearing_completed}
+        <span class="loading loading-spinner loading-xs"></span>
+      {/if}
+      Clear completed
```
sfierro:
```suggestion
      Clear Completed
```

app/web_ui/src/lib/components/jobs_table.svelte
```diff
@@ -0,0 +1,346 @@
...
+    >
+      <div slot="icon" class="w-12 h-12 text-gray-400" aria-hidden="true">
+        <JobsIcon />
+      </div>
+    </Intro>
+  </div>
+{:else}
+  <div class="flex flex-row justify-end mb-3">
+    <button
+      class="btn btn-xs btn-ghost"
```
sfierro: XS looks a little too small, and ghost is not used much in kiln. do the gray background (no btn modifier) 
```suggestion
      class="btn btn-sm"
```

sfierro (changes requested):

sfierro:
Requested some changes, let me know if you want to push back on any and we can discuss! Otherwise ping me when they're all fixed and I'll approve!

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Addressed in https://github.com/Kiln-AI/Kiln/pull/1435/commits/aed0074ba38f8d24031fa0b86e4715052883a6af

app/web_ui/src/lib/components/jobs_dialog.svelte
leonardmq: Used your text in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Used your text in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Used your text in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Used your text in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Used your text in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Used your class in https://github.com/Kiln-AI/Kiln/commit/0cb68b22e0256d8a4b76998a340b4bee311448c0

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Two main reasons:
- ephemeral and realtime; time here is more about giving a sense of what is going on more or less right now (jobs don't survive a restart, so you won't have ancient jobs from weeks ago or whatever), we are not going to have filtering, sorting and all that searching type thing we need for real datamodels that accumulate over time
- this is a modal/dialog and so limited room there

No strong opinion about the look of the table either way - at this stage we don't really know if we like or not since it is not being used by anything real. First use case is chatbot in the next PR - we can see what makes sense in context then. In the meantime, this whole panel is hidden away without the dev flag

app/web_ui/src/lib/components/jobs_table.svelte
leonardmq: Took your title / description. 

Not sure we will have actual docs for this, maybe. 

To clarify also the purpose of this - it is mainly not for the user to view results (the results and errors are primarily made for the agent to have access to; though user can look at them if they want). For the user this is meant more as a way to manage the lifecycle of jobs and even mostly to stop them (e.g. they figure some eval is running and takes forever, they want to terminate it); otherwise they have no handle on anything running in the background so cannot stop them

sfierro (approved):

PR #1436, opened 2026-05-29
feat: run eval, finetune via background jobs

leonardmq:
## Summary

Routes the specs-page **Run Eval** flow through the new background job system, with a special **pause-on-disconnect** behavior scoped entirely to this endpoint (not baked into the job system itself).

- **One eval job per run-config**: per-row Run Eval → 1 tracked job; "Run All Evals" → N jobs (one per run-config). The new SSE endpoint creates the jobs and **aggregates** their progress back into the legacy `{progress, total, errors}` + `complete` SSE shape so `run_eval.svelte`'s counter dialog still works unchanged — only the URL is repointed.
- **Pause-on-disconnect**: on client disconnect the endpoint fires a single detached cleanup task that **cancels PENDING jobs first** (so they leave the queue before any slot frees), **then pauses RUNNING ones**. Race-free under a small concurrency cap where Run-All exceeds it. On normal completion (`data: complete` sent) the cleanup is skipped.
- **Approval-gated** (`agent_policy_require_approval("Run eval comparison?")`) to match the legacy `run_comparison` — it spends compute/credits.
- **No changes to `EvalJob`/`EvalJobParams` or `run_comparison`**: new module `app/desktop/studio_server/eval_jobs_api.py`, new endpoint `GET .../eval_config/{eval_config_id}/run_comparison_jobs?...` mirroring `run_comparison`'s params. Jobs are created via `job_registry.create("eval", {dict}, ...)` to avoid a circular import.

## Bonus fix — shared `iter_with_keepalive` helper

Found and fixed a latent SSE bug with the same root cause in the existing `/api/jobs/events` stream: `asyncio.wait_for(subscription.__anext__(), timeout)` cancels the bus async-generator on each timeout, terminating its `finally` (unsubscribe) and ending the stream after one ping. For the eval stream that would have **paused/cancelled live jobs on a still-connected client** during any quiet window. Fixed by extracting `iter_with_keepalive(subscription, timeout)` into `jobs/events.py` — a feeder-task pattern where the keepalive `wait_for` operates on a local queue, never on the subscription itself. Used by **both** the eval-jobs stream and the existing `/api/jobs/events` stream.

## Test plan

- [x] `uv run ./checks.sh --agent-mode` — passes
- [x] Backend: N-jobs-per-run-config + aggregate stream + `complete`; **disconnect cancels pending + pauses running** (with `max_concurrent < N` — the load-bearing case); deleted-job-unblocks-stream; **keepalive does not tear down the subscription** (real `JobEventBus` + monkeypatched short keepalive); normal-completion-doesn't-pause; empty-ids → 400; `all_run_configs` resolution
- [x] Frontend: `run_eval.svelte` builds the new URL for both explicit-ids and run-all
- [x] `make annotations` regenerated → new endpoint's annotation is approval-gated
- [ ] Manual UI: click Run Eval (per-row + Run-All) — counter updates and shows complete; closing the page mid-run pauses RUNNING jobs and cancels PENDING ones; both show in the project-scoped jobs widget


leonardmq:
Closing for now - superseded by narrower single giant PR: https://github.com/Kiln-AI/Kiln/pull/1517

PR #1438, opened 2026-06-01
Always persist task_run_trace on eval runs

tawnymanticore:
## Why

The trace is the only on-disk record of what the model actually saw during an
eval — the rendered system + user message that was sent to the provider. Today
that record is dropped for any eval whose `evaluation_data_type` is not
`full_trace`. So for `final_answer` and `reference_answer` evals (the common
case), `eval_run.kiln` has `task_run_trace: null`.

That gap matters most when something between the dataset item and the model
mutates the input — `input_transform` (the new Jinja2 layer on
`scosman/templates` / #1433) is the immediate example, but any future
provider-side prompt assembly will have the same property. Without the trace,
"did my transform fire?" and "what did the model literally read?" become
unanswerable after the fact; the eval row's `input` field stores the raw
dataset item, not what reached the model.

While debugging an input-feature ablation, I ran an eval, saw an unexpected
result, and had no way to confirm whether the configured Jinja template
actually rendered or was silently bypassed. The trace was being captured in
`result_task_run.trace` — it just wasn't being written.

## What

- `eval_runner.py`: drop the `evaluation_data_type == full_trace` gate. The
  trace is now persisted whenever `result_task_run.trace` is non-empty.
- `eval.py`: drop the `EvalRun` validator that rejected setting
  `task_run_trace` on `final_answer` evals. The full_trace-requires-trace
  invariant is preserved.
- `eval_runner.py`: use `pydantic_core.to_json` instead of `json.dumps` so
  real provider-SDK trace objects (which contain Pydantic `BaseModel`
  instances — litellm's `Message`, `Choices`, etc.) serialize correctly. The
  stdlib encoder choked on these the first time real data hit the path;
  Pydantic's encoder handles them, with a `default=repr` fallback that logs
  a warning and writes a degraded trace if even `to_json` can't.
- Tests updated to assert traces are preserved on `final_answer` evals; the
  parametrized matrix in `test_validate_output_fields_parametrized` reflects
  the relaxed validator.

`EvalDataType` still controls judging behavior (full_trace evals still
require the trace; reference_answer evals still write the reference answer
field). The change is scoped narrowly to "the trace gets written if we have
one to write."

## Notes

- Storage: traces are typically 5–15 KB each. On a 1000-row dataset across
  several Specs and run-configs this adds up, but evals already persist
  `intermediate_outputs`, `input`, `output`, and `task_run_usage` per row,
  and a trace is in the same order of magnitude. The verification value
  more than pays the disk cost.
- The two pre-existing pre-commit failures on this branch
  (`test_benchmark_get_model` timing-sensitive perf test,
  `test_adapter_reuse_preserves_data` depending on the lancedb/pandas
  transitive issue) are unrelated to this change and not affected by it.
  138 eval-related tests pass.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_runner.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_runner.py
@@ -232,14 +232,39 @@ class EvalRunner:
                 task_output = result_task_run.output.output
                 task_run_usage = result_task_run.usage
 
-                parent_eval = job.eval_config.parent_eval()
-                if (
-                    parent_eval
-                    and parent_eval.evaluation_data_type == EvalDataType.full_trace
-                    and result_task_run.trace
-                ):
-                    trace = json.dumps(result_task_run.trace, indent=2)
+                # Always persist the full trace when available. The trace is the
+                # only record of what the model actually saw — needed to verify
+                # input_transform rendering, system prompt content, and to debug
+                # eval failures after the fact. EvalDataType still controls
+                # judging behavior; it no longer gates trace persistence.
+                #
+                # Use pydantic_core.to_json so the encoder handles BaseModel
+                # instances and other non-stdlib-serializable types that can
+                # appear in real provider SDK trace objects. Fall back to a
+                # repr-based encoder on the off chance to_json can't handle
+                # something — a degraded trace is still more useful than
+                # crashing the whole eval job.
+                if result_task_run.trace:
+                    try:
+                        from pydantic_core import to_json
 
+                        trace = to_json(result_task_run.trace, indent=2).decode()
+                    except Exception as e:
+                        # Broad catch: pydantic_core.to_json can raise
+                        # PydanticSerializationError (subclass of RuntimeError)
+                        # plus the usual TypeError / ValueError. Falling back
+                        # to a repr-based encoder is always preferable to
+                        # crashing the eval job over an unprintable trace.
+                        logger.warning(
+                            "Falling back to repr trace encoding (%s) for dataset item %s",
+                            type(e).__name__,
+                            job.item.id,
+                        )
+                        trace = json.dumps(
+                            result_task_run.trace, indent=2, default=repr
+                        )
+
+                parent_eval = job.eval_config.parent_eval()
                 if (
                     parent_eval
                     and parent_eval.evaluation_data_type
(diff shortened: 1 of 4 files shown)


libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -232,14 +232,39 @@ async def run_job(self, job: EvalJob) -> bool:
...
+                #
+                # Use pydantic_core.to_json so the encoder handles BaseModel
+                # instances and other non-stdlib-serializable types that can
+                # appear in real provider SDK trace objects. Fall back to a
+                # repr-based encoder on the off chance to_json can't handle
+                # something — a degraded trace is still more useful than
+                # crashing the whole eval job.
+                if result_task_run.trace:
+                    try:
+                        from pydantic_core import to_json
```
leonardmq, line 249: Move import to the top?

Seems like the only reason `json.dumps` might not work is due to the `MessageUsage` being a Pydantic model. We should redo the typing of `.trace` at some point, too much union of `TypedDict` that may contain non-normal-serializable properties - but, different problem altogether

libs/core/kiln_ai/adapters/eval/test_eval_runner.py
```diff
@@ -775,11 +783,16 @@ async def run_task_and_eval(self, eval_job_item: TaskRun):
...
     eval_runs = mock_eval_config.runs()
     assert len(eval_runs) == 1
     saved_run = eval_runs[0]
-    assert saved_run.task_run_trace is None
+    assert saved_run.task_run_trace is not None
+    assert isinstance(saved_run.task_run_trace, str)
+    import json
+
+    parsed_trace = json.loads(saved_run.task_run_trace)
+    assert parsed_trace == mock_trace
```
leonardmq, line 795: At line 758 above, can you make `mock_trace` contain a `Usage` pydantic object so we can test the scenario that you used the `to_json` for:

```py
from kiln_ai.datamodel.usage import MessageUsage

mock_trace: list[ChatCompletionMessageParam] = [
    {"role": "user", "content": "test"},
    {
        "role": "assistant",
        "content": "response",
        "usage": MessageUsage(input_tokens=10, output_tokens=5, total_tokens=15),
    },
]
```

Then on this line here, compare the serialized traces (otherwise I think it won't work with the pydantic object):
```py
from pydantic_core import to_json

parsed_trace = json.loads(saved_run.task_run_trace)
assert parsed_trace == json.loads(to_json(mock_trace).decode())
```

leonardmq:
PR likely not to be merged due to competing Eval V2 coming up (but need to check on Eval V2 that we do persist the trace in every scenario)

PR #1443, opened 2026-06-02
Add a /api/statistics REST endpoint (confidence intervals, significance)

tawnymanticore:
## Summary

Adds a `POST /api/statistics` endpoint so Kiln agents can compute statistics reliably — standard errors, confidence intervals, and significance verdicts — instead of doing the arithmetic in their own reasoning (which is error-prone). The assistant reaches it through the **existing Kiln API tool** (`call_kiln_api`); there is **no new top-level built-in tool**.

> **History / review feedback:** This PR originally added the statistics helpers as a consolidated `statistics` *built-in tool*. Review (Mike + Leonard) found that too niche for a top-level tool — a built-in tool is always visible in the system prompt and works against progressive disclosure, since not every query needs significance testing. The final commit converts the same logic into a REST endpoint, surfaced only inside eval/spec/compare workflows via the skill's api_docs. The kiln_server side is Kiln-AI/kiln_server#210.

## The endpoint

`POST /api/statistics` with an `operation` discriminator — one route, mirroring the earlier single-tool design:

| `operation` | Purpose |
|------|---------|
| `proportion_ci` | Wilson score CI + standard error for a single proportion |
| `compare_proportions` | Difference of two **independent** proportions (Newcombe/Wilson) with a significance verdict; labeled conservative and points to `mcnemar_paired` when the two proportions come from the same items |
| `mcnemar_paired` | Exact McNemar test for **paired** binary outcomes over the same items — the correct, more powerful test for comparing pass/fail across two conditions on one dataset, especially at small discordant counts |
| `compare_paired` | Paired bootstrap CI + Wilcoxon signed-rank for continuous/count metrics (latency, tokens, cost) |

Each operation accepts either a fraction + sample size or raw counts, and returns LLM-ready JSON (the statistic, a confidence interval, a boolean `significant`, and a one-sentence interpretation). Bad input → `HTTPException(422)`.

## Implementation

- Pure math lives in `libs/server/kiln_server/statistics_lib.py` — **stdlib only, no numpy/scipy**, so it ships with no new dependencies.
- `libs/server/kiln_server/statistics_api.py` registers the route via `connect_statistics_api(app)` (the `spec_api.py` / `project_api.py` pattern), tagged `["Statistics"]` and marked `openapi_extra=ALLOW_AGENT` (pure stateless compute, no approval gate).
- The agent-policy annotation `post_api_statistics.json` (allow) is regenerated so the chat `call_kiln_api` precondition resolves the call; the frontend `api_schema.d.ts` is regenerated (a pre-commit hook enforces it).
- No `KilnBuiltInToolId` member and no `tool_registry` branch — the endpoint is the only consumer of the lib.

## Tests

`test_statistics_lib.py` + `test_statistics_api.py` (FastAPI `TestClient`, one route per operation). `ruff`, `ruff format`, and `pyright` all clean.


tawnymanticore:
Addressed the bot review in 76480cfe1:

- **Harden `calculate` AST call path** (the Major item): the `pow()` function path now enforces `_MAX_POW_EXPONENT`, large-magnitude *negative* exponents are guarded via `abs()`, and `fn(*args)` is wrapped so wrong-arity / non-numeric / math-domain calls (`min()`, `round(1,2,3)`, `factorial(1.5)`, `sqrt(-1)`) return a structured tool error instead of an uncaught `TypeError`/`ValueError`. Added tests for each.
- **`resolve_successes`**: catch `int()`/`float()` coercion errors with a clear message.
- **`wilcoxon_signed_rank_p`**: documented that the variance omits the tie-correction term (slightly conservative vs `scipy.stats.wilcoxon`).

Skipped, with reasons:
- **Pydantic model for `calculate` input**: the sibling built-in tools (`AddTool`, `DivideTool`, …) validate via plain kwargs/`TypedDict`, not Pydantic request models — kept consistent with them rather than introducing a one-off pattern.
- **`bootstrap_difference_ci` Bernoulli-loop perf**: core supports Python ≥3.10, so `random.binomialvariate` (3.12+) isn't available; the manual loop is required and is already documented as bounded (n≲200, ~150ms).

tawnymanticore:
Re-review feedback on the consolidated `statistics` tool addressed in 09cff8ddf:

- **Nullable `compare_paired` arrays**: `values_a`/`values_b` items are now `["number", "null"]` to match `_clean_pairs`, which drops pairs with a missing side (so a `null` no longer gets rejected at the schema boundary before `run()` sees it).

Skipped (same reasoning as before): **moving input validation onto Pydantic v2 models** for `statistics`/`calculate`. The sibling built-in tools (`AddTool`/`DivideTool`/…, and `KilnApiCallTool`) all validate via plain kwargs / `TypedDict`, not Pydantic request models — keeping these consistent with that established convention rather than introducing a one-off discriminated-union model here. Happy to revisit if the team wants to move all built-in tools onto Pydantic input models together.

tawnymanticore:
@gemini-code-assist thanks — fixed the one correctness item, and here's the rationale on the rest:

**Wilcoxon continuity correction (`stats_lib.py:297`, HIGH) — fixed in 8349e90dd.** The correction now shrinks `|W − mu|` toward 0 by 0.5 *clamped at 0*, so a perfectly symmetric sample gives `z = 0 / p = 1.0` (and it no longer flips sign when `|W − mu| < 0.5`). Added a `test_symmetric_sample_p_is_one` regression test.

The three perf/seed items are intentionally left as-is — each is bounded by this tool's operating range (eval scorecards, n ≈ 100–200), where the bootstraps run ~2M Python iterations / sub-second:

- **`bootstrap_difference_ci` Bernoulli loop (`:156`)** — this is the *secondary* cross-check in `compare_proportions` (Wilson/Newcombe is the headline). A C-speed fix wants `random.binomialvariate`, which is 3.12+; kiln-core targets 3.10, so there's no clean drop-in. Sub-second at eval n.
- **`paired_bootstrap_diff_ci` manual loop (`:246`)** — `rng.choices(diffs, k=n)` is reasonable, but it changes the seeded draw sequence and would break the pinned `test_paired_bootstrap_deterministic` values, for a sub-second gain outside the operating range. Not worth the churn here.
- **`_deterministic_seed` 0xFFFF mask (`:99`)** — only collides when a sample size exceeds 65,535; eval datasets are ~100–200 items, so it can't trigger in practice.

Happy to revisit the bootstrap perf if we ever expose this tool on large-n inputs.

tawnymanticore:
## ✅ Benchmark validation — kintsugi `kiln_chat` run

Drove the 4 statistics cases (one per `operation`) through production kiln-chat **after** the tool→endpoint conversion, and compared against the pre-conversion (built-in-tool era) baseline.

**Setup:** 4 cases `eval_significance_01–04` (`proportion_ci` / `mcnemar_paired` / `proportion_ci`×dims / `compare_proportions`), 3 turns each, scoring mode `driver_case`. Target model `glm_5_1 via fireworks_ai`. Endpoint era run `2026-06-05-stats-endpoint` vs tool era `2026-06-03-stats-significance`. **n_a=5, n_b=4, n_matched=4.**

**Quality held — no regression from the conversion:**

| Metric (matched 4 cases) | tool era | endpoint era | Δ |
|---|---|---|---|
| Criteria passed (deterministic) | 100% | 100% | +0.0pp · ns |
| Criteria passed (semantic) | 100% | 100% | +0.0pp · ns |
| Mode routing accuracy | 100% | 100% | +0.0pp · ns |
| API-failure cases | 25% | 0% | −25pp · ns |

**Tool-call decomposition confirms the conversion fingerprint** (paired-Δ/case, all `sig`): `call_kiln_api` 2→9 (**+6.5**), `other`/built-in-tool bucket 11.5→0 (**−10.75**), skill loads 4→5.5 (**+2.0**, the `post_api_statistics.md` precondition load). Statistics moved cleanly from the always-present built-in tool to the progressively-disclosed endpoint, and the agent correctly distinguished paired (`mcnemar_paired`) from unpaired (`compare_proportions`) by whether the configs shared eval items.

**Cost/latency:** median tokens/case 251k→227k (ns), cost/case $0.206→$0.176 (ns). User-perceived latency +120s/case (`sig`), decomposed ~41% provider (Fireworks/GLM, environmental) / ~59% agent-mix (extra `call_kiln_api` round-trips from load-doc-then-call); first-turn latency *improved* (4,380→2,840ms, `sig`) from the lighter system prompt. Acceptable cost of progressive disclosure at n=4.

**Two measurement artifacts (not code issues):** the `output_corruption` 100%→25% drop is mostly the old built-in tool-name leaving the trace; the remaining 1/4 flag is a single Greek glyph (`χ²`/`α`) — legitimate stats notation, a detector false-positive. The `knowledge_precision` dip was a benchmark-corpus gap (cases didn't list `post_api_statistics.md` as expected) and has been fixed in the corpus.

libs/server/kiln_server/statistics_lib.py
```diff
@@ -0,0 +1,503 @@
...
+        avg_rank = (i + j) / 2 + 1
+        for k in range(i, j + 1):
+            ranks[indexed[k][0]] = avg_rank
+        i = j + 1
+
+    w_plus = sum(r for r, d in zip(ranks, nz) if d > 0)
+    mu = n * (n + 1) / 4.0
+    sigma = math.sqrt(n * (n + 1) * (2 * n + 1) / 24.0)
+    if sigma == 0:
+        return None
```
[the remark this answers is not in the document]
tawnymanticore: its fine

libs/server/kiln_server/statistics_api.py
```diff
@@ -0,0 +1,485 @@
...
+def _require_n(value, label: str) -> int:
+    if value is None:
+        raise ValueError(f"'{label}' (sample size) is required.")
+    n = int(value)
+    if n <= 0:
+        raise ValueError(f"{label} must be a positive integer (got {n}).")
+    return n
+
+
+def _coerce_binary(values, label: str) -> list[int]:
```
leonardmq: Type `values: list[int | float]` instead of no typing here? Seems to be the only thing we pass into it

leonardmq (approved):

libs/server/kiln_server/statistics_api.py
tawnymanticore: done!

PR #1450, opened 2026-06-04
feat: run RAG via background jobs

leonardmq:
## Summary

- New `RagJobWorker` runs the existing `RagWorkflowRunner` under the job system, with idempotency keyed on `rag_config_id` (re-launches supersede in-flight predecessors).
- `GET /api/projects/{project_id}/rag_configs/{rag_config_id}/run` moved out of the shared `kiln_server.document_api` into a new desktop-only `rag_jobs_api`. Returns `{kiln_job_tracking_id}` immediately instead of streaming SSE; `build_rag_workflow_runner` stays in the shared lib as a pure helper.
- Frontend `rag_progress_store` no longer opens per-config `EventSource`s — it derives live progress, logs, and status from the project-scoped `\$jobs` store via a module-level reflector. Refreshing the configs list page now picks up runs spawned from any session.
- `JobContext` grows `report_display(primary?, secondary?)` and `report_metadata_patch(patch)` for workers that need to rewrite the row's Details lines or attach structured per-kind state (RAG stamps the full `RagProgress` snapshot under `metadata.rag_progress` so the existing four-bar dialog keeps working).
- `rag_runners.py` simplified: step runners yield absolute cumulative counts (was cumulative for the three doc-phases but per-batch deltas for indexing); workflow `update_workflow_progress` collapses from a four-arm match to a dict-mapped uniform assignment; logs accumulate on `RagProgress.logs` instead of being replaced per yield.
- **Re-snapshot at lock acquire + end of run.** Parallel RAG runs that share an extractor / chunker / embedder used to leave each other's counters stuck at the pre-lock baseline (sibling A's work skips through B's `collect_jobs`, so B's per-phase `step.success_count` stays 0). The workflow now recomputes `initial_progress` after acquiring the workflow lock; the worker also re-snapshots once more at end of run for the final reported numbers, carrying runtime error counts and accumulated logs from the workflow's tracked state.

## Test plan

- [ ] Trigger a RAG run from the configs list — row flips to Running, % advances, per-phase lines update under Details.
- [ ] Refresh the configs list page mid-run — row stays Running with up-to-date counters (the new reflector).
- [ ] Kick off 5+ RAG configs that share an extractor — all complete with accurate counters (no phantom "N documents could not be processed").
- [ ] Trigger the same RAG config twice — the second launch supersedes the first (no stacked rows in the jobs panel).
- [ ] Cause a real per-item failure (bad doc) — error appears under View Errors on both the RAG dialog and the job row.
- [ ] Run from MCP / assistant via `GET .../run` — receives `{kiln_job_tracking_id}` JSON.


leonardmq:
Dropping support for RAG migration altogether

PR #1455, opened 2026-06-08
test: add test for pydantic ser

leonardmq:
## What does this PR do?

PR going into https://github.com/Kiln-AI/Kiln/pull/1438 to add a test

Add a test to check the serialization works fine with `to_json` in the `EvalRunner` - the reason we do `to_json` there is because of `Usage` that is a Pydantic model so we should have a test to make sure it does indeed work in that case.

## Checklists


leonardmq:
Base https://github.com/Kiln-AI/Kiln/pull/1438 probably going to be closed without merging due to Eval V2 - dropping this one too

PR #1461, opened 2026-06-11
Ask User Question: suggested-answers prompt tool

leonardmq:
Adds an `ask_user_question` flow: the assistant can ask a question and offer **up to 5 suggested answers** (each a main answer line + a 1–2 line explanation), always with a **"Chat about this"** option to refine via free chat. Built on the auto-mode interception/SSE/tool-result machinery and composes with auto-mode (a question surfaces and the burst goes idle, flag stays on).

Targets `leonard/kil-692-assistant-auto-mode` (depends on the auto-mode engine).

## What's included
- **libs/core** — `AskUserQuestionTool` (`kiln_tool::ask_user_question`), schema `question` (required) + `suggested_answers` (≤5 × `{answer, explanation}`); signal-only `run()`, intercepted/never executed.
- **App server** — intercept the tool in `ChatStreamSession` (interactive) and `AutoChatRunner` (auto → IDLE, pending unresolved); `POST /api/chat/ask/answer` resolves the tool call via a clean `role:"tool"` message (pick → answer text; "Chat about this" → `{"choice":"chat"}`) and continues (interactive stream or auto-resume).
- **Web UI** — inline question card (options + "Chat about this"), one-click send, input gated while pending, reattach re-renders from snapshot/buffer.
- **Backend (`kiln_server`)** — registers `ask_user_question` as a client-visible tool + system-prompt guidance. **(Companion branch `leonard/ask-user-question` on the `kiln_server` repo — must be deployed together.)**

## UX decisions
- "Chat about this" → model asks an open follow-up, then free chat.
- Input blocked until the user picks an answer or "Chat about this".
- Picking an answer = one-click send (main line becomes the user's message).
- In auto-mode: question surfaces + burst pauses (stays armed); answering continues in auto.

## Known follow-up (mild)
After answering, the interactive continuation is consumed without a `tool-calls-pending` handler — so if the model's *next* step is an approval-required tool, the approval box wouldn't surface (mirrors the existing auto-mode decline-resume precedent). Worth wiring `resumePendingToolCalls` into both paths if that case is reachable.

## Pre-merge
The `kiln_server` branch points `kiln-ai`/`kiln-server` at local editable `../Kiln/libs/{core,server}`; repoint to a published Kiln rev (incl. this tool) before merging the backend branch.


leonardmq:
Closing due to rescoping - may bring back later

PR #1472, opened 2026-06-16
KIL-714 Show thinking level in run config details

sfierro:
Linear ticket: https://linear.app/kiln_ai/issue/KIL-714/thinking-level-is-not-shown-in-run-config-details

## Description

When selecting run config options, some models support selecting a thinking level. For these run configs, we now show the thinking level back when viewing the run config details. There are multiple places we show run config details (evals compare run configs, run config details page, etc.) — all of these render through the shared `getRunConfigUiProperties()`, so a single change covers them all.

### Handling of `thinking_level: str | None`
The ticket raised a question about what `None` means. Resolution:
- **`null`/`undefined`** → the row is omitted entirely. This covers both legacy run configs and models that don't support thinking levels — there is no meaningful value to show.
- **A stored value** (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`) → shown with a friendly label. Note that `none` is an explicit 'Off/None' choice the user made in the selector, so it is displayed as **None** rather than hidden.

### Changes
- Added `getThinkingLevelDisplayName(value)` mapping canonical values to friendly labels (None / Minimal / Low / Medium / High / Extra High / Max), with a capitalized fallback for unknown values.
- Added a **Thinking Level** row in the `kiln_agent` branch of `getRunConfigUiProperties`, placed after **Top P**, rendered only when `thinking_level` is non-null, with an explanatory tooltip.
- Added unit tests for the label mapping and the row behavior (shown with a value, `none` → None, omitted when null, ordering).

No backend changes; no schema regeneration needed.

### DIFF
--- a/app/web_ui/src/lib/ui/run_config_component/advanced_run_options.svelte
+++ b/app/web_ui/src/lib/ui/run_config_component/advanced_run_options.svelte
@@ -3,6 +3,7 @@
   import type { OptionGroup } from "$lib/ui/fancy_select_types"
   import type { StructuredOutputMode, InputTransform } from "$lib/types"
   import { structuredOutputModeToString } from "$lib/utils/formatters"
+  import { THINKING_LEVEL_INFO_DESCRIPTION } from "$lib/utils/run_config_formatters"
   import InputTransformSelector from "./input_transform_selector.svelte"
 
   export let temperature: number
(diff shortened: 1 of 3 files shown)


sfierro:
Thanks @coderabbitai — applied both nitpicks in 59779d3b6: switched the Thinking Level gate to an explicit `!= null` check, and added a test for the undefined/absent `thinking_level` case. Gemini had no feedback.

leonardmq (approved):

PR #1497, opened 2026-06-19
Archived tools shouldn't be able to run on existing config

chiang-daniel:
## What does this PR do?

Archiving a tool (MCP server, Kiln Task, RAG config, or Skill) previously only hid it from the tool picker. Run configs that already referenced the tool kept executing it, because run-time resolution ignored is_archived entirely.

This makes archived tools fail closed at run time. The guard lives in tool_from_id, which the agent run path (base_adapter.available_tools) and the single-tool MCP adapter both funnel through, and in _resolve_skills for skills (which resolve separately). RAG configs and Kiln Task tools are checked the same way inside tool_from_id.

Describing a historical run is different from executing it: the fine-tune dataset formatter and eval utils need to resolve tools that were archived after a past run. So tool_from_id and tool_definitions_from_ids take an allow_archived flag (default False), and those two callers pass True. Dataset export and evals over past data keep working.

Tests: MCP, Kiln Task, and RAG each get a pair (rejected by default, resolves with allow_archived); skills get a rejected-by-default test (skills have no allow_archived path). Existing tests whose mocks needed the new signature were updated. Diff coverage is 100%.

Follow-up KIL-724 adds the proactive UI notice, since this fix only surfaces the problem at execution time (a saved run config that references an archived tool errors on run, and saved configs aren't editable, so the resolution today is to unarchive the tool or create a new run config).

I, @aiyakitori, confirm that I have read and agree to the [Contributors License Agreement](https://github.com/Kiln-AI/Kiln/blob/main/.config/CLA.md).

## Checklists

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_utils/eval_utils.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_utils/eval_utils.py
@@ -39,7 +39,11 @@ class EvalUtils:
 
         for index, tool in enumerate(available_tools):
             try:
-                tool_object = tool_from_id(tool, task_run.parent_task())
+                # Past runs may reference now-archived tools; resolve them so we
+                # can describe what was available when the run happened.
+                tool_object = tool_from_id(
+                    tool, task_run.parent_task(), allow_archived=True
+                )
                 tool_definition = await tool_object.toolcall_definition()
 
                 tool_name = tool_definition["function"]["name"]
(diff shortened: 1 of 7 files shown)


leonardmq (changes requested):
Behavior should always assume the equivalent `allow_archived = False` to avoid messing with the immutability of configs and silently causing differences in execution based on current setup. 

Error messages seem pretty good though - so probably just need to make it not an arg?

PR #1525, opened 2026-06-27
Quality: Unbounded split on "::" in split_global_chunk_id crashes on multi-separator document IDs

### LINKED ISSUE #1524: fix(core): unbounded split on "::" in split_global_chunk_id crashes on multi-separator document ids
## Description

`split_global_chunk_id` uses `global_chunk_id.split("::")` without a maxsplit argument, then unpacks into exactly two variables. If a `document_id` contains the substring "::" (e.g. a UUID variant, namespaced ID, or any string with that sequence), `global_chunk_id()` would produce "doc::with::colons::0", and `split("::")` returns 4 elements, crashing with `ValueError: too many values to unpack`. The inverse function `global_chunk_id` doesn't guard against this either, so it's possible to create IDs that cannot be parsed back.

**Severity**: `high`
**File**: `libs/core/kiln_ai/utils/rag_utils.py`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

kumburovicbranko682-boop:
### Problem
`split_global_chunk_id` uses `global_chunk_id.split("::")` without a maxsplit argument, then unpacks into exactly two variables. If a `document_id` contains the substring "::" (e.g. a UUID variant, namespaced ID, or any string with that sequence), `global_chunk_id()` would produce "doc::with::colons::0", and `split("::")` returns 4 elements, crashing with `ValueError: too many values to unpack`. The inverse function `global_chunk_id` doesn't guard against this either, so it's possible to create IDs that cannot be parsed back.

**Severity**: `high`
**File**: `libs/core/kiln_ai/utils/rag_utils.py`

### Solution
`split_global_chunk_id` uses `global_chunk_id.split("::")` without a maxsplit argument, then unpacks into exactly two variables. If a `document_id` contains the substring "::" (e.g. a UUID variant, namespaced ID, or any string with that sequence), `global_chunk_id()` would produce "doc::with::colons::0", and `split("::")` returns 4 elements, crashing with `ValueError: too many values to unpack`. The inverse function `global_chunk_id` doesn't guard against this either, so it's possible to create IDs that cannot be parsed back.

### Changes
- `libs/core/kiln_ai/utils/rag_utils.py` (modified)

## Checklists

---

<details>
<summary>🤖 About this PR</summary>

This pull request was generated by [ContribAI](https://github.com/tang-vu/ContribAI), an AI agent
that helps improve open source projects. The change was:

1. **Discovered** by automated code analysis
2. **Generated** by AI with context-aware code generation
3. **Self-reviewed** by AI quality checks

If you have questions or feedback about this PR, please comment below.
We appreciate your time reviewing this contribution!

</details>

Closes #1524

### DIFF
--- a/libs/core/kiln_ai/utils/rag_utils.py
+++ b/libs/core/kiln_ai/utils/rag_utils.py
@@ -9,7 +9,7 @@ def global_chunk_id(document_id: str, chunk_idx: int) -> str:
 
 
 def split_global_chunk_id(global_chunk_id: str) -> Tuple[str, int]:
-    document_id, chunk_idx = global_chunk_id.split("::")
+    document_id, chunk_idx = global_chunk_id.rsplit("::", 1)
     return document_id, int(chunk_idx)
 
 


libs/core/kiln_ai/utils/rag_utils.py
```diff
@@ -9,7 +9,7 @@ def global_chunk_id(document_id: str, chunk_idx: int) -> str:
 
 
 def split_global_chunk_id(global_chunk_id: str) -> Tuple[str, int]:
-    document_id, chunk_idx = global_chunk_id.split("::")
+    document_id, chunk_idx = global_chunk_id.rsplit("::", 1)
```
leonardmq, line 12: Both `document_id` and `chunk_idx` are set by Kiln and never contains more than one occurrence of `::` - if either does, then we have a bug upstream and should be throwing. Turning this into an `rsplit` means we are now accepting an ID that violates the expected scheme.

That said, we could improve the error message - how about adding a step after the `rsplit` to check that neither part contains `::` and throw an explicit error message otherwise? We could also validate the `int(chunk_idx)` with a more user friendly error message as well

leonardmq (changes requested):

libs/core/kiln_ai/utils/rag_utils.py
leonardmq: Agreed on all of this. Two notes, and then a replacement PR.

**The reported bug isn't reachable.** `document_id` is a Kiln model ID — `str(uuid.uuid4().int)[:12]`, a 12 digit numeric string (`libs/core/kiln_ai/datamodel/basemodel.py:37`) — and it reaches `global_chunk_id()` via `doc.document_id` in `lancedb_adapter.py:131`. `chunk_idx` is an `int`. Neither half can contain `::`, so the multi-separator ID in the description can't be constructed.

**But there is a real reason to want the better error message.** The only caller is `libs/core/kiln_ai/tools/rag_tools.py:182`, which parses `result.document.id` — a string that has round-tripped through a third-party reranker API. Kiln generates the ID, but an external provider hands it back, so a provider that truncates, reorders or fabricates an ID is the one realistic source of malformed input. "The reranker returned an ID we never sent" beats `ValueError: not enough values to unpack` for that investigation.

**One simplification on the suggested fix:** `rsplit("::", 1)` followed by a "neither part contains `::`" check is exactly equivalent to a plain `split("::")` plus a length check — and the latter states the invariant directly (exactly one separator) in one branch instead of two. Worth noting that after `rsplit(..., 1)` the right-hand part can never contain `::` anyway, so only the left check would do anything.

Opened #1741 with that version plus tests. Closing this one in favour of it — the CLA attestation here is also unfilled, so it can't be merged as-is. Thanks for the report; the parsing was genuinely worth tightening, just in the strict direction rather than the permissive one.

---

leonardmq:
Closing in favour of #1741, which takes the strict version of this fix.

Thanks for the report — the parsing here was genuinely worth tightening, just in the opposite direction from what this PR does. Summarising the [review thread](https://github.com/Kiln-AI/Kiln/pull/1525#discussion_r3490444911):

- **The reported crash isn't reachable.** `document_id` is a Kiln model ID — `str(uuid.uuid4().int)[:12]`, a 12 digit numeric string (`libs/core/kiln_ai/datamodel/basemodel.py:37`) — reaching `global_chunk_id()` via `doc.document_id` in `lancedb_adapter.py:131`. `chunk_idx` is an `int`. Neither half can contain `::`, so the multi-separator ID in the description can't be constructed.
- **`rsplit` weakens an invariant we want enforced.** The scheme is exactly one `::` separator. An ID with more than one means something upstream is broken, and the right response is to throw. `rsplit("::", 1)` instead parses such an ID "successfully" into a `document_id` that doesn't exist — a loud crash becomes a silent wrong lookup in the RAG path.
- **The error message was worth improving, though.** The only caller (`libs/core/kiln_ai/tools/rag_tools.py:182`) parses IDs that have round-tripped through a third-party reranker API, so a provider echoing back a mangled ID is the one realistic source of malformed input here.

#1741 validates the format explicitly and raises an error naming the offending ID, with tests. It also leaves the CLA question moot — the attestation in this PR's description is unfilled, so it couldn't be merged as-is regardless.

---

PR #1568, opened 2026-07-13
Fix KIL-764 + KilnEvalHelpers extensions (KA2-Full port fixes)

tawnymanticore:
Re-pointing our KA2-Full / kintsugi-benchmark fixes at the new megabranch per Daniel's branch consolidation. Three commits, cherry-picked clean from `mike/kil-761-eval-runner-execute-multi_turn_synthetic-evalinputs-su-drive`; original implementation notes are on the tickets.

> **Scope notes:**
> - **KIL-759** (MCP LIFO teardown) moved to its own PR against main — #1569 — since it's a shipped core bug. Whichever lands first, the other side picks it up on the next merge.
> - **KIL-760** (tool raises → is_error messages) was dropped from this PR after scope review: the built-in `call_kiln_api` lane it primarily protects is not a user path (clients register `call_kiln_api` via local MCP, whose errors already return in-band), and across the 109-case benchmark cycle the loop-level wrapper never fired on the MCP lane. The implementation exists at `591f221ad` if the residual scope (MCP transport raises, weak-model malformed tool args) is still wanted on the ticket.

## Commits

- **KilnEvalHelpers: trace dialect + pairing + markdown links**: `get_tool_results` now matches `role: "tool"` (the shape Kiln actually stores via `ChatCompletionToolMessageParamWrapper`) as well as the `"tool_result"` variants — without this, code evals see zero tool results on real stored traces. Adds `get_tool_result_content`, tool-call/result pairing, `get_markdown_links`.
- **KilnEvalHelpers: health-metric primitives**: `get_usage_totals` (duck-types the MessageUsage pickle transport), `get_total_latency_ms`, `count_messages`, `get_error_tool_results`, `get_assistant_emitted_text`. Needed by the kintsugi health eval (KIL-764 context).
- **KIL-764 — `Allow custom-typed eval output scores; enforce finite score values`**: `custom` score type = any finite numeric, code-eval only (judge configs rejected on custom-score evals). Also closes the NaN validation hole for ALL score types — NaN previously passed range checks.

## Testing

Full `adapters/eval/`, `datamodel/`, `tools/` suites + litellm adapter tool tests on top of the megabranch after the rebase: **2417 passed, 0 failed** (skips are the API-key-gated tests).

## Notes

- No data-migration path is needed for any of this: the old per-EvalInput SU-config format from our frozen-branch work never shipped — the only project using it is our integration test bed, which we've already migrated by hand.
- Follow-up (not in this PR): drive-reuse across evals per (eval_input, run_config) — our `566e724db` conflicts with the megabranch's new `_run_v2_multi_turn_synthetic_job`, and it's a design question first. Details on KIL-761.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_helpers.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_helpers.py
@@ -66,14 +66,44 @@ class KilnEvalHelpers:
 
     @staticmethod
     def get_tool_results(trace: list[dict[str, Any]] | None) -> list[dict[str, Any]]:
-        """Return all tool-result entries from a trace."""
+        """Return all tool-result entries from a trace.
+
+        Matches OpenAI-format ``role: "tool"`` messages — the shape Kiln
+        actually stores (``ChatCompletionToolMessageParamWrapper``) — as well
+        as the ``role``/``type`` == "tool_result" variants.
+        """
         if not trace:
             return []
         return [
             entry
             for entry in trace
-            if entry.get("role") == "tool_result" or entry.get("type") == "tool_result"
+            if entry.get("role") in ("tool", "tool_result")
+            or entry.get("type") == "tool_result"
         ]
+
+    @staticmethod
+    def get_tool_result_content(
+        trace: list[dict[str, Any]] | None, tool_call_id: str
+    ) -> str:
+        """Return the text content of the tool result answering *tool_call_id*.
+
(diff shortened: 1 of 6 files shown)


chiang-daniel:
Landed via #1579 "Eval helper fixes + custom-typed eval scores" as a split commit series (bug fixes first, then the helper/datamodel features), with Mike credited as co-author on the derived commits. Closing this one in favor of that PR — thanks Mike for catching the dead tool_result matching and the NaN validation gap.

PR #1565, opened 2026-07-10
Artifact provenance: shared lineage submodel across compiled/tunable artifacts

scosman:
## What does this PR do?

Adds **artifact provenance** — a shared, optional `KilnArtifactProvenance` submodel recorded on Kiln's compiled/tunable on-disk artifacts, so we can answer "why does this artifact exist and what was it derived from?" It is agent/disk-facing metadata only — **no UI displays it anywhere**.

The submodel has three fields: `notes` (free-text "why"), `derived_from_ids` (lineage — sibling artifact IDs this was derived/cloned from), and `origin` (`"human"` or `"agent"`). Validation follows Kiln's "load any data, create perfect data" rule: strict on create, lenient on load (so forward/back-compat files never fail to load or 500 on read). Provenance is optional, but when a provenance object *is* present, `origin` is strictly required — fail-loud, never silently mislabeled. It is immutable post-create (PATCH/update models never accept it).

Built in five reviewed phases:

- **Phase 1 — core.** `KilnArtifactProvenance` value object + pure `validate_derived_from_ids` helper (raises `ValueError`, no FastAPI in `libs/core`), exported from `datamodel/__init__.py`; 42 unit tests over the full create-vs-load validator matrix.
- **Phase 2 — Tier-1 backend.** Field added to `Skill`, `Prompt` (not `BasePrompt`), `TaskRunConfig`, `CodeTool`; new `kiln_server/provenance_api.py::validate_provenance_or_400` wrapper (maps the core `ValueError` → HTTP 400); create-endpoint validation on all five paths (incl. TaskRunConfig's two MCP create paths) with archived-inclusive sibling-existence checks. Read/update paths made lenient-on-load so a forward-written provenance never 500s.
- **Phase 3 — Tier-1 clone wiring + client.** The four Tier-1 clone/create forms stamp lineage (clone → `{ origin: "human", derived_from_ids: [source.id] }`, fresh create → `{ origin: "human" }`); OpenAPI client regenerated; `KilnArtifactProvenance` added to `types.ts`; vitest coverage.
- **Phase 4 — Tier-2 backend.** Same pattern applied to `EvalConfig`, `Finetune` (threaded through `create_and_start`), `RagConfig`, `ExtractorConfig`, `ChunkerConfig`, `EmbeddingConfig`, `VectorStoreConfig`, `RerankerConfig`.
[...]

PR #1570, opened 2026-07-13
Artifact providence -> AGI

scosman:


scosman:
Closing in favor of #1571, which is the same change with the `agi-anyting_goes_into` merge conflict resolved.

The conflict was a single `__all__` collision in `libs/core/kiln_ai/datamodel/__init__.py` (both branches appended an export at the same spot — `KilnArtifactProvenance` here, `GradedClaim` on agi); #1571 keeps both. The resolution was done on a fresh branch off `claude/new-project-spec-0zdve1` rather than on that branch directly, so the production-path PR #1565 (`-> scosman/evals_v2`) is untouched.

---

PR #1581, opened 2026-07-15
Re-port KIL-761: driven eval conversations persist as TaskRuns, reused per (eval_input, run_config)

tawnymanticore:
## What

**This PR is now KIL-761 only** — the KIL-754 read-path fix it originally carried was superseded by `5501a424d` (the `expected_item_ids_for_eval` helper version is better than the per-endpoint switch this PR had, and covered `get_run_config_eval_scores` + the review nits in one shot). Rebased onto the current megabranch tip; the remaining two commits are the KIL-761 piece:

**Commit 1 — persist + reuse driven conversations.** The megabranch's eval-time drives are transient (`allow_saving=False`, no parent chaining): the only artifact is a serialized `task_run_trace` blob inside each EvalRun — driven conversations are invisible in the runs UI, untaggable, unratable, and every eval re-drives the same (eval_input, run_config) pair at full cost. This commit makes every turn persist via the adapter with `parent_task_run` chaining exactly like an SU batch drive; the adapter stamps `output.source.run_config_id`; the runner tags the leaf `sei_<eval_input_id>` (pair reuse) + `synthetic_eval_drive_<eval_id>` (cleanup sweeps) and reuses a stored conversation for the same pair instead of re-driving. Reuse never crosses run configs, so per-run-config comparison integrity is preserved.

**Commit 2 — single-flight concurrent drives** (from PR review): a per-(eval_input, run_config) `asyncio.Lock` covers the scan→drive→tag path so concurrent jobs (e.g. two eval configs over one input) share one drive. Test proves one drive across two concurrent jobs; verified to fail with the lock neutralized.

## Design note

This deliberately reverses the transient-drive decision — the motivating case is scoring the SAME conversation with a code judge and an LLM judge and joining scores per case, which transient drives make impossible (each eval re-drives, so scores come from different samples and can't be correlated). Persisting per (eval_input, run_config) also enables golden tagging/rating of driven conversations for judge calibration, and halves drive cost for the code-eval + judge-eval pair. If you'd rather implement KIL-761 your own way, happy to close this in favor of yours — the test cases may still be useful.

## Testing

- `synthetic_user/` + `adapters/eval/` + `test_eval_api.py`: 831 passed on the rebased tip (tests: leaf persisted + tagged, reuse hit for same pair, re-drive for different run config, parent chaining in the drive loop, concurrent single-flight).

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_runner.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_runner.py
@@ -625,11 +625,14 @@ class EvalRunner:
 
         The run config under evaluation drives the agent while the eval's
         multi_turn_drive_config plays the synthetic user, so each run config
-        gets its own fresh conversation — the property that makes run-config
-        comparison meaningful for multi-turn. The drive is transient (nothing
-        persisted); for full_trace evals the EvalRun record carries the
-        serialized conversation when scoring succeeds, otherwise the driven
-        conversation is not retained.
+        gets its own conversation — the property that makes run-config
+        comparison meaningful for multi-turn. Driven conversations persist as
+        real TaskRuns keyed by (eval_input, run_config): the leaf carries a
+        `sei_<eval_input_id>` tag and the adapter stamps
+        `output.source.run_config_id`, so any eval scoring the same pair
+        reuses the stored conversation instead of re-driving it (KIL-761).
+        For full_trace evals the EvalRun record additionally carries the
+        serialized conversation when scoring succeeds.
         """
         drive_config = self.eval.multi_turn_drive_config
         if drive_config is None:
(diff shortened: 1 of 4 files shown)


tawnymanticore:
Review feedback addressed in 2238279c8:

- **get_run_config_eval_scores missing EvalInput support** (CodeRabbit, major): fixed — same `id_source` switch as the other two summary endpoints, plus an EvalInput-backed test. Good catch: this endpoint was missed in the original KIL-754 fix as well.
- **Concurrent drive race** (Gemini + CodeRabbit, major): fixed — per-`(eval_input, run_config)` `asyncio.Lock` single-flights the scan→drive→tag path, so concurrent jobs for the same pair reuse the first drive. The new test runs two eval configs' jobs concurrently and asserts exactly one drive; it was verified to fail with the lock neutralized. Note the lock is per-runner-instance: cross-process/cross-runner dedup still relies on the disk rescan inside the lock, which closes the window for sequential invocations but not two literally simultaneous sweeps — considered acceptable for now.
- **Empty-filter 400 message** (Gemini, nit): fixed — message now matches the filter type.
- **eval_input filter cache in eval_results_summary** (Gemini + CodeRabbit, nit): fixed — separate typed cache mirroring `dataset_ids_cache`.
- **Diff coverage**: the previously-uncovered `eval_input` branches in `score_summary` and `eval_results_summary` are now exercised by EvalInput-backed tests for all three summary endpoints.

chiang-daniel:
Thanks for the thorough re-port — and good catch (via CodeRabbit) that `get_run_config_eval_scores` was missed in the original fix too. Both halves of this are covered on our side though, so asking to close this PR rather than merge it.

**KIL-754 half (commits 1+3):** implemented on the megabranch in today's pre-bug-bash session — staged, landing with my next push. Same source-aware keying as yours, slightly wider surface: all three summary endpoints plus `get_eval_progress` unified behind one `expected_item_ids_for_eval` helper, the raw 400 copy and the eval-UI dead ends fixed, and the paid e2e harness grew a read-path leg asserting score_summary / eval_results_summary / run-config eval_scores serve real numbers for EvalInput-backed evals. Full detail is on KIL-754 (closed earlier today with a note about this overlap).

**KIL-761 half (commit 2):** the need is real and it's queued as its own phase immediately after the bug bash — but shaped as EvalRun-as-trace-store + drive fingerprint rather than persisted TaskRun chains. Why not the TaskRun-persistence shape:

- Reuse keyed only on the `sei_` tag + `run_config_id` goes stale the moment an eval's `multi_turn_drive_config` is edited (turns, SU model) — old-config conversations get silently rescored. The fingerprint (drive config + run config + seed) exists for exactly that.
- Persisted driven chains land in the dataset (runs UI, `all`-type filters, delete-cascade interactions) — the pollution we deliberately avoided when eval-time drives were made transient.
- The paid e2e harness asserts that transience today (test_eval_builder_e2e_paid.py:839) — it goes red on this branch as-is.

Your visibility/ratability point (human-rating driven conversations for the benchmark) is a genuine gap in the trace-store shape and I've folded it into that phase's design inputs so it doesn't get lost. Until it lands, running your branch locally for KA2 is the pragmatic move — the reuse semantics you need arrive with the phase, fingerprinted.

KIL-761 stays open tracking that work.

chiang-daniel:
Update: the read-path fix is on the branch now — pull `dchiang/multiturn-eval-megabranch` @ `5501a424d` (score summaries + cross-eval summaries for EvalInput evals; `09241f53d` cleans up the related UI dead ends, `4637d5fa3` adds the harness read-path assertions). And the KIL-761 work is being pulled forward to the next session rather than waiting for the bug bash, with your ratability requirement in scope — so the fingerprinted version of commit 2 shouldn't be far behind. This PR can close whenever you've confirmed the pull works for you.

tawnymanticore:
Superseded by #1582 — same KIL-761 behavior, rewritten as two concise commits on the current megabranch tip (this branch's history was WIP iteration).

PR #1582, opened 2026-07-15
KIL-761: persist eval-time driven conversations, reuse per (eval_input, run_config)

tawnymanticore:
## What

Two commits implementing KIL-761 on the current megabranch tip:

1. **Drives persist as chained TaskRuns** — `drive_case_for_eval` saves every turn through the adapter with `parent_task_run` chaining (exactly like an SU batch drive) and stamps `output.source.run_config_id` via a new `task_run_config_id` param. Driven conversations land in the dataset: visible in the runs UI, taggable, ratable.
2. **Reuse per (eval_input, run_config)** — a new `EvalRunner._stored_or_driven_leaf` tags fresh leaves `sei_<eval_input_id>` (+ `synthetic_eval_drive_<eval_id>` for cleanup sweeps) and reuses any stored leaf whose tag and `run_config_id` match, so every eval scores the identical conversation. A per-pair `asyncio.Lock` single-flights concurrent jobs (two eval configs over one input drive once). Reuse never crosses run configs — per-run-config comparison integrity is preserved.

## Why

The motivating case: score the SAME driven conversation with a deterministic code eval and LLM judges, and join the scores per case. With transient per-eval drives, the judges and the code eval score different samples — the scores can't be correlated — and each additional eval re-pays the full drive cost. This also unblocks golden tagging/rating of driven conversations for judge calibration.

This deliberately replaces the transient-drive design; if you'd rather build KIL-761 differently, happy to close in favor of yours — the tests may still be useful.

## Testing

`synthetic_user/` + `adapters/eval/`: 713 passed. New coverage: parent chaining in the drive loop, leaf persisted + tagged, reuse hit for the same pair, fresh drive for a different run config, concurrent single-flight (verified to fail with the lock removed).

Replaces #1581 (WIP iteration; will be closed and its branch deleted).

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_runner.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_runner.py
@@ -1,7 +1,8 @@
+import asyncio
 import json
 import logging
 from dataclasses import dataclass
-from typing import AsyncGenerator, Dict, List, Literal, Set
+from typing import AsyncGenerator, Dict, List, Literal, Set, Tuple
 
 import litellm
 
@@ -25,6 +26,7 @@ from kiln_ai.datamodel.eval import (
     EvalRun,
     EvalScores,
     EvalTaskInput,
+    MultiTurnDriveConfig,
     MultiTurnSyntheticEvalInputData,
     SingleTurnEvalInputData,
     SkippedReason,
@@ -114,6 +116,10 @@ class EvalRunner:
         self.eval = target_eval
         self._skills: SkillsDict = self._preload_skills()
         self._save_context: SaveContext = save_context or default_save_context
+        # Single-flight guard for multi-turn drives: concurrent jobs for the
+        # same (eval_input, run_config) pair (e.g. two eval configs scoring
+        # the same input) must not both drive the conversation.
+        self._drive_locks: Dict[Tuple[str, str], asyncio.Lock] = {}
 
     def collect_tasks(self) -> List[EvalJob]:
         if self.eval_run_type == "eval_config_eval":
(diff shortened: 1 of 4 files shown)


tawnymanticore:
Review feedback addressed in d68c0b31b:

- **In-memory pair cache** (Gemini): added `_driven_leaves` keyed (eval_input_id, run_config_id), checked before and after lock acquisition; repeat jobs skip the dataset rescan. New test proves the second job succeeds with `Task.runs` patched to raise.
- **Orphaned turns on mid-drive failure** (CodeRabbit, major): every driven turn now carries `synthetic_eval_drive_<eval_id>` from creation via `AdapterConfig.default_tags`, so partial chains from failed drives stay discoverable by cleanup sweeps. Full rollback-on-failure was deliberately not attempted — deleting runs on error is riskier than sweepable orphans. The `sei_` pair tag still marks only the successful leaf, so a partial chain can never satisfy the reuse scan or get scored.
- **Reuse-key test assertion** (CodeRabbit, minor): the fresh-trace test now stamps and asserts `output.source.run_config_id` on the persisted leaf, plus asserts `default_tags`.

Also added a sampling-semantics note to the `_stored_or_driven_leaf` docstring: the stored conversation is pinned per pair (re-runs re-score the same sample); untag/delete the stored leaf to draw a fresh one; first-class N-samples-per-pair would add a sample ordinal to the reuse key — proposed as a follow-up rather than scope creep here.

tawnymanticore:
Closing — superseded by the upstream implementation (f0d024a50 + 14fde35cb, Phase 6.9 trace reuse): the drive-fingerprint key is strictly better than this PR's sei_ tag (content changes auto-invalidate reuse; turns changes re-drive), task-wide reuse covers judges on separate Evals, and always-persist task_run_trace achieves same-trace multi-judge scoring without dataset writes. One workflow we lose vs this PR: driven conversations aren't TaskRuns, so they can't be tagged/rated in the runs UI — that's really KIL-758's territory (render/rate EvalRun traces) and worth a follow-up when golden human ratings return at corpus scale.

PR #1586, opened 2026-07-16
Eval V2: Review only (do not merge): eval trace reuse — drive once, score with every config

chiang-daniel:
Review vehicle for the two trace-reuse commits already on dchiang/multiturn-eval-megabranch (f0d024a50, 14fde35cb), pinned between marker branches so comments can thread inline on the exact diff. Do not merge — requested changes will land as follow-up commits on the megabranch and this PR closes unmerged when review is done.

What it does: multi-turn EvalInput evals drive each conversation once per (scenario, run config), and every v2 eval config task-wide scores that same stored conversation. EvalRun gains an additive drive_fingerprint ("v1:" + sha256 of drive config + resolved run-config properties + scenario content; reference data and judge properties are excluded — they shape judgment, not the drive). Traces now always persist on v2 EvalRuns, including judge skips. This supersedes #1438 "Always persist task_run_trace on eval runs", which removed the same full_trace/success-only gates — credit to that PR for the direction; legacy g_eval and golden-calibration paths are untouched.

Reuse mechanics: index built once per runner invocation (O(1) per job), healthy traces only (full turn count, ends with assistant text), reuse never crosses run configs, first-write-wins with a deterministic pick, tombstones never satisfy reuse, and zero TaskRuns are written by eval-time drives. The paid harness gained a reuse leg: a second judge config re-running run_comparison consumes stored conversations byte-identically (equal fingerprints, no new drives) with the transience assertions unchanged.

Diffs to eval_runner.py and datamodel/eval.py were kept deliberately surgical.

Data model, before and after:

```mermaid
flowchart LR
  subgraph BEFORE ["Before"]
    direction TB
    EI1["EvalInput (scenario)"] -->|"fresh drive per eval config"| D1["transient TaskRun chain (discarded)"]
    D1 --> ER1["EvalRun
    task_run_trace: only if full_trace eval AND scoring succeeded
    no fingerprint — stored traces unfindable/unreusable"]
  end
  subgraph AFTER ["After"]
    direction TB
    EI2["EvalInput (scenario)"] -->|"one drive per (scenario, run config) — still zero TaskRuns"| D2["transient drive"]
    D2 --> ER2["EvalRun
    + drive_fingerprint: 'v1:' + sha256(drive config, resolved run-config properties, scenario content)
    task_run_trace: always persisted on v2 (judge skips included)"]
    ER2 -->|"reuse key: (drive_fingerprint, task_run_config_id)
    task-wide, healthy traces only, never across run configs"| J2["every other v2 eval config judges the same stored conversation"]
  end
```

Field-level delta on EvalRun (datamodel/eval.py): one additive optional field drive_fingerprint (None on all pre-existing and non-driven records — no migration), plus classmethod EvalTaskInput.from_eval_input_trace to build judge input from a stored conversation instead of a live TaskRun. Everything else on EvalRun is unchanged; task_run_trace changes are behavioral (when it is populated), not structural.

### DIFF
--- a/app/desktop/studio_server/test_eval_builder_e2e_paid.py
+++ b/app/desktop/studio_server/test_eval_builder_e2e_paid.py
@@ -60,7 +60,12 @@ calls, so the harness starts at Step 4 with the spec text already "written":
                        per config. run_calibration validates the judge
                        against the golden answer key over the STORED rated
                        traces, and the harness reports the judge-vs-human
-                       agreement.)
+                       agreement. A REUSE leg then adds a second judge
+                       config and re-runs the comparison: the drive was
+                       paid once per (scenario, run config), so the second
+                       judge must consume the stored conversations —
+                       byte-identical traces, equal drive fingerprints,
+                       zero new drives — at judge-only speed.)
   Step 8   READ        GET .../score_summary, .../eval_results_summary,
                        .../run_configs/{id}/eval_scores
                        (The endpoints the post-save spec detail page,
@@ -90,8 +95,9 @@ Requirements (hard failures, never skips — a broken pipeline must be loud):
 Cost per run: one plan, one SU batch call, the 4x2 drive + judge + claim
 builder, one refine call at save, plus the runner's work over the saved
 eval: a fresh 4x2 re-drive per run config (two configs) with a judge call
-each, and the golden calibration judge calls — ~70 small model calls,
-still cents.
+each, the golden calibration judge calls, and the reuse leg's 8 judge-only
+calls (no drives — that's the assertion) — ~80 small model calls, still
+cents.
 
 A second paid test (`test_eval_builder_pipeline_tools_e2e`) drives a
 tool-calling task via its SAVED run config (2x2, built-in calculator tools)
@@ -102,6 +108,7 @@ cost.
 
 import json
 import os
+import time
 import warnings
 
 import httpx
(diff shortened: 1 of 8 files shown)


chiang-daniel:
Closing this review as superseded; the commits stay in branch history.

This PR proposed fingerprint-based reuse of eval-time multi-turn drives: hash everything that shapes a conversation (synthetic-user drive config, resolved run config properties, scenario content) onto each EvalRun, always persist the driven conversation, and let any later judge on the task reuse a healthy stored trace matching the fingerprint and run config.

The design goals now land in a stronger shape via #1676 "Migrate eval runs from inline traces to trace pointers" (merged into scosman/evals_v2, completing the run/score split). Eval-time generations persist as real TaskRuns, flagged with eval_source and excluded from the dataset, durable before scoring, so a judge failure can never lose a paid generation. Reuse goes through a live TraceIndex keyed on (source_type, source_id, run_config_id) with per-key locks, so concurrent judges serialize on one generation instead of both paying. The key omits the eval config, which gives the task-wide cross-eval sharing this PR's second commit added. No fingerprint is needed because the drive config now rides each eval input (MultiTurnSyntheticEvalInputData.drive_config, stamped at save, immutable), so item id plus run config id capture drive identity, and config immutability covers staleness.

Three things here do not exist upstream and move into the adoption work rather than dying with this PR. Multi-turn support itself: the merged runner skips multi-turn items, and our re-drive gets rebuilt as a generate callable under TraceIndex.get_or_create. Trace health vetting: upstream has no content gate, so the per-item gate (expected turn count from the item's drive config, and the conversation must end with assistant text so a truncated record cannot promote a mid-conversation reply to the final answer) gets built on the re-drive path. And the paid end-to-end reuse proof (second judge consumes the stored conversation, no re-drive, no new generation spend) gets re-expressed against scored_run_id pointers instead of fingerprints and byte-identical trace copies.

Data note: records written from this branch carry a drive_fingerprint field. Builds without the field load them fine and drop it on rewrite, so it is inert; the run/score split migration sheds it naturally.

#1645 "evals: persist synthetic-user model usage on EvalRun (synthetic_user_usage)" touches adjacent code but depends on the branch, not this review, and is being reworked separately to store drive cost on the trace side.

PR #1603, opened 2026-07-21
Builder Step 4 adopts the shared /generate batch-plan components

chiang-daniel:
Per Sam's demo feedback: builder Step 4 now renders the /generate kiln_pro batch-plan components verbatim instead of the builder-local fork. kiln_pro_batch_plan handles plan approval; kiln_pro_plans_table (its plain-text status column) is the live board while the pipeline streams. The forked batch_plan_approval + plan_prompts_table are deleted (-394 lines). Wizard chrome (Back / Continue to Review) stays outside the shared components since they have no slots for it.

Consciously traded away with the fork: per-row edit-in-place, colored status pills, and pagination. Plan edits are now delete-row or regenerate, same as /generate.

No API or event changes — pure component swap. The kiln_pro components themselves are untouched and byte-identical to main, so the PR-A assembly should dedupe them entirely (only the +page.svelte hunks and the two deletions are new). Checks green.


chiang-daniel:
Closing per Daniel — this lands as a phase commit; PR/merge plan comes later at the assembly cut.

PR #1612, opened 2026-07-22
fix(evals): V2 LLM judges persist no reasoning — ask for it in the schema

tawnymanticore:
## The bug

Judge EvalRuns are meant to record **why** a verdict was reached: `intermediate_outputs` is persisted on the EvalRun, and `llm_judge_result.svelte` has a **"View reasoning"** control that reads it (`intermediate_outputs?.reasoning || ?.chain_of_thought`).

For **V2 `llm_judge` configs that record is empty**, so no verdict can be audited. The control never appears.

### Cause

The V2 judge invokes the adapter with `prompt_id=PromptGenerators.SIMPLE` — a single call, no chain-of-thought turn — so nothing ever populates `intermediate_outputs`. The sole exception is a reasoning-capable provider returning native `reasoning_content` (`litellm_adapter._extract_reasoning_to_intermediate_outputs`), which is strategy-independent.

The V1 G-Eval / LLM-as-Judge lane hardcodes `SIMPLE_CHAIN_OF_THOUGHT` ("We always use Simple COT for G-Eval and LLM as Judge") and therefore always records a `chain_of_thought`.

`d1209663e` wired `intermediate_outputs` from the judge through `V2EvalResult` to `EvalRun` persistence (D31). The plumbing is correct — it's just connected to a source that produces nothing.

### Measured on two real projects

| Lane | EvalRuns with reasoning |
|---|---|
| V1 (`SIMPLE_CHAIN_OF_THOUGHT`) | **19,673 / 19,673** carry `chain_of_thought` |
| V2 (`SIMPLE`) | **108 / 109 empty**; the 1 exception is a `deepseek_4_pro` config whose native `reasoning_content` was captured |

### Compounding: a dead field

`materialize_llm_judge_properties` — the shared create+test path — **always** sets `thinking_instruction` to `"Think step by step, explaining your reasoning."`. Nothing in the V2 path ever read it (`grep thinking_instruction libs/core/kiln_ai/adapters/eval/v2_eval_llm_judge.py` → no hits). It was stored, surfaced in the config, and silently ignored.

## The fix

Keeps the single `SIMPLE` call — no CoT turn, nothing added to the deprecated chain-of-thought path.

- `build_score_schema()` takes an optional `reasoning_instruction`. When set, it **prepends** a `reasoning` string field, so the judge states its rationale *before* committing to scores rather than rationalising after the fact.
- The V2 judge passes `thinking_instruction` as that instruction — giving the dead field a real job — then pops `reasoning` out of the output before scoring and files it under `intermediate_outputs["reasoning"]`, the key the web UI already reads. **No frontend change needed.**
- Native provider reasoning is preserved: if it already occupies `reasoning`, the judge's stated rationale lands in `judge_rationale` rather than clobbering it.
- **Skipped in `g_eval` mode.** That path rebuilds raw JSON from logprobs and `metric_offsets` requires each metric name to appear *exactly once*; a free-text rationale that quoted a score name would turn a missing rationale into a hard scoring failure. g_eval keeps native provider reasoning only.

Scores are unaffected — `build_llm_as_judge_score` and `build_g_eval_score` both derive their metric list from `output.keys()`, and `reasoning` is removed before either runs.

## Why the existing test didn't catch it

`test_intermediate_outputs_propagated` mocks the adapter's return value, so it proves the plumbing while the source stays empty. The new tests assert the **schema the judge is actually handed** and that the rationale never leaks into scores.

## Tests

- 5 new tests: reasoning field leads the schema and carries the instruction; rationale persisted and kept out of scores; no instruction → no field; native reasoning not clobbered; g_eval schema stays clean; blank rationale not persisted.
- `libs/core/kiln_ai/adapters/` + `libs/core/kiln_ai/datamodel/`: 4005 passed.
- `app/desktop/studio_server/test_eval_api.py`: 120 passed.
- `ruff format` + `ruff check`: clean.
- One unrelated pre-existing failure (`test_create_document_content_type_detection[document.md]`) reproduces on the base branch with these changes stashed. Web checks fail only for missing `node_modules` in a fresh worktree; no frontend files touched.

## Base branch

Targets `dchiang/multiturn-eval-megabranch`: `v2_eval_llm_judge.py` does not exist on `main`, so this cannot be based there. Present and identical on `agi-anyting_goes_into`, which will need the same fix or a merge.

## Note for existing data

This is capture-forward only. V2 judge EvalRuns already on disk have no rationale and would need re-scoring to gain one.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/base_eval.py
+++ b/libs/core/kiln_ai/adapters/eval/base_eval.py
@@ -27,6 +27,16 @@ from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 DEFAULT_SYSTEM_PROMPT = "You are an evaluator."
 _DEFAULT_THINKING_INSTRUCTION = "Think step by step, explaining your reasoning."
 
+# Schema key the LLM judge writes its rationale into. Lifted out of the judge's
+# structured output into EvalRun.intermediate_outputs, so it must never collide
+# with an output score's json_key. Matches the key the web UI's "View reasoning"
+# control reads.
+REASONING_SCHEMA_KEY = "reasoning"
+
+# Fallback slot used only when a reasoning-capable provider already deposited
+# native reasoning under REASONING_SCHEMA_KEY, so neither record is lost.
+JUDGE_RATIONALE_KEY = "judge_rationale"
+
 _JINJA_OPENERS = ("{{", "{%", "{#")
 
 
@@ -359,7 +369,12 @@ class BaseEval:
         pass
 
     @classmethod
-    def build_score_schema(cls, eval: Eval, allow_float_scores: bool = False) -> str:
+    def build_score_schema(
+        cls,
+        eval: Eval,
+        allow_float_scores: bool = False,
+        reasoning_instruction: str | None = None,
+    ) -> str:
         """
         Build a JSON schema for the scoring output of the task requirements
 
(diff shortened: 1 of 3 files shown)


tawnymanticore:
Superseded by #1613, which targets `scosman/evals_v2` — the right home for a V2-evals fix. Same commit, cherry-picked cleanly, full test run re-done on that base. The megabranch and `agi-anyting_goes_into` carry the identical bug and will pick it up on their next merge from evals_v2.

PR #1622, opened 2026-07-23
Add per-score direction flag to eval output scores

scosman:
## What does this PR do?

Eval output scores had no way to declare which way is better. This adds `direction` on `EvalOutputScore` — `higher_is_better` (default), `lower_is_better`, or `informational` (context only, should never drive decisions) — so API consumers comparing results across runs can interpret score deltas without guessing.

- Rating scales (`five_star`, `pass_fail`, `pass_fail_critical`) are higher-is-better by definition: they accept `higher_is_better` and `informational`, and reject `lower_is_better`. `lower_is_better` is reserved for custom scores (unbounded numeric metrics), which evaluators do not support yet.
- Backward compatible: the field defaults to `higher_is_better`, so existing evals load unchanged; requests from clients that predate the field get the default.
- The field flows through the create_evaluator and read APIs automatically via `EvalOutputScore`; web client schema regenerated.

## Related Issues

Companion to #1620 and #1621 (eval-API improvements).

## Checklists

https://claude.ai/code/session_0135zWTBh8MRPwCXiWvuE597

---
_Generated by [Claude Code](https://claude.ai/code/session_0135zWTBh8MRPwCXiWvuE597)_

### DIFF
--- a/libs/core/kiln_ai/datamodel/eval.py
+++ b/libs/core/kiln_ai/datamodel/eval.py
@@ -55,6 +55,21 @@ class EvalConfigType(str, Enum):
     llm_as_judge = "llm_as_judge"
 
 
+class ScoreDirection(str, Enum):
+    """
+    The direction of improvement for an eval output score.
+
+    Tells consumers how to interpret a change in the score's value: 'higher_is_better'
+    means an increase is an improvement, 'lower_is_better' means a decrease is an
+    improvement, and 'informational' scores carry context only and should never drive
+    decisions in either direction.
+    """
+
+    higher_is_better = "higher_is_better"
+    lower_is_better = "lower_is_better"
+    informational = "informational"
+
+
 class EvalOutputScore(BaseModel):
     """
     A definition of a score that an evaluator will produce.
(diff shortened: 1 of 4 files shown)


libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -72,6 +87,10 @@ class EvalOutputScore(BaseModel):
     type: TaskOutputRatingType = Field(
         description="The type of rating to use ('five_star', 'pass_fail', 'pass_fail_critical').",
     )
+    direction: ScoreDirection = Field(
```
chiang-daniel: Nothing reads direction yet — score summaries and the run config comparison still treat all scores alike, so "informational should never drive decisions" is contract-by-docstring until a consumer lands. When #1620/#1621 pick this up, should the comparison's "best" math exclude informational scores?

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -89,6 +108,25 @@ def validate_type(self) -> Self:
             )
         return self
 
+    @model_validator(mode="after")
+    def validate_direction(self) -> Self:
+        match self.type:
+            case (
```
chiang-daniel, line 114: This validator allows a score to be both pass_fail_critical and informational at the same time. Those two make opposite promises: critical means "this failure should dominate the outcome", informational means "never use this score to decide anything". A score carrying both is a contradiction, and any consumer reading it has no idea which promise to believe.

Suggest rejecting the combination in this PR rather than later. If it's forbidden now and a real use ever appears, removing the restriction is free and breaks nothing. If it ships allowed and turns out to be a mistake, it can't be tightened later without breaking loads of any file that used the combination in the meantime.

PR #1628, opened 2026-07-28
Compare radar chart: relative/full scale toggle, removable usage axes, no false zeros, rename axes

tawnymanticore:
## What does this PR do?

Makes the compare screen's radar chart readable with a small number of run configs. Before: every axis was scaled to the max across the selected configs, so a single run config sat on the outer ring of every axis and the chart said nothing — and a score it had no result for was plotted at 0, indistinguishable on a radar from scoring 0.

- **Relative / Full Scale toggle.** Full Scale scales each axis to the score type's own range instead — 1 for pass/fail, 5 for 5-star — from the eval definitions the page already caches. Scores with no known range keep the relative max, so nothing is clipped. Defaults to Full Scale when only one run config has results.
- **Missing results are no longer drawn as zero.** echarts maps a missing radar value to the center of the chart by design (`radarLayout`'s `getValueMissingPoint`), so a gap can't be drawn. The chart now plots only the scores every selected run config has a result for, and reports the count of the rest under the title.
- **Cost, latency and token axes.** All five usage metrics are axes, scored relative to the selected configs (bigger means less cost, less time, fewer tokens), so they stay relative in Full Scale mode too. Each is hidden with the same `✕` the eval sections already have, on its row in the cost table.
- **Hide or rename any row.** Every row and section has a `✕` - eval score rows and the usage section itself included - and every row has a `✎` to rename it for display. Several evals on a task tend to share score names, so three "Overall Correct" axes were indistinguishable on the radar; a renamed row is renamed in the table, on its axis, in the tooltip and in the "Hidden (n)" dropdown, which lists hidden rows with their eval beneath. Both are display-only, tracked in the URL as `hidden_metrics` and `metric_labels`, and never touch the eval or its results.
- **Layout.** Larger plot, legend below the chart for one or two configs, area fill for a single series, tooltip confined to the chart and trimmed to the ten weakest scores.
- **One page-level fix.** `updateURL()` now passes `noScroll`/`keepFocus` to `goto()`. Every state change on the compare page used to jump back to the top and drop focus. Not specific to the chart.

Not included: skipping scores whose direction is `lower_is_better` or `informational`, since a radar reads "further out is better". That needs `EvalOutputScore.direction`, which isn't on main yet.

Screenshot 2026-07-28 at 12 00 00 PM
Above: Note key things here. there is an X beside the usage/latency metrics. There is a full scale vs relative button. and things that are empty are omitted from the plot

## Related Issues

None — this wasn't tracked in an issue.

## Checklists

Web UI only, so the `/lib` item doesn't apply. From `app/web_ui`: `npm run check` (0 errors), `npm run lint`, `npm run format_check` and `npx vitest run` (2478 tests, including new ones for the hide/rename helpers in `compare_view.ts`) all pass.


app/web_ui/src/lib/components/compare_radar_chart.svelte
```diff
@@ -400,18 +639,53 @@
 </script>
 
 <!-- Radar charts don't really work with <3 items -->
-{#if dataKeys.length >= 3}
+{#if dataKeys.length >= MIN_RADAR_AXES}
```
tawnymanticore: This gate changed meaning when `dataKeys` stopped including `COST_KEY`/`LATENCY_KEY`. On main it read "at least one eval score"; now it reads "at least three eval scores", while the chart actually draws `dataKeys + visibleUsageKeys`. A task with one or two eval scores now hides the whole card even though six or seven axes would be drawn. `generateChartData()` already applies `MIN_RADAR_AXES` to the real axis count, so this outer gate could be `> 0` (or go away).

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -348,28 +373,42 @@
...
+      .filter(
+        (section) =>
+          section.eval_id === "kiln_cost_section" ||
+          !hidden.includes(section.eval_id),
+      )
+      .map((section) =>
+        section.eval_id === "kiln_cost_section" && hiddenUsage.length > 0
+          ? {
+              ...section,
+              items: section.items.filter(
```
Comment (tawnymanticore) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte:395]: Hiding all five cost rows leaves this section with `items: []`. The table's `{#if section.items.length == 0}` branch then falls through to `Unknown issue - no scores found`, because the cost section's `has_default_eval_config` is `undefined` rather than `false`. Easy to hit now that each row has its own `✕`.

app/web_ui/src/lib/components/compare_radar_chart.svelte
```diff
@@ -169,120 +280,219 @@
...
+    // entirely rather than emptying every axis.
+    const plottedConfigs = selectedRunConfigIds
+      .map((configId) => run_configs.find((c) => c.id === configId))
+      .filter((config): config is TaskRunConfig => !!config)
+      .filter((config) =>
+        dataKeys.some(
+          (key) => getModelValueRaw(config.id ?? null, key) !== null,
+        ),
+      )
+    const plottedConfigCount = plottedConfigs.length
```
tawnymanticore: This shadows the module-level `$: plottedConfigCount`, and nothing reads the returned copy — `chartSummary` destructures it and then drops it. Same for `usageValues` in the return: no caller destructures it. Both could come off the return type.

app/web_ui/src/lib/components/compare_radar_chart.svelte
```diff
@@ -169,120 +280,219 @@
...
   }
 
-  // Check if there's data to display (reactive, depends on dataKeys and selectedRunConfigIds)
-  $: hasData = (() => {
-    if (!dataKeys || dataKeys.length === 0 || !selectedRunConfigIds) {
-      return false
+  // Check if there's data to display (reactive - references every input that can
+  // change what generateChartData() returns)
+  $: chartSummary = (() => {
+    // visibleUsageKeys is referenced so this re-runs when the usage chips change
```
tawnymanticore: "usage chips" is stale — the chips became the `✕` on the cost rows. Worth renaming so the next reader doesn't go looking for a control that isn't there.

tawnymanticore (commented):
A few things I hit reading this through.

app/web_ui/src/lib/components/compare_radar_chart.svelte
tawnymanticore: Fixed in `0de4b1a0e`: the gate is now `dataKeys.length + visibleUsageKeys.length >= MIN_RADAR_AXES`, with a comment on why the usage axes count.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte
Comment (tawnymanticore) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte:395]: Fixed in `0de4b1a0e`: the empty-section branch is now skipped for `kiln_cost_section`, so an emptied cost section renders as just its header.

app/web_ui/src/lib/components/compare_radar_chart.svelte
tawnymanticore: Fixed in `5fdb03b71`: `plottedConfigCount` and `usageValues` are off the return type, the shadowing local is gone, `chartSummary` destructures only what it uses, and axis maxima now iterate `plottedConfigs` to match the rest of the function.

app/web_ui/src/lib/components/compare_radar_chart.svelte
tawnymanticore: Fixed in `0de4b1a0e`: the comment now reads "a usage row is hidden".

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -348,28 +373,42 @@
...
   $: visibleComparisonFeatures = filterVisibleFeatures(
     comparisonFeatures,
     hiddenEvalIds,
+    hiddenUsageKeys,
   )
   $: visibleChartComparisonFeatures = filterVisibleFeatures(
     chartComparisonFeatures,
     hiddenEvalIds,
+    hiddenUsageKeys,
   )
```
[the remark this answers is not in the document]
Comment (tawnymanticore) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte:412]: Fixed in `0de4b1a0e`, same conclusion you reached: the empty-section branch is now `section.items.length == 0 && section.eval_id !== "kiln_cost_section"`, so hiding every cost row leaves the section header alone instead of the error text. Verified in the browser.

scosman:
We need cost and speed on the chart. It's essential to every pitch/demo we have done. Up for improving it, but the values have meaning even if relative

tawnymanticore:
> We need cost and speed on the chart. It's essential to every pitch/demo we have done. Up for improving it, but the values have meaning even if relative

@scosman Right I agree, this PR just unifies all eval rows to make any X'able in the Compare Run Config screen (today all are X'able except for speed/tokens/latency)... the user may want to make the radar chart demo-able and may need to modify any of it. Also if there is just 1 run config shown on the radar chart, then speed/cost/tokens don't make any sense on the Radar Chart because it is hard coded to be at 50 out of 100, its a relative measure so requires >1 run config to make any sense since we don't have an upper bound on Low Cost or High Cost, or Low Latency or High Latency etc unlike the rest of the radar chart axes which are bounded by the eval scores (fail --> pass, 0-5 stars)

tawnymanticore:
Closing this PR in favour of #1765. The new PR carries the same feature on top of the current `main`, moves the chart maths into tested modules, and fixes the items from this thread: the tooltip now escapes score labels, and two run configs that share a name no longer collapse into one series. The commit history of this branch stays available here for reference.

---

PR #1629, opened 2026-07-28
evals: persist judge model usage on the legacy G-Eval path

tawnymanticore:
> **Rescoped.** `scosman/evals_v2` has since grown judge-usage capture for V2 evals on its own. This branch is now only the one lane the base does not cover: the legacy G-Eval / LLM-as-judge path. 12 files → 6, 308 insertions → 127.

## What the base already does

Everything this PR originally added for V2 evals is on `scosman/evals_v2` today, and is dropped here in favour of the base's version:

- `EvalRun.eval_usage` and `V2EvalResult.usage` in the datamodel.
- `LlmJudgeEval.evaluate` keeping the judge TaskRun's usage.
- `_persist_score` / `_persist_judgment` writing it on every V2 record.
- A datamodel default-None + round-trip test (`test_eval_usage_defaults_to_none_and_round_trips`), so this branch's duplicate is gone.

## What is left

`EvalRunner.run_job` dispatches on `eval_config.config_type`: `v2` goes to `_run_v2_job`, and `g_eval` / `llm_as_judge` go to `_run_legacy_job`. The legacy path still throws the judge run away, so no `g_eval`-configured eval on disk records what its own scoring cost.

- **`GEval.run_eval`** keeps the judge TaskRun instead of dropping it (`_, run_output = ...`) and returns its usage. The value covers the **full judgment** — the adapter sums usage across every model call in the invocation, so the two-call COT heuristic is aggregated, not just the last call.
- **`BaseEval.run_eval` / `run_task_and_eval`** carry it through as an extra tuple element; `BaseV2EvalBridge` forwards the base's `V2EvalResult.usage`.
- **`_run_legacy_job`** sets `eval_usage` on the EvalRun it saves.

Non-LLM evals persist `None` — honestly unset rather than zero.

## Why not just close it

The legacy path is live, not sunset: `legacy_eval_adapter_from_type` still returns `GEval` for both `g_eval` and `llm_as_judge`, and any eval config already on disk with those types still runs through it. Until that dispatch arm is deleted, this is the last hole in per-lane usage accounting. If legacy evals are being removed outright, closing this is a one-click alternative — nothing else in the branch survives on its own.

## Back-compat

Unchanged from the original: `eval_usage` is an optional field defaulting to `None`, already on the base. The only behaviour change is that legacy-judge records written after this carry a value where they previously carried `None`.

## Testing

- `libs/core/kiln_ai/adapters/eval/` + `test_eval_model.py` + `test_eval_api.py`: **1090 passed**.
- New: `test_run_job_persists_judge_eval_usage` — a judged legacy EvalRun carries the judge's usage, through `run_job` and a disk round-trip.
- Updated: evaluator mocks and stubs widened to the new tuple arity, including the base's own `test_run_job_full_trace_serializes_per_message_usage`.
- Full `libs/` + `app/desktop` run: **7507 passed**. The 5 failures and 5 errors (vector-store, chunker, model-cache benchmark, document-api, and tkinter-less desktop imports) all reproduce on `origin/scosman/evals_v2` with this branch's changes absent.
- `ruff format --check` clean on every touched file.


tawnymanticore:
Closing — legacy evals are being sunset, and the legacy G-Eval path was all this branch had left.

For the record of what was verified before closing: everything else this PR originally added has independently landed on `scosman/evals_v2` — `EvalRun.eval_usage`, `V2EvalResult.usage`, the judge-usage capture in `LlmJudgeEval.evaluate`, and the wiring through `_persist_score` / `_persist_judgment`. Judge usage on the V2 lane is fully covered there, including a datamodel round-trip test. Nothing is lost by closing this.

The only thing that goes with it is `_run_legacy_job` recording `eval_usage` for `g_eval` / `llm_as_judge` configs. Those still run today (`legacy_eval_adapter_from_type` returns `GEval` for both, and `run_job` routes to `_run_legacy_job` when `config_type != v2`), so until that dispatch arm is deleted, legacy-judge EvalRuns will keep persisting `eval_usage=None`. That is the accepted cost of the sunset — the field is optional and `None` already means "not measured".

If legacy evals turn out to have a longer tail than expected, the change is ~15 lines of production code: keep the judge TaskRun in `GEval.run_eval` instead of discarding it, widen the `BaseEval.run_eval` / `run_task_and_eval` tuples, and set `eval_usage=` on the EvalRun in `_run_legacy_job`. The branch `fix/evalrun-judge-usage` is left in place at `cb19043c` (all CI green) if it's ever wanted back.

---

PR #1636, opened 2026-07-31
Eval score summaries: report percentiles, not just the mean

tawnymanticore:
## What

Adds six optional distribution fields to `ScoreSummary` in `app/desktop/studio_server/eval_api.py`:

`min_score`, `p25_score`, `median_score`, `p75_score`, `p90_score`, `max_score` — alongside the existing `mean_score`.

Both aggregation sites now route through one shared helper, `score_summary_from_values(values, n_excluded)`:

- `compute_score_summary()`
- the run-config `eval_scores` endpoint

Both previously kept a running total; they now accumulate the individual scores into a list so the summary can report the distribution. Skipped runs are excluded exactly as they already were for the mean, and the `count > 0 or n_excluded > 0` emit condition is unchanged.

## Why

Eval score aggregation currently reports only a mean. For numeric (custom-type) scores — tool-call counts, tokens per call, per-turn latency — the distribution is heavily right-skewed, and the mean badly misrepresents it.

Measured on a 140-conversation sweep: on one lane the mean and median diverged by **1.70×** — **224.5s mean vs 132.2s median** — driven by a single ~30-minute outlier conversation. Reading the mean alone, that lane looks uniformly slow; it is actually a well-behaved lane with one pathological run. The mean is the wrong summary for deciding whether a regression is broad or a tail event, and the tail is what dominates cost and risk.

Sharing one helper between the two sites is deliberate: it makes it impossible for them to drift onto different percentile definitions.

## Percentile definition

Reuses the existing `kiln_server.statistics_lib.percentile` — **linear interpolation between the two nearest order statistics**, matching the `numpy.percentile` / `statistics.quantiles(method="inclusive")` default. Chosen because it is the definition already in the codebase and the one most readers assume, so an even-length list's median is the average of the two middle values rather than the lower one, and p90 of a short list is interpolated rather than snapped to an existing datum. No new dependency.

Empty input yields `None` for every statistic, never `0.0` — a `0` would flow into downstream aggregation as if it were a real datum. The percentiles follow `mean_score`'s existing convention here exactly.

## Merge with the run/score split

`scosman/evals_v2` merged in. Production code and `api_schema.d.ts` merged cleanly; only `test_eval_api.py` conflicted, in two places:

- **Import list** — union of the base's new exports (`scored_trace_usage_for_run_config`, `split_size`, `summary_eval_config`) with this branch's `score_summary_from_values`.
- **`test_score_summary_percentiles`** — `compute_score_summary` now takes a `ResolvedSplit` rather than a bare id set, so the split is built with the base's `stub_split` helper. The item set (four scored + one skipped) and every distribution assertion are unchanged.

Both aggregation sites still route through the shared helper afterwards, including the `eval_scores` endpoint, which the base reworked around `eval_run_item_key` / `ResolvedSplit` but left on the same score-accumulation shape.

## Back-compat

Strictly additive:

- `mean_score`, `n_used`, `n_excluded` are unchanged in name and semantics.
- Every new field is optional with a `None` default, so older stored payloads and existing API/TypeScript consumers keep working untouched.
- `api_schema.d.ts` was regenerated with the sanctioned `app/web_ui/src/lib/generate_schema.sh` (never hand-edited) — a 30-line additive diff, all fields emitted as optional (`?`). Re-verified after the merge: the regenerated file byte-matches the checked-in one on every `ScoreSummary` field.

## Tests

- `libs/server/kiln_server/test_statistics_lib.py` — new `TestPercentile` covering the previously untested `percentile()`: empty, single value, odd/even length, unsorted input, all-equal, out-of-range raises, and a parametrized case pinned to numpy's default gold values.
- `app/desktop/studio_server/test_eval_api.py` — `test_score_summary_percentiles` at `compute_score_summary`; all-skipped assertions confirming percentiles are `None` rather than `0.0` at both the helper and the endpoint; and endpoint-level distribution assertions over three scores (3.5 / 4.0 / 4.5), which exercise both the exact-index and interpolation branches and pin a case the mean alone cannot distinguish from any other set summing to 12.0.

Post-merge: `test_eval_api.py` + `test_statistics_lib.py` — **224 passed**. `ruff format --check` clean on both touched files.


tawnymanticore:
Merged `scosman/evals_v2` in — conflicts resolved, PR is mergeable again.

Production code (`eval_api.py`) and `api_schema.d.ts` merged cleanly; the run/score split only collided in `test_eval_api.py`:

- **Import list** — union of the base's new exports (`scored_trace_usage_for_run_config`, `split_size`, `summary_eval_config`) and this branch's `score_summary_from_values`.
- **`test_score_summary_percentiles`** — `compute_score_summary` now takes a `ResolvedSplit` instead of a bare id set, so the split is built with the base's `stub_split` helper. The item set (four scored + one skipped) and every distribution assertion are unchanged.

Both aggregation sites still route through `score_summary_from_values` after the merge — including the run-config `eval_scores` endpoint, which the base reworked around `eval_run_item_key` / `ResolvedSplit` but left on the same score-accumulation shape.

`test_eval_api.py` + `test_statistics_lib.py`: **224 passed**. `ruff format --check` clean on both touched files. The regenerated `api_schema.d.ts` byte-matches the checked-in one on every `ScoreSummary` field.

---

tawnymanticore:
**Retargeted to `dchiang/eb-v2-merge`** (was `scosman/evals_v2`).

Clean merge, same shape as #1591's retarget: the branch already carried evals_v2 at `4ee3cb97`, which *is* the merge-base with eb-v2-merge, so this brings eb-v2's 197 commits on top rather than reconciling two lineages.

**One conflict**, an import-list collision in `test_eval_api.py` — eb-v2 added `scored_trace_usage` where this branch added `score_summary_from_values`. Union of the two.

The thing worth checking on a base that reworked `eval_api` this heavily is whether a third aggregation site appeared that would bypass the shared percentile definition. It didn't: both sites still route through `score_summary_from_values` (`compute_score_summary` and the run-config `eval_scores` endpoint), and the only `ScoreSummary(...)` constructions left in the file are the two inside the helper itself.

`api_schema.d.ts` regenerated against the new base — the same 30-line additive diff, every field optional. No new endpoints, so the agent-policy annotations are byte-identical to the base's.

**Verification on the new base**: `test_eval_api.py` + `test_statistics_lib.py` **240 passed**; full `libs/` + `app/desktop` **8146 passed**. The diff is unchanged in shape from where it was on evals_v2 — 4 files, 290 insertions. `ruff format --check` clean on all three touched files.

⚠️ Heads up: `dchiang/eb-v2-merge` is currently **red in its own CI** — its last three runs all fail `TestValidateMultiTurnDriveReadiness::test_two_message_chain_of_thought_run_config_rejected` with `DID NOT RAISE`. That test arrives with the merge, so this PR will now inherit the failure where it was previously green on evals_v2. Nothing this PR touches — the assertion resolves `gpt_4o`'s reasoning capability through the remote model list, and `_splits_a_turn_into_two_messages` swallows any lookup failure into `False`, which produces exactly that symptom. Same inherited failure as #1591 and #1645.

---

chiang-daniel (approved):

PR #1645, opened 2026-08-04
evals: record the synthetic user's tokens, not just its cost

tawnymanticore:
> **Retargeted to `dchiang/eb-v2-merge` and rescoped.** Previously based on `dchiang/multiturn-eval-megabranch` at `dccd792a`. That base and this one are 300+ commits apart with a mid-July merge base, so a merge would have been meaningless; the branch was rebuilt as one commit on `eb-v2-merge`. 12 files → 9, and the datamodel change is gone entirely.

## What the base already does

`dchiang/eb-v2-merge` is the branch where the eval-run/score split and the synthetic-user lane both exist, so the review feedback — *"the persisted field must move onto the stored task run"* — is already satisfied there:

- `TaskRun.synthetic_user_usage: Usage | None` exists on the datamodel.
- The driven conversation is persisted as a TaskRun and the EvalRun points at it via `scored_run_id`, so SU spend is booked once per trace rather than once per judge.
- `usage` on that TaskRun is honestly assistant-only (`_conversation_usage`), with the SU on its own field.

None of that needs changing, so all of it is dropped from this branch.

## What's left, and why it matters

The base fills that field with a cost-only stub:

```python
synthetic_user_usage=Usage(cost=drive_result.su_total_cost)
if drive_result.su_total_cost > 0
else None,
```

Every token count the driver model reported is discarded on the way in. That's the same gap the original PR was about, one layer down:

- The SU is normally a **different model on a different provider** from the agent under test (on the engagement that found this: agent `gpt_5_4@openrouter`, SU `glm_5_2@fireworks_ai`). A cost with no tokens reconciles against **neither** invoice and cannot be split per model at all.
- It can't be recomputed at a different price, so it's worthless for "what would this sweep cost on a cheaper driver".
- **SU turns are never persisted as TaskRuns.** Whatever the drive loop drops exists nowhere on disk afterwards — there is no second chance to recover it.

### Scale of what's unreadable

Reconstructed from 850 stored drives on the test model engagement (traces replayed through the exact filtering `respond()` applies): SU calls run at exactly `turns` per conversation, mean **3,548 input tokens/conversation** (p90 6,556, max 25,312) — **~3.0M input tokens over 2,550 driver calls**. And a reconstruction is a floor, not a substitute: the SU is often a reasoning model (reasoning bills as output, invisible offline) and its calls share a growing prefix, so cache behaviour moves the invoice and can't be inferred.

## The change

- **`SyntheticUserDriver.respond` returns `Usage | None`** instead of a float cost. `None` when the provider reported nothing — a zeroed `Usage` would read as a genuinely free call rather than an unmeasured one.
- **`DriveCaseResult.su_usage`** replaces the `su_total_cost` field, summed across the case's turns with `Usage.__add__`. That addition is None-graceful per field, so a turn that reports only cost can't wipe out another turn's token counts — there's a test pinning exactly that.
- **`su_total_cost` survives as a derived property**, so the interactive runner's `CaseCompletedEvent.total_cost` is unchanged in behaviour: it wants a float it can add, and "no usage reported" has always summed as zero there.
- **The eval runner stores the whole `Usage`** on the trace TaskRun.

The per-turn `on_turn(su_cost=...)` hook contract is untouched, so the batch runner's event shapes and per-attempt spend accounting are unaffected.

## Back-compat

`TaskRun.synthetic_user_usage` is unchanged in name, type and default — this only changes how much of it gets filled in. Records written before this carry cost-only values that still load and read correctly; `None` still means "not measured". No datamodel change, so no schema regeneration.

The one breaking change is internal: `DriveCaseResult(su_total_cost=...)` no longer constructs, since that name is now a property. That's deliberate — every construction site should be handing over the full `Usage`.

## Drive fingerprints

Untouched. `compute_drive_fingerprint` hashes drive config + run config properties + scenario content; none of those change here, so no stored conversation is invalidated and no re-drive is triggered.

## Testing

- New: SU usage summed across turns; `None` when no turn reports; token counts preserved when one turn reports only cost; `respond()` returns the full `Usage`; the persisted trace carries the driver's tokens alongside the agent's.
- Updated: existing SU mocks widened from `(msg, float)` to `(msg, Usage | None)`.
- `libs/core/kiln_ai/adapters/eval/` + `synthetic_user/`: **813 passed**.
- Full `libs/` + `app/desktop`: **8127 passed**. The 5 failures and 5 errors (vector-store, chunker, model-cache benchmark, document-api, and tkinter-less desktop imports) all reproduce on `origin/dchiang/eb-v2-merge` with this branch's changes absent.
- `ruff format --check` clean on all touched files.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_runner.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_runner.py
@@ -1048,10 +1048,12 @@ class EvalRunner:
             trace=leaf.trace,
             usage=_conversation_usage(drive_result.chain),
             # The synthetic-user driver's spend rides its own field so `usage`
-            # stays honestly assistant-only; a zero-cost drive records None.
-            synthetic_user_usage=Usage(cost=drive_result.su_total_cost)
-            if drive_result.su_total_cost > 0
-            else None,
+            # stays honestly assistant-only. The whole Usage, not a cost-only
+            # stub: the SU is usually a different model on a different provider
+            # from the agent, so its tokens are the half that makes the figure
+            # reconcilable. `drive_case` already returns None for a drive whose
+            # provider reported nothing.
+            synthetic_user_usage=drive_result.su_usage,
             cumulative_usage=leaf.cumulative_usage
             or MessageUsage.from_trace(leaf.trace),
             eval_source=EvalItemSource(source_type=source_type, source_id=source_id),
--- a/libs/core/kiln_ai/synthetic_user/test_runner.py
+++ b/libs/core/kiln_ai/synthetic_user/test_runner.py
@@ -21,6 +21,7 @@ from kiln_ai.datamodel.datamodel_enums import (
 from kiln_ai.datamodel.run_config import KilnAgentRunConfigProperties, ToolsRunConfig
 from kiln_ai.datamodel.task import Task
 from kiln_ai.datamodel.task_run import TaskRun
+from kiln_ai.datamodel.usage import Usage
 from kiln_ai.synthetic_user import runner as runner_mod
 from kiln_ai.synthetic_user.case import SyntheticUserCase
 from kiln_ai.synthetic_user.driver import SyntheticUserDriver
@@ -122,7 +123,7 @@ def _patch_su_driver(
         instance = Mock(spec=SyntheticUserDriver)
         # respond() returns (message, cost). Tests that don't care about
         # cost get 0.0 — the runner adds it to total_cost regardless.
-        instance.respond = AsyncMock(side_effect=[(r, 0.0) for r in replies])
+        instance.respond = AsyncMock(side_effect=[(r, None) for r in replies])
         return instance
 
     monkeypatch.setattr(runner_mod, "SyntheticUserDriver", _ctor)
@@ -254,7 +255,7 @@ async def test_total_cost_sums_target_and_su_driver_spend(
     # at $0.01 → $0.01 SU per case.
     def _ctor(info, config):
         instance = Mock(spec=SyntheticUserDriver)
-        instance.respond = AsyncMock(side_effect=[("u2", 0.01)])
+        instance.respond = AsyncMock(side_effect=[("u2", Usage(cost=0.01))])
         return instance
 
     monkeypatch.setattr(runner_mod, "SyntheticUserDriver", _ctor)
@@ -450,7 +451,7 @@ async def test_malformed_blob_surfaces_as_case_failed(
 
     def _ctor(info, config):
         instance = Mock(spec=SyntheticUserDriver)
-        instance.respond = AsyncMock(return_value=("ok", 0.0))
+        instance.respond = AsyncMock(return_value=("ok", None))
         return instance
 
     _patch_su_driver_factory(monkeypatch, _ctor)
@@ -574,7 +575,7 @@ async def test_su_failure_deletes_chain_including_just_persisted_run(
         instance = Mock(spec=SyntheticUserDriver)
         # Turn 1's SU reply succeeds; turn 2's SU call dies mid-case.
         instance.respond = AsyncMock(
-            side_effect=[("u2", 0.0), ValueError("su blew up")]
+            side_effect=[("u2", None), ValueError("su blew up")]
         )
         return instance
 
@@ -774,7 +775,7 @@ async def test_retried_case_batch_total_includes_both_attempts_costs(
     # One SU call per attempt (turns=2), at $0.01.
     def _ctor(info, config):
         instance = Mock(spec=SyntheticUserDriver)
-        instance.respond = AsyncMock(side_effect=[("u2", 0.01)])
+        instance.respond = AsyncMock(side_effect=[("u2", Usage(cost=0.01))])
         return instance
 
     monkeypatch.setattr(runner_mod, "SyntheticUserDriver", _ctor)
@@ -824,7 +825,7 @@ async def test_dead_case_failed_event_reports_all_attempts_spend(
 
     def _ctor(info, config):
         instance = Mock(spec=SyntheticUserDriver)
-        instance.respond = AsyncMock(side_effect=[("u2", 0.01)])
+        instance.respond = AsyncMock(side_effect=[("u2", Usage(cost=0.01))])
         return instance
 
     monkeypatch.setattr(runner_mod, "SyntheticUserDriver", _ctor)
--- a/libs/core/kiln_ai/synthetic_user/test_runner.py
+++ b/libs/core/kiln_ai/synthetic_user/test_runner.py
@@ -121,8 +121,9 @@ def _patch_su_driver(
             else list(replies_per_case)
         )
         instance = Mock(spec=SyntheticUserDriver)
-        # respond() returns (message, cost). Tests that don't care about
-        # cost get 0.0 — the runner adds it to total_cost regardless.
+        # respond() returns (message, Usage | None). Tests that don't care about
+        # the driver's spend hand back None — the shape a provider that reported
+        # nothing produces, which the runner totals as zero.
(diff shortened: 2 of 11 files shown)


tawnymanticore:
**This cannot be retargeted to `scosman/evals_v2` yet — the lane it modifies does not exist there.** Leaving the base as `dchiang/multiturn-eval-megabranch` and writing down what I found.

## Why the retarget is blocked

`libs/core/kiln_ai/synthetic_user/` is 19 files on the megabranch and **0 files** on `scosman/evals_v2`. Eight of this PR's twelve changed files live in that tree. On evals_v2 there is no synthetic-user driver at all: `_is_multi_turn()` exists only to make `_run_v2_job` **skip** multi-turn items.

The two branches share an ancestor 322 commits back on evals_v2 and 359 back on the megabranch, so this is a real fork, not a stale rebase.

## On "the persisted field must move onto the stored task run"

Correct for evals_v2, and not expressible on this base:

- On **evals_v2**, `_persist_trace` saves the drive's TaskRun to disk and `EvalRun.scored_run_id` points at it; the inline `task_run_trace` / `task_run_usage` fields are deprecated. One trace is scored by N judges → N EvalRuns, so SU usage on `EvalRun` would be **counted N times**, and a reused trace would attribute spend to a record that made no call. It belongs on the trace TaskRun.
- On **this base**, the drive's leaf TaskRun is never persisted — the trace is serialized inline onto `EvalRun.task_run_trace`. There is no stored task run to move the field onto, so `EvalRun.synthetic_user_usage` is the only place the value can go here.

So the requested shape and this PR's base are mutually exclusive. Retargeting now would mean deleting the drive-side capture too (it lives in `synthetic_user/`), which is the half the review called solid.

## Does this still matter?

Yes, and the gap is unchanged: an eval sweep spends on three models and the synthetic user is the one whose spend no invoice can check. The reconstruction in the description stands — ~3.0M input tokens over 2,550 driver calls that nothing accounted for.

Of the three lanes on evals_v2 today:

| lane | where usage lives on evals_v2 |
| --- | --- |
| agent under test | `TaskRun.usage` on the persisted trace ✅ |
| judge | `EvalRun.eval_usage` ✅ (V2); legacy G-Eval closed by #1629 |
| synthetic user | lane not present — multi-turn items are skipped |

So evals_v2 has no SU spend to lose *today*, and will have unmeasured SU spend the moment the multi-turn lane lands there unless this lands with it.

## Suggested disposition

Two coherent options, both Mike's call:

1. **Hold this PR against the megabranch** as the record of the design, and port it as part of whatever brings the multi-turn lane to evals_v2 — at which point `synthetic_user_usage` goes on the **trace TaskRun**, not on `EvalRun`, and `_agent_usage` stays as-is.
2. **Merge it here as-is** if the megabranch is going to ship on its own. It is `mergeable_state: clean` and self-consistent on this base; the field placement is then a known migration when the branches meet.

What I would *not* do is retarget and strip it — that discards the drive-side capture and leaves nothing behind.

---

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -1048,10 +1048,12 @@ async def _persist_driven_conversation(
...
-            # stays honestly assistant-only; a zero-cost drive records None.
-            synthetic_user_usage=Usage(cost=drive_result.su_total_cost)
-            if drive_result.su_total_cost > 0
-            else None,
+            # stays honestly assistant-only. The whole Usage, not a cost-only
+            # stub: the SU is usually a different model on a different provider
+            # from the agent, so its tokens are the half that makes the figure
+            # reconcilable. `drive_case` already returns None for a drive whose
+            # provider reported nothing.
+            synthetic_user_usage=drive_result.su_usage,
```
chiang-daniel, line 1056: Persisting the full `Usage` here is right, but it silently changes what the score-summary rollup reports. `scored_trace_usage` in `app/desktop/studio_server/eval_api.py` sums `synthetic_user_usage` into the per-trace figure null-tolerantly, and its docstring is explicit that it was written against a cost-only field: "Today that field carries cost only, so token counts pass through from the assistant side." Once this field carries tokens and latency, three things change in the run-config usage summary with no code in this diff touching them:

- Mean input/output/total tokens absorb the synthetic user's tokens (~3.5k input per conversation by this PR's own numbers), attributed to the agent under test.
- Reported latency stops meaning agent latency: the litellm adapter stamps `total_llm_latency_ms` on the usage `respond()` now returns, `Usage.__add__` sums latency across the drive's turns, and the summary blend sums it again into the agent's figure.
- New records diverge from migrated legacy records, which have cost blended into `usage` but agent-only tokens — the same summary field reads two different quantities depending on record age.

Cost blending is total-spend semantics and matches the legacy records. Token and latency blending conflates two models, which is the exact conflation this PR argues against. Suggest making `scored_trace_usage` blend cost only (e.g. add `Usage(cost=trace.synthetic_user_usage.cost)` rather than the whole object) and updating its docstring in this PR, since this PR is what changes the field's contents. If blending everything is the intent, the docstring and PR description should say so instead.

libs/core/kiln_ai/synthetic_user/test_runner.py
```diff
@@ -122,7 +123,7 @@ def _ctor(info, config):
         instance = Mock(spec=SyntheticUserDriver)
         # respond() returns (message, cost). Tests that don't care about
         # cost get 0.0 — the runner adds it to total_cost regardless.
-        instance.respond = AsyncMock(side_effect=[(r, 0.0) for r in replies])
+        instance.respond = AsyncMock(side_effect=[(r, None) for r in replies])
```
chiang-daniel, line 127: The comment above this line is now stale on both counts: `respond()` returns `(message, Usage | None)`, and these tests hand back `None` rather than `0.0`.

chiang-daniel (approved):

libs/core/kiln_ai/adapters/eval/eval_runner.py
tawnymanticore: Confirmed and fixed in `09751db7`. You're right on all three counts, and the magnitude is worse than it reads — measured on the exact shape this branch now writes (agent 100 in / 50 out / 4000ms, SU 3548 in / 61 out / 9000ms):

```
before  input_tokens=100   total_llm_latency_ms=4000   cost=1.25
after   input_tokens=3648  total_llm_latency_ms=13000  cost=1.25
```

Input tokens **36×**, latency **3.25×**, both attributed to the agent under test — the exact conflation `synthetic_user_usage` exists to undo, reappearing one layer up in the rollup. And silently: nothing in the diff touched `eval_api.py`, which is what made it worth catching at review rather than in a dashboard.

Taken your suggestion exactly — `scored_trace_usage` now adds `Usage(cost=trace.synthetic_user_usage.cost)` rather than the whole object, and the docstring says why per quantity:

- **Cost** blends: total-spend semantics, what the trace cost to produce with both models in, and what migrated legacy records already carry fused inside `usage`.
- **Tokens** don't: different model, different provider, so folding them in makes `cost / total_tokens` meaningless.
- **Latency** doesn't: this summary reports agent responsiveness, and driver wall clock would make every driven run config look slower than it is.

Your third point — that blending everything gives one field two meanings by record age — is the one I'd have been most likely to miss, and it's called out in the docstring now so the next person changing this field sees the constraint.

Two tests pin the boundary: cost blends while tokens and latency stay agent-only, and an SU record carrying tokens but no cost leaves the total untouched.

---

libs/core/kiln_ai/synthetic_user/test_runner.py
tawnymanticore: Fixed in `09751db7` — it now reads:

```python
# respond() returns (message, Usage | None). Tests that don't care about
# the driver's spend hand back None — the shape a provider that reported
# nothing produces, which the runner totals as zero.
```

Stale on both counts as you said: the return type and the `0.0`.

---

PR #1650, opened 2026-08-05
[eb-v2 review] Generate/SDG: load-failure handling and coordination notes

chiang-daniel:

Surface: the /generate landing route and its synth-data guidance model. One commit (landed as d8e558a77) — surface a task-load failure on /generate instead of spinning forever, and correct a stale comment about which legacy config types carry eval_steps. About 80% of the diff is tests. One **Personal review requested** comment flags the failed-key retry behavior. This PR also carries three coordination notes as thread comments: a sync map for the overlap with #1630, the synthetic-user event contract the SDG stream consumes, and a shared note on the score-key hazard.

---
*CI note: the "Check API Schema Bindings" failure here is an artifact of this review PR's pinned snapshot — it flagged an annotation file that has since been added on `dchiang/eb-v2-merge` (78081dd6d). No action needed from reviewers.*

### DIFF
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
@@ -11,12 +11,14 @@
   import { client } from "$lib/api_client"
   import { load_task } from "$lib/stores"
   import type { Task } from "$lib/types"
+  import { KilnError, createKilnError } from "$lib/utils/error_handlers"
   import Warning from "$lib/ui/warning.svelte"
 
   // watch out because query param value is not the same as gen_type
   type SynthReasonQueryParam = "eval" | "fine_tune" | "qna"
 
   let loading = true
+  let load_error: KilnError | null = null
   let task: Task | null = null
   $: is_multiturn = task?.turn_mode === "multiturn"
   $: project_id = $page.params.project_id!
@@ -74,10 +76,22 @@
 
   async function handle_routing(req_project_id: string, req_task_id: string) {
     loading = true
+    load_error = null
 
     // Synthetic data generation is single-turn only. For multi-turn tasks
     // don't redirect into the synth/qna flows — show the disabled notice.
-    const loaded_task = await load_task(req_project_id, req_task_id)
+    let loaded_task: Task | null
+    try {
+      loaded_task = await load_task(req_project_id, req_task_id)
+    } catch (e) {
+      if (req_project_id !== project_id || req_task_id !== task_id) return
+      // Show the error instead of spinning forever. The key stays marked as
+      // handled so we don't hot-loop retrying a failing load; navigating to a
+      // different key (or remounting the route) retries naturally.
+      load_error = createKilnError(e)
+      loading = false
+      return
+    }
     if (req_project_id !== project_id || req_task_id !== task_id) return
     task = loaded_task
     if (task?.turn_mode === "multiturn") {
@@ -203,6 +217,15 @@
     {#if loading}
       <div class="w-full min-h-[50vh] flex justify-center items-center">
         <div class="loading loading-spinner loading-lg"></div>
+      </div>
+    {:else if load_error}
+      <div
+        class="w-full min-h-[50vh] flex flex-col justify-center items-center gap-2"
+      >
+        <div class="font-medium">Error Loading Task</div>
+        <div class="text-error text-sm">
+          {load_error.getMessage()}
+        </div>
       </div>
     {:else if is_multiturn}
       <div class="flex flex-col items-center justify-center min-h-[60vh]">
--- a/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts
+++ b/app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts
@@ -203,8 +203,9 @@ export class SynthDataGuidanceDataModel {
   // are the eval's real definition. With no default judge, rebuild the steps the
   // same way the create-judge page would, from the eval's template.
   private judge_eval_steps(): string[] {
-    // Only legacy llm_as_judge configs carry eval_steps; the typed properties
-    // union doesn't declare the key, so read it untyped.
+    // Both legacy config types (llm_as_judge and g_eval) carry eval_steps in
+    // properties; the typed properties union doesn't declare the key, so read
+    // it untyped.
     const props = this.default_judge?.properties as
       | Record<string, unknown>
       | undefined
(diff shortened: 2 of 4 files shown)


app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -74,10 +76,22 @@
...
-    const loaded_task = await load_task(req_project_id, req_task_id)
+    let loaded_task: Task | null
+    try {
+      loaded_task = await load_task(req_project_id, req_task_id)
+    } catch (e) {
+      if (req_project_id !== project_id || req_task_id !== task_id) return
+      // Show the error instead of spinning forever. The key stays marked as
+      // handled so we don't hot-loop retrying a failing load; navigating to a
+      // different key (or remounting the route) retries naturally.
+      load_error = createKilnError(e)
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:91]: The /generate routing awaited `load_task` with no catch and no error state, so any task-fetch failure left the page on a permanent silent spinner — a regression against main. This catches the failure into `load_error` and renders an error branch. See the flag below on the retry behavior.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -74,10 +76,22 @@
...
 
     // Synthetic data generation is single-turn only. For multi-turn tasks
     // don't redirect into the synth/qna flows — show the disabled notice.
-    const loaded_task = await load_task(req_project_id, req_task_id)
+    let loaded_task: Task | null
+    try {
+      loaded_task = await load_task(req_project_id, req_task_id)
+    } catch (e) {
+      if (req_project_id !== project_id || req_task_id !== task_id) return
+      // Show the error instead of spinning forever. The key stays marked as
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:88]: **Personal review requested:** On a load failure the routing key stays marked as handled, so a failing load isn't retried in place — recovery is by navigating to a different key or remounting the route. I chose this because un-marking the key made the reactive routing hot-loop retry the failing load. Confirm you're happy with navigation-based recovery here, versus an explicit Retry button that clears `load_error` and re-runs the routing for the same key.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -204,6 +218,15 @@
       <div class="w-full min-h-[50vh] flex justify-center items-center">
         <div class="loading loading-spinner loading-lg"></div>
       </div>
+    {:else if load_error}
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:221]: The error branch itself, matching the load-failure states used on the other data pages: a titled message with the underlying error text, shown in place of the spinner once `load_error` is set. This is what the user now sees instead of an indefinite spinner when the task can't be loaded.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts
```diff
@@ -203,8 +203,9 @@ export class SynthDataGuidanceDataModel {
   // are the eval's real definition. With no default judge, rebuild the steps the
   // same way the create-judge page would, from the eval's template.
   private judge_eval_steps(): string[] {
-    // Only legacy llm_as_judge configs carry eval_steps; the typed properties
-    // union doesn't declare the key, so read it untyped.
+    // Both legacy config types (llm_as_judge and g_eval) carry eval_steps in
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_data_guidance_datamodel.ts:206]: The comment here claimed only llm_as_judge configs carry `eval_steps`, but g_eval configs carry it too — so a reader syncing against main could wrongly assume g_eval judges have no steps to rebuild and drop them. Corrected to name both legacy types; the untyped read is unchanged (the typed union still doesn't declare the key).

chiang-daniel:
**#1630 overlap — sync map and an offer**

Samantha — heads up on where #1630 touches the same files as the eval-builder fixes on this integration branch, so the merge is boring when it lands. None of these are asks; just a map.

When #1630 flows in:
- `deterministic_forms.test.ts`, `test_eval_model.py`, `test_eval_api.py`: our changes rewrote several describes to be behavior-named; resolve those to our describes and take your copy updates on top. No logic conflict, just naming.
- `datamodel/eval.py`: your Eval priority/status fields and our EvalInput tag validation + EvalRun docstring/properties-dispatch changes are in different regions — expect a clean merge; worth a glance that both validators still stand afterward.
- `eval_api.py`: your creation/display paths and our score-summary addition are different endpoints — clean merge expected.
- `api_schema.d.ts`: regenerate after the merge rather than hand-resolving it.

One concrete offer: we removed an orphaned `output_value_expression` binding in the judge builder (write-only, nothing read it). Your refactor relocates that binding into the new `JudgeConfigFields` component rather than dropping it. The cheapest fix is to just not carry it over in #1630 — happy to point at the exact removal if useful.

chiang-daniel:
**Synthetic-user event contract the SDG stream consumes**

Samantha — for the SDG stream, here's the current shape of the synthetic-user batch events so your consumer stays aligned (`kiln_ai/synthetic_user/runner.py`):

- `TurnCompletedEvent.turn_index` is 1-based within the CURRENT drive attempt — a retried case restarts at 1, so read this field rather than counting events.
- `TurnCompletedEvent.su_next_message` is `str | None`, and it is None on the case's final turn: the drive loop skips the synthetic-user call when no further target turn will consume it, so don't treat a null there as an error.
- Cost is attempt-inclusive but split so you can add it up honestly: `CaseCompletedEvent.total_cost` is the surviving chain only, and `CaseCompletedEvent.discarded_attempts_cost` is the real provider spend of that case's earlier failed attempts (chains deleted, billing real) — add the two for what the case actually cost. `CaseFailedEvent.total_cost` is the spend across all of a failed case's attempts (nothing survives on disk). `BatchCompletedEvent.total_cost` already rolls all of that up.

If your stream currently derives turn numbers by counting or omits the discarded-attempt spend, those two are the ones to double-check.

chiang-daniel:
**Score-key hazard — shared contract**

Samantha — a shared gotcha worth stating once, since your starter-code fix and our result-badge fix both live on it: deterministic eval scores are keyed by the eval's spec-name json_keys, not a fixed literal like `"match"`. Anything that reads a hardcoded score key (starter code, a result renderer, a fixture) works only against fabricated data and silently does nothing on real results — that's exactly how the Pass/Fail badge never rendered. The safe contract is to read the score values generically (they're uniform binaries) or resolve keys from the eval's declared output scores, never from a literal. Flagging so the fix in #1630's starter code and ours don't drift back to a hardcoded key.

leonardmq (approved):

chiang-daniel:

PR #1652, opened 2026-08-05
[eb-v2 review] Eval V2 core: sandbox, trace persistence, and datamodel guard fixes

chiang-daniel:

Surface: the V2 eval data model and the eval adapters (sandbox scorer runner, LLM judge, runner). Pure Python; about half the diff is tests. Three commits: d625c94f6 (landed as c22bb4799) — make the EvalConfig properties dispatch real, validate EvalInput tags, rewrite the EvalRun docstring, drop a dead tool-id branch; fdeda160d (landed as 1a5f81c9b) — drain the scorer queue before joining so large scorer output isn't misread as a timeout, handle SystemExit from user scorers, restore single-turn full_trace trace persistence, fail the g_eval logprobs preflight loudly, classify template-authoring errors correctly, parser-based trace-reference detection, superseded-tombstone dedup parity; 551107df4 (landed as 1e19a1883) — make the EvalRun full_trace trace gate V2-aware. Four **Personal review requested** comments flag the contract/invariant calls.

Note: several of these fixes dedupe when #1454 syncs into the branch — they touch the same datamodel/adapter regions, so expect to resolve to this behavior and adopt any copy updates from that PR.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_runner.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_runner.py
@@ -261,19 +261,26 @@ class EvalRunner:
 
         # already_run[eval_config_id][run_config_id][dataset_id]
         already_run: Dict[ID_TYPE, Dict[ID_TYPE, Set[ID_TYPE]]] = {}
+        superseded: Dict[Tuple[ID_TYPE, ID_TYPE, ID_TYPE], List[EvalRun]] = {}
         for eval_config in self.eval_configs:
             already_run[eval_config.id] = {}
             for run_config in self.run_configs or []:
                 already_run[eval_config.id][run_config.id] = set()
             for run in eval_config.runs(readonly=True):
                 if (
-                    run.task_run_config_id is not None
-                    and run.task_run_config_id in already_run[eval_config.id]
-                    and self._counts_as_already_run(run, eval_config)
+                    run.task_run_config_id is None
+                    or run.task_run_config_id not in already_run[eval_config.id]
                 ):
+                    continue
+                if self._counts_as_already_run(run, eval_config):
                     already_run[eval_config.id][run.task_run_config_id].add(
                         run.dataset_id
                     )
+                else:
+                    superseded.setdefault(
+                        (eval_config.id, run.task_run_config_id, run.dataset_id),
+                        [],
+                    ).append(run)
 
         return [
             EvalJob(
@@ -281,6 +288,9 @@ class EvalRunner:
                 task_run_config=run_config,
                 type="task_run_eval",
                 eval_config=eval_config,
+                superseded_tombstones=superseded.get(
+                    (eval_config.id, run_config.id, task_run.id), []
+                ),
             )
             for task_run in self.task.runs(readonly=True)
             if filter(task_run)
@@ -646,6 +656,22 @@ class EvalRunner:
             eval_task_input = EvalTaskInput.from_eval_input(job.item, run_output)
             result = await evaluator.evaluate(eval_task_input)
 
+            # Like the legacy runner, only successful task-run-eval records of
+            # a full_trace eval carry the serialized trace (single-shot
+            # generations don't always produce one).
+            trace_json: str | None = None
+            if (
+                result.skipped_reason is None
+                and self.eval.evaluation_data_type == EvalDataType.full_trace
+                and eval_task_input.trace
+            ):
+                trace_json = json.dumps(
+                    eval_task_input.trace,
+                    indent=2,
+                    ensure_ascii=False,
+                    default=_trace_json_default,
+                )
+
             async with self._save_context():
                 eval_run = EvalRun(
                     parent=job.eval_config,
@@ -666,6 +692,7 @@ class EvalRunner:
                     else None,
                     skipped_detail=result.skipped_detail,
                     intermediate_outputs=result.intermediate_outputs,
+                    task_run_trace=trace_json,
                     task_run_usage=run_output.usage,
                 )
                 eval_run.save_to_file()
@@ -681,6 +708,22 @@ class EvalRunner:
             # never persisted so its id is None (EvalRun would fail validation),
             # and dedup compares this field against the item's id.
             dataset_id = job.item.id
+
+            # Like the legacy runner, only successful task-run-eval records of
+            # a full_trace eval carry the serialized trace (single-shot
+            # generations don't always produce one).
+            trace_json: str | None = None
+            if (
+                result.skipped_reason is None
+                and self.eval.evaluation_data_type == EvalDataType.full_trace
+                and eval_task_input.trace
+            ):
+                trace_json = json.dumps(
+                    eval_task_input.trace,
+                    indent=2,
+                    ensure_ascii=False,
+                    default=_trace_json_default,
+                )
 
             async with self._save_context():
                 eval_run = EvalRun(
@@ -700,6 +743,7 @@ class EvalRunner:
                     else None,
                     skipped_detail=result.skipped_detail,
                     intermediate_outputs=result.intermediate_outputs,
+                    task_run_trace=trace_json,
                     task_run_usage=run_output.usage,
                 )
                 eval_run.save_to_file()
--- a/libs/core/kiln_ai/adapters/eval/eval_utils/v2_eval_helpers.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_utils/v2_eval_helpers.py
@@ -1,8 +1,8 @@
 import json
-import re
 from typing import Any
 
-from jinja2 import Undefined
+from jinja2 import TemplateSyntaxError, Undefined, meta, nodes
+from jinja2.parser import Parser
 from kiln_ai.datamodel.eval import (
     EvalOutputScore,
     EvalScores,
@@ -10,7 +10,7 @@ from kiln_ai.datamodel.eval import (
     SkippedReason,
     V2EvalResult,
 )
-from kiln_ai.utils.jinja_engine import JinjaExtractionError, extract
+from kiln_ai.utils.jinja_engine import JinjaExtractionError, _expression_env, extract
 
 
 def build_binary_scores(
@@ -26,14 +26,24 @@ def build_binary_scores(
     return {score.json_key(): value for score in output_scores}
 
 
-# Matches the `trace` identifier as a whole word (e.g. `trace`, `trace[-1]`,
-# `trace | length`) without matching substrings like `retrace` or `tracer`.
-_TRACE_REFERENCE_RE = re.compile(r"\btrace\b")
-
-
 def references_trace(expression: str) -> bool:
-    """True if a Jinja expression references the `trace` variable."""
-    return bool(_TRACE_REFERENCE_RE.search(expression))
+    """True if a Jinja expression references the `trace` variable.
+
+    Uses Jinja's own parser so only genuine variable references count --
+    'trace' appearing as data (a quoted string, a dict key, an attribute of
+    another variable) does not.
+    """
+    try:
+        # Same parse compile_expression performs, wrapped into a template node
+        # so meta can resolve the expression's undeclared (i.e. input) names.
+        parser = Parser(_expression_env, expression, state="variable")
+        template = nodes.Template([nodes.Output([parser.parse_expression()])], lineno=1)
+    except TemplateSyntaxError:
+        # Invalid expressions can't reference anything; the extraction path
+        # owns reporting the syntax error.
+        return False
+    template.set_environment(_expression_env)
+    return "trace" in meta.find_undeclared_variables(template)
 
 
 def extract_value(
--- a/libs/core/kiln_ai/adapters/eval/sandbox_worker.py
+++ b/libs/core/kiln_ai/adapters/eval/sandbox_worker.py
@@ -10,7 +10,9 @@ import multiprocessing
 import multiprocessing.queues
 import queue
 import sys
+import time
 import traceback
+from multiprocessing.process import BaseProcess
 from typing import Any
 
 from kiln_ai.sandbox.entrypoint import call_entrypoint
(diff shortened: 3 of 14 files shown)


libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -261,26 +261,36 @@ def collect_tasks_for_task_run_eval(self) -> List[EvalJob]:
 
         # already_run[eval_config_id][run_config_id][dataset_id]
         already_run: Dict[ID_TYPE, Dict[ID_TYPE, Set[ID_TYPE]]] = {}
+        superseded: Dict[Tuple[ID_TYPE, ID_TYPE, ID_TYPE], List[EvalRun]] = {}
```
chiang-daniel, line 264: The task-run-eval collector was missing the superseded-tombstone dedup map its sibling collectors already build, so it couldn't recognise runs superseded by a newer config and carry their tombstones onto the job. This adds the same map — keyed by (eval_config, run_config, dataset) — and threads superseded_tombstones onto each EvalJob, bringing this collector to parity with the others.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -646,6 +656,22 @@ async def _run_v2_job(self, job: EvalJob) -> bool:
             eval_task_input = EvalTaskInput.from_eval_input(job.item, run_output)
             result = await evaluator.evaluate(eval_task_input)
 
+            # Like the legacy runner, only successful task-run-eval records of
```
chiang-daniel, line 659: On the single-turn V2 lanes, a full_trace eval scored the conversation but never persisted the trace onto the EvalRun — the V1 validator was exempting V2 records from the trace requirement, so paid generations lost the conversation unrecoverably and the record couldn't be re-scored. This restores parity with the legacy lanes: a successful task-run eval of a full_trace eval serializes and stores the trace. (The same block appears on both the eval_input and dataset lanes.) The gate that let this slip silently is tightened separately in `551107df4`.

libs/core/kiln_ai/adapters/eval/eval_utils/v2_eval_helpers.py
```diff
@@ -26,14 +26,24 @@ def build_binary_scores(
...
-    """True if a Jinja expression references the `trace` variable."""
-    return bool(_TRACE_REFERENCE_RE.search(expression))
+    """True if a Jinja expression references the `trace` variable.
+
+    Uses Jinja's own parser so only genuine variable references count --
+    'trace' appearing as data (a quoted string, a dict key, an attribute of
+    another variable) does not.
+    """
+    try:
+        # Same parse compile_expression performs, wrapped into a template node
```
chiang-daniel, line 37: references_trace decided whether an expression reads the `trace` variable with a word-boundary regex, so `trace` appearing as data — a quoted string, a dict key, an attribute of another variable — falsely counted as a trace reference. This uses Jinja's own parser and undeclared-variable analysis, so only a genuine variable reference to `trace` counts. Invalid expressions return False here; the extraction path owns reporting their syntax errors.

libs/core/kiln_ai/adapters/eval/sandbox_worker.py
```diff
@@ -93,6 +95,19 @@ def _execute_scorer(
                 "stderr": captured_stderr.getvalue(),
             }
         )
+    except SystemExit as e:
```
chiang-daniel, line 98: A `sys.exit()` in user scorer code raises SystemExit, which the broad `except Exception` didn't catch — so the child died before queueing a result and the user saw a generic "process exited without returning results" with no clue why. This catches SystemExit and queues an explanatory error naming the exit code. KeyboardInterrupt is left to propagate on purpose: it means the operator interrupted the process group, not that the user's code failed.

libs/core/kiln_ai/adapters/eval/sandbox_worker.py
```diff
@@ -124,24 +139,59 @@ def run_scorer(
...
 
     return result
+
+
+def _wait_for_result(
+    p: BaseProcess,
+    q: multiprocessing.queues.Queue,  # type: ignore[type-arg]
+    timeout: float,
+) -> dict[str, Any] | None:
+    """Wait for the child's result dict; None on timeout or silent child death.
```
chiang-daniel, line 174: The runner joined the scorer process before reading its result queue. When a scorer returns more than the OS pipe buffer holds (~64KB), the child's queue feeder thread blocks until the parent reads, so the child never exits, the join times out, and a perfectly good run is reported as a timeout — which also stalls the batch behind the global lock. This inverts the wait: poll the queue in short slices (so a silently-dead child is noticed well before the full timeout), then reap the process. See the flag on the wait shape.

libs/core/kiln_ai/adapters/eval/sandbox_worker.py
```diff
@@ -124,24 +139,59 @@ def run_scorer(
...
+
+    The queue -- not the process -- is the primary wait: a result payload
+    larger than the OS pipe buffer keeps the child's queue feeder thread (and
+    so the child) alive until the parent reads it, so joining first would
+    deadlock and misreport a successful run as a timeout. Short poll intervals
+    let a child that died without reporting be noticed well before the full
+    timeout.
+    """
+    deadline = time.monotonic() + timeout
```
chiang-daniel, line 183: **Personal review requested:** The wait replaces join-first with a queue-first poll in 0.1s slices until the deadline, then force-reaps the process (with a final grace read to catch a result flushed between the empty read and the liveness check). Does the 0.1s slice / force-reap-after-grace shape look right to you, or would you prefer a different poll granularity or a hard join timeout as a backstop?

libs/core/kiln_ai/adapters/eval/v2_eval_llm_judge.py
```diff
@@ -111,6 +122,17 @@ async def evaluate(self, eval_input: EvalTaskInput) -> V2EvalResult:
                 skipped_detail=f"Template rendering failed: {e}",
             )
         except UndefinedError as e:
+            # A typo'd template variable and genuinely missing reference data
+            # both raise UndefinedError; only the latter is a data problem.
+            unknown = _unknown_template_variables(props.prompt_template, namespace)
```
chiang-daniel, line 127: A typo'd variable in the judge prompt template and a genuinely-missing reference-data key both raise Jinja's UndefinedError, and both were labeled missing_reference_key — so a template-authoring bug masqueraded as absent data. This distinguishes them by parsing the template for top-level names that aren't EvalTaskInput fields: unknown names are an authoring error (extraction_failed, naming them), everything else stays a data problem.

libs/core/kiln_ai/adapters/eval/v2_eval_llm_judge.py
```diff
@@ -132,7 +154,18 @@ async def evaluate(self, eval_input: EvalTaskInput) -> V2EvalResult:
 
         if props.g_eval:
             model_provider = built_in_models_from_provider(provider, model_name)
-            if model_provider is not None and not model_provider.supports_logprobs:
+            if model_provider is None:
```
chiang-daniel, line 157: The g_eval preflight only rejected models it could confirm lack logprobs. A model absent from the built-in list returned None and slipped through — then the paid judge call failed deterministically downstream, after spending. This fails the preflight loudly when the model can't be verified, before the call, with a message telling the author to pick a built-in logprobs model or turn off G-Eval.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -428,6 +432,17 @@ class EvalInput(KilnParentedModel):
         description="Tags for filtering eval inputs.",
     )
 
+    @model_validator(mode="after")
+    def validate_tags(self) -> Self:
```
chiang-daniel, line 436: EvalInput.tags were unvalidated, so an empty or space-containing tag could be written. Every tag filter treats those as unselectable, so the item silently disappeared from filtered views with no error at write time. This mirrors the TaskRun tag rules (no empty strings, no spaces) and rejects them at creation where the author can fix it.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -428,6 +432,17 @@ class EvalInput(KilnParentedModel):
...
+        # Empty or space-containing tags can't be selected by tag filters, so
+        # reject them at creation instead of silently dropping the item later.
+        for tag in self.tags:
```
chiang-daniel, line 439: **Personal review requested:** EvalInput.tags are now validated like TaskRun.tags (reject empty and space-containing), which is a deliberate tightening of your schema. Existing data with such tags would fail on the next revalidating write. Confirm you want the TaskRun rules applied to EvalInput, and that a write-time (not load-time) tightening is the right posture.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -545,13 +560,21 @@ def validate_type(self) -> Self:
 
 class EvalRun(KilnParentedModel):
     """
-    The results of running an eval on a single dataset item.
+    The result of running an eval on a single item, stored as a child of the
```
chiang-daniel, line 563: The EvalRun docstring still described a V1-only world, but the branch had already added the eval_input_id source and skipped-item semantics that contradicted it — so the class's own documentation was actively wrong about when task_run_config_id and the two input sources apply. Rewritten to state the real contract: the two eval purposes, the mutually-exclusive dataset_id / eval_input_id sources, and that output/scores may be absent only when skipped.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -625,29 +648,44 @@ def validate_input_source(self) -> Self:
         return self
 
     @model_validator(mode="after")
-    def validate_output_fields(self) -> Self:
+    def validate_output_fields(self, info: ValidationInfo) -> Self:
```
chiang-daniel, line 651: **Personal review requested:** I chose enforcement over exemption for the full_trace trace gate: single-shot no-trace runs are held to it, and an MCP+full_trace pairing that produces no trace now fails at save, matching V1. The argument is V1 parity — V1 already required the trace for exactly this shape. Your call to ratify: is holding V2 single-shot full_trace runs to the trace requirement correct, or is there a legitimate V2 full_trace flow that produces no trace and should be exempt?

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -625,29 +648,44 @@ def validate_input_source(self) -> Self:
...
+        # files predating this gate are exempt so they still load; new writes
+        # and rebuilds are held to it.
+        if (
+            not self.eval_config_eval
+            and self.skipped_reason is None
+            and evaluation_data_type == EvalDataType.full_trace
+            and self.task_run_trace is None
+            and not self.loaded_from_file(info)
+        ):
+            raise ValueError("full_trace task run eval runs should include trace")
```
chiang-daniel, line 673: The old validate_output_fields returned early for every V2 config, which is exactly why the missing-trace bug above could persist silently — the alarm was disconnected for V2. This makes the full_trace-requires-trace gate apply to both V1 and V2 writers of a scored, non-skipped task-run eval, so a writer that drops the trace now fails loudly instead of saving an un-re-scorable record. Historical files predating the gate are exempt via loaded_from_file so they still load. See the flag — enforcement was chosen over exemption here.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -730,26 +768,25 @@ class EvalConfig(KilnParentedModel, KilnParentModel, parent_of={"runs": EvalRun}
         default=EvalConfigType.g_eval,
         description="This is used to determine the type of eval to run.",
     )
-    properties: V2EvalConfigProperties | dict[str, Any] | None = Field(
+    properties: dict[str, Any] | V2EvalConfigProperties | None = Field(
```
chiang-daniel, line 771: **Personal review requested:** The legacy/V2 properties-parsing contract is now explicit: the union lists `dict` first so any raw dict stays a plain dict, and V2 configs are parsed into the typed union by a module-level TypeAdapter in the before-validator. This is your data model — please sign off on the mechanism (field-order-as-contract plus TypeAdapter dispatch) rather than relying on discriminated-union fallback ordering.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -730,26 +768,25 @@ class EvalConfig(KilnParentedModel, KilnParentModel, parent_of={"runs": EvalRun}
...
 
     @model_validator(mode="before")
     @classmethod
-    def dispatch_properties_parsing(cls, data: Any, info: ValidationInfo) -> Any:
-        # Pydantic's discriminated-union parsing would reject a plain dict for
-        # `properties` because dicts don't carry a discriminator field. V1 (legacy)
-        # configs store properties as an untyped dict, so we shallow-copy and
-        # re-assign it here to force Pydantic to accept the dict branch of the union.
+    def dispatch_properties_parsing(cls, data: Any) -> Any:
+        # The union lists dict first, so a raw dict always stays a plain dict —
```
chiang-daniel, line 779: The old dispatch validator was a no-op: it shallow-copied the properties dict back onto itself, and legacy dicts only loaded by luck of the union's fallback ordering. A legacy config whose properties happened to carry a colliding "type" key was unloadable — the discriminated union would try to parse it as a typed V2 shape and reject the whole file. This makes the dispatch real: legacy dicts stay plain dicts (the union lists dict first), and V2 configs are parsed into the typed union up front via a module-level TypeAdapter. The field order flip on `properties` (dict first) is part of the same contract — see the flag.

libs/core/kiln_ai/datamodel/tool_id.py
```diff
@@ -127,11 +127,8 @@ def _check_tool_id(id: str) -> str:
 
     # Code tools must have format: kiln_tool::code::<code_tool_id>
     if id.startswith(CODE_TOOL_ID_PREFIX):
-        ct_id = code_tool_id_from_tool_id(id)
-        if not ct_id:
-            raise ValueError(
-                f"Invalid code tool ID: {id}. Expected format: 'kiln_tool::code::<code_tool_id>'."
-            )
+        # Raises ValueError on malformed IDs; the extracted ID isn't needed here.
```
chiang-daniel, line 130: Dead defensive branch: this guarded `if not ct_id: raise`, but code_tool_id_from_tool_id already raises on a malformed id, so the extracted value is never falsy by the time the check runs — the branch could never fire. Removed, keeping the raising call for its validation side effect and noting that the return value isn't used here.

chiang-daniel:

PR #1654, opened 2026-08-05
[eb-v2 review] Git import: bind the trust prompt to the repo it was granted for — sign-off requested

chiang-daniel:

Surface: the Git-import project wizard and the Code Tool test panel. One commit, `e9c4b294c (landed as ccf082f54)` — bind git-import trust to the repo URL it was granted for (any URL change clears the stale clone/branch/project state that the trust-skip relies on), refuse a local-trust import with no selected path, and stop the code-tool test panel from silently swallowing input-validation errors. Roughly half the diff is the wizard-stub harness and tests. The trust-model trade-offs are called out in two **Personal review requested** comments below — those are the ones I'd like your explicit sign-off on.

### DIFF
--- a/app/web_ui/src/lib/components/code_tools/code_tool_test_panel.svelte
+++ b/app/web_ui/src/lib/components/code_tools/code_tool_test_panel.svelte
@@ -40,6 +40,9 @@
 
   // Stores the last-built param values for preview display
   let param_values: Record<string, unknown> = {}
+  // Validation error from building the edit-inputs form. Shown inside the edit
+  // dialog so a bad input keeps the dialog open instead of silently closing.
+  let input_error: KilnError | null = null
 
   function parse_schema(
     schema: { [key: string]: unknown } | undefined,
@@ -71,13 +74,19 @@
   function save_and_close_inputs(): boolean {
     try {
       param_values = build_params()
-    } catch {
-      // Let build_params errors surface at run time
+      input_error = null
+      return true
+    } catch (e) {
+      // A required field is missing or an object param is invalid JSON. Keep
+      // the dialog open and surface the error so the typed inputs aren't
+      // discarded and Run Test can't post stale values.
+      input_error = createKilnError(e)
+      return false
     }
-    return true
   }
 
   function open_edit_inputs() {
+    input_error = null
     edit_inputs_dialog.show()
   }
 
@@ -372,6 +381,13 @@
           bind:this={input_components[prop.id]}
         />
       {/each}
+      {#if input_error}
+        <Warning
+          warning_color="error"
+          tight
+          warning_message={input_error.getMessage()}
+        />
+      {/if}
     </div>
   {/if}
 </Dialog>
--- a/app/web_ui/src/lib/components/import/import_project.svelte
+++ b/app/web_ui/src/lib/components/import/import_project.svelte
@@ -103,6 +103,14 @@
   }
 
   function redirect_if_missing_state(step: WizardStep): boolean {
+    // The local trust step depends on the component-local selected file path,
+    // which the store-only validate_step_requirements can't see. A remount or
+    // deep-link to #local-trust starts with no path, so send the user back to
+    // pick a file rather than letting Trust Project import an empty path.
+    if (step === "local_trust_confirm" && !import_project_path) {
+      set_step("local_file")
+      return true
+    }
     if (!validate_step_requirements(step)) {
       clear_wizard_store()
       replaceState(window.location.pathname + window.location.search, {})
@@ -147,7 +155,10 @@
   onMount(() => {
     const url_param = read_url_query_param("url")
     if (url_param) {
-      update_store({ git_url: url_param })
+      // Seed the url step from the deep-link param. Route through adopt_git_url
+      // so a param pointing at a different repo than the persisted session
+      // clears its stale downstream state (and its trust skip).
+      adopt_git_url(url_param)
       if (window.location.hash !== "#git") {
         window.location.hash = "#git"
       }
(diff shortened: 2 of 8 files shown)


app/web_ui/src/lib/components/code_tools/code_tool_test_panel.svelte
```diff
@@ -71,13 +74,19 @@
...
       param_values = build_params()
-    } catch {
-      // Let build_params errors surface at run time
+      input_error = null
+      return true
+    } catch (e) {
+      // A required field is missing or an object param is invalid JSON. Keep
+      // the dialog open and surface the error so the typed inputs aren't
+      // discarded and Run Test can't post stale values.
+      input_error = createKilnError(e)
```
chiang-daniel, line 83: The edit-inputs dialog built its param values in a try with an empty catch: when build_params threw (a missing required field, or an object param that isn't valid JSON) the error was swallowed, the dialog closed as if it succeeded, and the just-typed inputs were dropped — Run Test then posted the previous, stale values. Now a build failure keeps the dialog open, returns false, and renders the error inline, so bad input is fixed instead of silently discarded.

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -103,6 +103,14 @@
   }
 
   function redirect_if_missing_state(step: WizardStep): boolean {
+    // The local trust step depends on the component-local selected file path,
+    // which the store-only validate_step_requirements can't see. A remount or
+    // deep-link to #local-trust starts with no path, so send the user back to
+    // pick a file rather than letting Trust Project import an empty path.
+    if (step === "local_trust_confirm" && !import_project_path) {
```
chiang-daniel, line 110: The local-trust step's only real input — the selected file path — lives in component state, which the store-only step validator can't see. A remount or a deep-link straight to `#local-trust` arrives with no path, and the old code let the user reach "Trust Project" and POST an empty path into a server error. This sends them back to the file picker instead. It's paired with the guard in on_local_trust_confirmed so an empty path can't be imported by either route.

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -147,7 +155,10 @@
   onMount(() => {
     const url_param = read_url_query_param("url")
     if (url_param) {
-      update_store({ git_url: url_param })
+      // Seed the url step from the deep-link param. Route through adopt_git_url
+      // so a param pointing at a different repo than the persisted session
+      // clears its stale downstream state (and its trust skip).
+      adopt_git_url(url_param)
```
chiang-daniel, line 161: Deep links carry a `?url=` param. Previously this wrote git_url directly and left any persisted session state untouched, which is the same trust-skip hole as the Back-nav path: land on the wizard with a stale clone_path for repo A, deep-link to repo B, and B inherits A's passed trust. Seeding through adopt_git_url means a deep link to a different repo clears that state like any other URL change.

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -166,13 +177,39 @@
     set_step("credentials")
   }
 
+  // Adopt the repo URL the user entered/confirmed on the url step. Trust is
+  // granted per-repo, and clone_path/branch/project all describe whichever repo
+  // was entered before. When the URL changes we drop that downstream state so
+  // it can never be mistaken for the current repo — in particular so the
+  // clone_path-based trust skip below only fires for the same repo the trust
+  // gate was passed for.
+  function adopt_git_url(
```
chiang-daniel, line 186: The trust gate was granted per-repo, but nothing tied it to the repo it was granted for. clone_path, selected_branch and the selected project all described whichever repo was entered earlier; changing the URL and coming back (e.g. via browser Back) left that state in place, so a different repo could sail past the trust prompt on the strength of the previous repo's clone. This routes every URL entry through one choke point that drops the downstream state whenever the URL actually changes, so the clone_path-based trust-skip below can only ever fire for the repo currently being imported.

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -166,13 +177,39 @@
...
+    url: string,
+    extra_fields: Partial<typeof $git_import_wizard_store> = {},
+  ) {
+    const url_changed = url !== $git_import_wizard_store.git_url
+    update_store({
```
chiang-daniel, line 191: **Personal review requested:** This makes the trust prompt fail-safe-annoying — re-entering the *same* repo URL still skips the prompt, but any *different* URL re-prompts, including a differently-spelled but equivalent one (trailing slash, .git suffix, http vs https, ssh vs https for the same repo). Is re-prompting on equivalent spellings the right trade for your trust model, or do you want URL normalization before the comparison?

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -166,13 +177,39 @@
...
+      git_url: url,
+      ...extra_fields,
+      ...(url_changed
+        ? {
+            clone_path: "",
```
chiang-daniel, line 196: **Personal review requested:** A URL change clears clone_path, branch and project selection, but deliberately keeps the stored credentials/auth — the credentials step re-verifies them against the new repo before any clone. So a token entered for repo A survives a switch to repo B and is re-checked rather than re-entered. Confirm that carrying credentials across a URL change matches your trust intent (the alternative is clearing them too and forcing re-entry).

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -190,9 +227,11 @@
         auth_mode: detected_auth_method,
       })
     }
-    // If clone_path is set, the user already passed the trust gate and the
-    // branch step redirected back here for credentials — skip trust and
-    // return directly to branch.
+    // A clone_path here means this repo already reached the branch step, which
```
chiang-daniel, line 230: This is the trust-skip itself: a set clone_path is taken as proof the repo already passed trust and the branch step merely bounced back for credentials. That inference is only sound once clone_path is guaranteed to belong to the current repo, which the adopt_git_url change now enforces. The comment is rewritten to state that dependency so the skip isn't later "simplified" back into a hole.

chiang-daniel:

PR #1655, opened 2026-08-05
[eb-v2 review] Code Tools: parameter validation, allowlist isolation, and wizard fixes

chiang-daniel:

Surface: code-tool execution runtime, the CodeTool data model, and the code-tool authoring wizard. Commits: `afe821bd1 (landed as c8876814e)` (isolate broken allowlist entries so one dead entry can't take the whole nested-tool surface down; carry tool timeouts as a typed flag instead of sniffing output text), `4a63c3336 (landed as 34e6367a0)` (reject reserved-word parameter names at authoring and generate docstrings that stay valid for any description), `9665236fc (landed as 5dd2487f8)` (preserve authored code when returning to the wizard's code step), plus one hunk of `1c9a0cb1d (landed as one hunk of 6f1afc872)` (make FormList row add/remove emit a change event so ancestor handlers notice structural edits). Around half the diff is tests. Three **Personal review requested** comments flag the layer/posture choices.

### DIFF
--- a/app/web_ui/src/lib/utils/code_tool_helpers.ts
+++ b/app/web_ui/src/lib/utils/code_tool_helpers.ts
@@ -156,6 +156,81 @@ export function isCodeUnmodified(
   return currentCode.trim() === "" || currentCode === originalPlaceholder
 }
 
+export interface Step2CodeResolution {
+  code: string
+  generatedPlaceholder: string
+  schemaChangedHint: boolean
+  // Whether the clone's seed code was used. When true the caller drops it so a
+  // later return to the Code step doesn't overwrite the user's edits with it.
+  cloneConsumed: boolean
+}
+
+/**
+ * Decide the Code step's editor contents when the user advances from the
+ * Define step.
+ *
+ * Keyed on whether code already exists — not on the wizard step — because the
+ * browser Back button pops the shallow-routing step state, so the step alone
+ * can't tell a first visit from a return. Defaulting to "first visit" would
+ * overwrite the user's authored code with a fresh placeholder. The cases:
+ *
+ *  - No code yet: seed from the clone (if any), else the generated placeholder.
+ *  - Untouched placeholder: regenerate it for the (possibly changed) schema.
+ *  - User-authored code: preserve it, and flag when the schema changed so the
+ *    user knows to reconcile run()'s parameters.
+ */
+export function resolveStep2Code(args: {
+  code: string
+  newPlaceholder: string
+  generatedPlaceholder: string
+  schemaChangedHint: boolean
+  cloneCode: string | null
+}): Step2CodeResolution {
+  const {
+    code,
+    newPlaceholder,
+    generatedPlaceholder,
+    schemaChangedHint,
+    cloneCode,
+  } = args
+
+  if (code === "") {
+    const useClone = !!cloneCode
+    return {
+      code: useClone ? (cloneCode as string) : newPlaceholder,
+      generatedPlaceholder: newPlaceholder,
+      schemaChangedHint: false,
+      cloneConsumed: useClone,
+    }
+  }
+
+  if (isCodeUnmodified(code, generatedPlaceholder)) {
+    return {
+      code: newPlaceholder,
+      generatedPlaceholder: newPlaceholder,
+      schemaChangedHint: false,
+      cloneConsumed: false,
+    }
+  }
+
+  if (newPlaceholder !== generatedPlaceholder) {
+    return {
+      code,
+      generatedPlaceholder: newPlaceholder,
+      schemaChangedHint: true,
+      cloneConsumed: false,
+    }
+  }
+
+  // User-authored code and the schema is unchanged: leave everything as-is.
+  return {
+    code,
+    generatedPlaceholder,
+    schemaChangedHint,
+    cloneConsumed: false,
+  }
+}
+
 /**
  * Format a parameter value for inline preview display.
  *
--- a/app/web_ui/src/lib/utils/code_tool_helpers.ts
+++ b/app/web_ui/src/lib/utils/code_tool_helpers.ts
@@ -58,51 +58,6 @@ function jsonTypeToPython(prop: JsonSchemaProperty): string {
   }
 }
 
-const PYTHON_RESERVED_WORDS = new Set([
-  "False",
-  "None",
-  "True",
-  "and",
-  "as",
-  "assert",
-  "async",
-  "await",
-  "break",
-  "class",
-  "continue",
-  "def",
-  "del",
-  "elif",
-  "else",
-  "except",
-  "finally",
-  "for",
-  "from",
-  "global",
-  "if",
-  "import",
-  "in",
-  "is",
-  "lambda",
-  "nonlocal",
-  "not",
-  "or",
-  "pass",
-  "raise",
-  "return",
-  "try",
-  "while",
-  "with",
-  "yield",
-])
-
-/**
- * Escape a parameter name if it collides with a Python reserved word.
- */
-function safePythonName(name: string): string {
-  return PYTHON_RESERVED_WORDS.has(name) ? `${name}_param` : name
-}
-
 export interface CodeToolParam {
   name: string
   pythonType: string
@@ -141,12 +96,14 @@ function buildParamList(params: CodeToolParam[]): string {
   const requiredParams = params.filter((p) => p.required)
   const optionalParams = params.filter((p) => !p.required)
 
+  // Schema property names are used verbatim: the sandbox calls run() with
+  // keyword arguments keyed by these exact names, so renaming any would break the call.
   const parts: string[] = []
   for (const p of requiredParams) {
-    parts.push(`${safePythonName(p.name)}: ${p.pythonType}`)
+    parts.push(`${p.name}: ${p.pythonType}`)
   }
   for (const p of optionalParams) {
-    parts.push(`${safePythonName(p.name)}: ${p.pythonType} | None = None`)
+    parts.push(`${p.name}: ${p.pythonType} | None = None`)
   }
   return parts.join(", ")
 }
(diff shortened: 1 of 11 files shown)


app/web_ui/src/lib/utils/code_tool_helpers.ts
```diff
@@ -141,12 +96,14 @@ function buildParamList(params: CodeToolParam[]): string {
   const requiredParams = params.filter((p) => p.required)
   const optionalParams = params.filter((p) => !p.required)
 
+  // Schema property names are used verbatim: the sandbox calls run() with
```
chiang-daniel, line 99: This deletes safePythonName and the reserved-word table it used to rewrite parameter names in the generated stub. That rewrite was the actual bug behind the datamodel guard: it silently produced a stub whose parameter name didn't match the schema, so the sandbox's keyword call could never reach it. Now the stub uses the schema names verbatim and the reserved-word case is rejected at authoring instead of being papered over here.

app/web_ui/src/lib/utils/code_tool_helpers.ts
```diff
@@ -160,9 +117,9 @@ export function generateCodeToolPlaceholder(
 ): string {
   const params = extractParams(schema)
   const paramList = buildParamList(params)
-  // Prevent the description from prematurely closing the triple-quoted docstring.
-  // Replace any embedded """ with '' (two single-quotes) which is safe inside """.
-  const safeDesc = toolDescription.replace(/"""/g, "''")
+  // Escape backslashes then double-quotes so the description can neither close the
+  // triple-quoted docstring nor leave a dangling escape (e.g. a trailing " or \).
+  const safeDesc = toolDescription.replace(/\\/g, "\\\\").replace(/"/g, '\\"')
```
chiang-daniel, line 122: The placeholder generator embeds the tool description inside a triple-quoted docstring. The old escaping only replaced literal `"""`, which left two ways to emit non-compiling Python: a description ending in a backslash escaped the closing quote, and a stray double-quote could still break out. Escaping backslashes first, then double-quotes, makes the description safe for any input — verified round-trip in the tests.

app/web_ui/src/lib/utils/code_tool_helpers.ts
```diff
@@ -199,6 +156,81 @@ export function isCodeUnmodified(
...
+  // Whether the clone's seed code was used. When true the caller drops it so a
+  // later return to the Code step doesn't overwrite the user's edits with it.
+  cloneConsumed: boolean
+}
+
+/**
+ * Decide the Code step's editor contents when the user advances from the
+ * Define step.
+ *
+ * Keyed on whether code already exists — not on the wizard step — because the
```
chiang-daniel, line 172: The wizard decided the code editor's contents from current_step, but the documented Back path pops the shallow-routing history so current_step reads "define" on a return — the discriminator was effectively always "first visit", and the code-preserving branch was unreachable. So returning to the code step wiped the user's authored code with a fresh placeholder. This pure helper keys the decision on whether code already exists instead of on the step, and the page (+page.svelte) now calls it. The clone seed is consumed once so a later return can't re-apply it over edits.

app/web_ui/src/lib/utils/form_list.svelte
```diff
@@ -13,6 +13,19 @@
...
+  // The add button, used as a stable in-tree anchor to announce row changes.
+  let add_button_el: HTMLButtonElement | undefined
+
+  // Adding/removing a row is a plain button click, which fires no native
+  // input/change event. Emit a bubbling change like a native control would
+  // (mirrors fancy_select) so ancestor on:change handlers — e.g. unsaved-changes
+  // and tested-state tracking — react to structural row edits. The initial
+  // start_with_one seed runs before mount, so add_button_el is still undefined
+  // then and no spurious event fires.
+  function announce_row_change() {
```
chiang-daniel, line 25: Adding or removing a FormList row is a plain button click, which fires no native input/change event — so ancestor `on:change` handlers (the unsaved-changes guard, and the eval builder's tested-state tracking) never saw structural row edits. This dispatches a bubbling change from a stable in-tree anchor when a row is added or removed, the way a native control would. The start_with_one seed runs before mount, while the anchor is still undefined, so no spurious event fires on init.

libs/core/kiln_ai/datamodel/code_tool.py
```diff
@@ -71,8 +72,20 @@ def validate_function_name(cls, v: str) -> str:
         return v
 
     @model_validator(mode="after")
-    def validate_parameters_schema(self) -> Self:
+    def validate_parameters_schema(self, info: ValidationInfo) -> Self:
         validate_schema_dict(self.parameters_schema, require_object=True)
+
+        # Parameters are passed to run() as keyword arguments keyed by these names,
```
chiang-daniel, line 78: **Personal review requested:** Reserved-word rejection lives in the data-model validator with a loaded_from_file exemption, so an already-broken tool still loads but a PATCH that revalidates it now fails loudly. Do you agree with the layer (data model rather than API or client) and the posture (load old broken tools silently, reject on the next write)? The alternative postures are reject on load too, or warn-not-block on write.

libs/core/kiln_ai/datamodel/code_tool.py
```diff
@@ -71,8 +72,20 @@ def validate_function_name(cls, v: str) -> str:
...
+        # and a Python reserved word can never be declared as a run() parameter, so
+        # reject it at authoring. Loaded files are exempt so old projects still open.
+        # Soft keywords (match/case/type/_) are valid identifiers and stay allowed.
```
chiang-daniel, line 81: **Personal review requested:** The check uses keyword.iskeyword, which deliberately allows soft keywords (match, case, type, _) because they are valid identifiers and work fine as run() parameters — only hard keywords are rejected. Confirm you want soft keywords allowed rather than reserved defensively (the stricter alternative would use keyword.issoftkeyword too).

libs/core/kiln_ai/datamodel/code_tool.py
```diff
@@ -71,8 +72,20 @@ def validate_function_name(cls, v: str) -> str:
...
+        if not self.loaded_from_file(info):
+            for param_name in self.parameters_schema.get("properties", {}):
+                if keyword.iskeyword(param_name):
+                    raise ValueError(
```
chiang-daniel, line 85: A parameter named after a Python keyword (e.g. `from`, `class`) is impossible: the sandbox invokes run() with keyword arguments keyed by the exact schema property names, and no reserved word can be a run() parameter. The old client helper hid this by renaming the parameter in the generated stub, so the stub declared `from_param` while the sandbox still called `from=...` — the tool was permanently broken and authoring never warned. Rejecting it here, in the data model, catches every write path. loaded_from_file is exempt so existing projects still open (see the flag on the posture).

libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -41,6 +41,10 @@ class ToolCallResult(BaseModel):
         default=None,
         description="Human-readable error message, set when is_error is True.",
     )
+    timed_out: bool = Field(
```
chiang-daniel, line 44: This is the structural signal that replaces the text-sniffing above. Adding it to ToolCallResult lets the timeout be reported once, where it's known (the execution runtime), and read as a boolean everywhere else — no caller has to re-derive "was this a timeout?" by parsing a human-readable message.

libs/core/kiln_ai/tools/code_tool.py
```diff
@@ -372,10 +374,9 @@ async def _serve(
             result = await tool.run(context, **arguments)
 
             if result.is_error:
-                is_timeout = isinstance(tool, PythonCodeTool) and "timed out" in (
-                    result.output or ""
-                )
-                kind = "timeout" if is_timeout else "call_error"
+                # The result carries timeouts as a typed flag; ordinary failures
+                # whose text merely mentions a timeout must stay call_error.
+                kind = "timeout" if result.timed_out else "call_error"
```
chiang-daniel, line 379: Nested-call timeout classification was a substring match on the result text (`"timed out" in output`). Any ordinary failure whose message happened to mention a timeout was misclassified as a real ToolTimeout, and conversely a timeout worded differently would be missed. The result now carries the timeout as a structural `timed_out` flag (added on ToolCallResult), so classification reads a boolean the runtime sets rather than guessing from free text.

libs/core/kiln_ai/tools/code_tool.py
```diff
@@ -467,7 +468,18 @@ async def _build_name_map(self) -> dict[str, list[ToolId]]:
 
         name_map: dict[str, list[ToolId]] = {}
         for tool_id in self._code_tool.tool_allowlist:
-            fn_name = await self._canonical_tool_name(tool_id)
+            try:
+                fn_name = await self._canonical_tool_name(tool_id)
+            except Exception as exc:
+                # A broken entry (e.g. a deleted RAG config) must not take the
```
chiang-daniel, line 474: A single unresolvable allowlist entry — e.g. a nested tool pointing at a RAG config the user deleted — used to throw while building the name map and take the entire nested-tool surface down with it, so every healthy tool became uncallable. This isolates the failure per entry: a broken one is logged and left out of the map, and calls to it fail with the standard "not available" error while the rest keep working. list_tools() gets the matching treatment just below.

libs/core/kiln_ai/tools/code_tool.py
```diff
@@ -488,31 +500,39 @@ async def _canonical_tool_name(self, tool_id: str) -> str:
...
+                    exc,
                 )
-            except Exception:
-                fn_name = await self._canonical_tool_name(tool_id)
                 result.append(
                     {
-                        "name": fn_name,
-                        "description": "(unavailable)",
+                        "name": tool_id,
+                        "description": f"(unavailable: {exc})",
```
chiang-daniel, line 524: **Personal review requested:** This changes the list_tools contract: a broken allowlist entry now renders as `(unavailable: <reason>)` under its raw id, instead of failing the whole call. That's a deliberate degraded-UX choice — the surface stays usable and the broken entry is visible, but its intended name is unknowable without resolving it. Confirm you're happy exposing the id and the raw exception text here, versus a generic "unavailable" with no detail.

libs/core/kiln_ai/tools/tool_registry.py
```diff
@@ -146,8 +146,8 @@ def tool_from_id_and_project(
                 f"Code tool not found: {ct_id} in project {project.id} for tool {tool_id}"
             )
 
-        # Lazy import — PythonCodeTool is phase 2; for now return a
-        # lightweight wrapper that has the tool's identity and definition.
+        # Lazy import — PythonCodeTool pulls in the multiprocessing execution
```
chiang-daniel, line 149: The comment described a design that no longer exists — a "phase 2 / lightweight wrapper" placeholder — when this returns the full PythonCodeTool execution runtime. Left as-is it would mislead the next reader into thinking the real tool wasn't wired up yet. Rewritten to state the actual reason the import is lazy: it pulls in the multiprocessing runtime that most registry lookups don't need.

chiang-daniel:

PR #1657, opened 2026-08-06
(do not merge): eval splits in the eval's own item source, not only TaskRuns

leonardmq:
## Problem

`split=train|val|test` hard-422s on **every** EvalInput-backed (V2) eval. `split_filter_id_from_eval` rejected them up front (`eval_api.py:132`), and both the jobs worker (`jobs/workers/eval.py`) and the results endpoint routed through it. Splits were modelled as TaskRun dataset filters, so a V2 eval — which scopes items by `eval_input_filter_id` — had nowhere to resolve one.

This blocks Auto-Research entirely on V2 eval suites: AR measures experiments on **train**, gates promotion on **val**, and leaves **test** frozen for reporting. Every `POST /api/jobs/evals/run?split=…` and every `GET …/results?split=…` 422s.

It also fails *silently* from AR's side: AR's setup readiness audit only checks that `train_set_filter_id`/`val_set_filter_id` are non-null, and the lazy migration mints those on every eval load — so the audit passes on a project where wave 1's first measurement is guaranteed to fail.

## Approach

The two filter-id types already share the `tag::<tag>` form (`dataset_filters.py:252`), so this is a **resolution** change, not a data change — no migration, no re-tagging.

`split_filter_id_from_eval` becomes `resolve_split`, returning a `ResolvedSplit` that carries exactly one of the two filter ids:

| eval type | `split=test` | `split=train\|val` |
|---|---|---|
| V1 (TaskRun-backed) | `eval_set_filter_id` | `{train,val}_set_filter_id` as a dataset filter |
| **V2 (EvalInput-backed)** | `eval_input_filter_id` — its test universe by definition | the stored tag, read as an **EvalInput** filter |

V1 behavior is unchanged. **Replace-universe semantics** throughout: the resolved filter replaces the eval's default item filter, it does not intersect with it.

Three call sites now follow the resolved kind rather than re-deriving the eval's source:

- **`EvalRunner`** gains `eval_input_filter_id_override`, the V2 counterpart of `eval_set_filter_id_override`. Each override binds to exactly one source mode and raises on the other (plus a both-at-once guard), so a split can never silently run a different item universe than the caller asked for.
- **The jobs worker's progress universe** collapses from three branches to two. This is the one that would have bitten quietly: measuring a V2 split's progress over TaskRuns — or over the full test set — makes `is_complete` wrong, and a resumed job can short-circuit to `succeeded` with real work outstanding.
- **The results endpoint** matches on `eval_input_id` for V2, since a V2 `EvalRun` carries no `dataset_id`.

A V1-only filter id stored in a V2 eval's train slot (e.g. `high_rating`) cannot scope EvalInputs; that 422s at the resolver rather than raising `ValueError` deep in the runner.

## Testing

15 new tests across the four layers — resolver/results endpoint, jobs API pre-resolve, jobs worker (`_build_eval_runner` + `compute_state`), and `EvalRunner` collection.

The three endpoint tests were verified to **fail against the pre-change source** (with the test file held constant) for the right reason — `"Eval 'eval1' is EvalInput-backed (V2); dataset splits are not supported for it."` — not vacuously.

`checks.sh`: `ty check`, full pytest, OpenAPI schema, and all web checks pass. `ruff check` reports 6 errors, all pre-existing on `agi-anyting_goes_into` (RUF003 in `judge_feedback_batch*`, files untouched here) — identical count with this branch's changes stashed.

## Relationship to `claude/eval-splits-v1-v2-q38412`

That branch is the proper long-term redesign (`SplitRef` union, a `splits` dict, provenance-preserving serialization) and should supersede this. But it is **2 of 7 phases done** — phases 4/5/6 (`EvalRunner`, jobs, API) are exactly the layers AR calls — and its phase 1 deliberately **backed out** the `split` surface entirely, so it is currently further from a working split than `agi` is. It also forked from a much earlier point and is ~74k lines behind `agi` (no memory API, copilot, multiturn SDG, or eval-builder endpoints).

So this PR is scoped to be **easy to delete**: it adds a V2 branch inside the existing resolvers rather than refactoring them, and phase 6 of that project removes these guards outright.


leonardmq:
Obsolete - now eval v2 merged, no longer need this temp workaround

PR #1661, opened 2026-08-07
Train/val/test splits across both V1 (TaskRun) and V2 (EvalInput) eval data sources

scosman:
## What does this PR do?

Makes an eval's **train / val / test** splits work for both data sources, addressable through one API surface, so no caller branches on where an eval's items live.

Specs are in `specs/projects/eval_splits_v1_v2/` — `functional_spec.md` for behavior, `architecture.md` for design. Both are worth reading before the diff.

### The model

- Three splits: **test**, **train**, **val**.
- Each is backed by exactly **one** source — `TaskRun` or `EvalInput`. Never both.
- An eval **may mix** backings across its splits. That's a requirement, not an accident: it's how someone adds an `EvalInput`-backed val set to an existing TaskRun-backed eval.
- **Golden is not a split.** `eval_configs_filter_id` stays TaskRun-only and untouched.

### Storage

`Eval.splits` is a `str`-keyed dict over a discriminated union (`TaskRunSplit` | `EvalInputSplit`), so a TaskRun-only filter form on an EvalInput-backed split is unrepresentable rather than validated. The old "exactly one of `eval_set_filter_id` / `eval_input_filter_id`" validator is deleted rather than ported — one split holds one value, so the invariant is structural.

`eval_set_filter_id` and `train_set_filter_id` **stay declared** — project files sync between clients on different app versions, so removing a field is breaking. Nothing reads them; serialization is provenance-preserving, so an untouched eval round-trips **byte-identically** and only new splits land in the dict.

### The accessor

`resolve_split()` and `eval_run_item_key()` are the only places that branch on backing. Everything keys on `ItemKey = (source, id)` rather than a bare id — ids are `str(uuid.uuid4().int)[:12]` from one generator shared by every model type, so a cross-store collision is unlikely rather than impossible, and would silently admit one item's result into another's split.

### Behavior

- `POST /api/jobs/evals/run` takes an optional `split`; omitting it runs the test set exactly as today.
- `GET .../results` takes a **required** `split` — its old default was only accidentally correct, since every stored run is a test run until val becomes runnable.
- `EvalProgress` gains `val_dataset_size`; the `eval_set_filter_id is None` guards are gone, so progress and summaries report real numbers for either backing.
- `EvalRunner` takes a `ResolvedSplit`. The two collection paths collapse into one, so a split can't be accepted and silently ignored — there's no second path left to forget.
- Judge calibration over `EvalInput` becomes **unexpressible** rather than guarded; the code that manufactured a skipped `EvalRun` per item is deleted with the mis-routing that reached it.

### Notable

The train/val **lazy migration is removed**. Minting a split requires choosing a backing, and defaulting to TaskRun silently locks the legacy source onto splits nobody configured. It also manufactured a split that never had items.

`run_eval.svelte` moves off `EventSource`, which can't read the status or body of a non-200 — so the 4xx responses this PR adds actually reach the user instead of rendering "Unknown error".

### Testing

- 6 mutation-sweep artifacts (`specs/projects/eval_splits_v1_v2/phase*_mutation_sweep.py`), one per phase, all mutations killed and each attributed to a named test. They're re-run cumulatively: phase 2's still kill against a tree five phases downstream.
- They earned it — one mutation survived its first run, catching a duplicated membership loop in `get_run_config_eval_scores` that neither implementation nor review had noticed.
- Byte-identical round-trip tests are the gate on the serializer.

### Follow-on

`specs/projects/eb_v2_splits_alignment/project_overview.md` records what aligning the eb-v2 line requires, written against the merged tree rather than a guess.

I confirm that I have read and agree to the Contributors License Agreement.

## Checklists

https://claude.ai/code/session_017oeRgVgXqSeKKb1VacoGyB


scosman:
Closing in favour of **#1662**, which is this same project rebuilt on a correct base.

This branch was cut from `scosman/eval-api-improvements`, which is itself the still-draft PR #1517 (`leonard/kil-686-eval-job`) plus two split commits. So it inherited 52 commits of that in-flight work — including files carrying explicit `# TODO (merge blocker — do not merge toward main until resolved)` markers on `judge_feedback_batch`. None of that may reach a main-destined branch, and retargeting alone couldn't remove it.

#1662 is built commit-by-commit on `scosman/evals_v2` and contains only this project's work:

| | #1661 | #1662 |
|---|---|---|
| Files changed | 460 | 50 |
| Commits | 52 inherited + ours | 20, all ours |
| Files under `studio_server/jobs/**` | present | 0 |
| `judge_feedback_batch` files | present | 0 |
| CodeRabbit review | skipped (460 files, 160 over its limit) | within limit |

One deliberate difference: phase 5 (the jobs layer) is not in #1662. Its code targets `jobs/workers/eval.py` and `jobs/api.py`, which belong to #1517 and don't exist on `scosman/evals_v2`. It ships separately — the plan and re-landing instructions travel with #1662 as `phase_plans/phase_5.md`, and the code stays on this branch as `369a32ef8`.

This branch is kept as the record; please don't delete it, since #1662 references `369a32ef8` for the jobs-layer work.

---

PR #1667, opened 2026-08-12
Spec builder: show a notice when full-trace evals leave the Kiln Pro flow

chiang-daniel:
Toggling 'Evaluate Complete Agent History' in the spec builder silently switched the form from the Kiln Pro guided flow to manual creation: the submit button changed from 'Create with Kiln Pro' to 'Create Eval' and the sample selector disappeared with no explanation. Reported by a design partner who only noticed after creating their eval manually.

This adds a visible warning above the submit button when Kiln Pro is available and the toggle moves the form off the Pro path: 'Kiln Pro does not support evaluating complete agent history yet. This eval will be created manually.' It uses the existing Warning component in the same outline style as other in-form advisories, and sits outside the collapsible Advanced Options section so it stays visible after the section is closed. No gating or creation logic changed.

Includes component tests covering the notice showing/hiding, the submit label flip, the sample selector unmounting, and a full toggle on/off cycle. Verified in the running app.

Fixes KIL-786.

image


chiang-daniel:
This is just a quick fix for better message.

leonardmq (approved):

PR #1669, opened 2026-08-13
Remove the tool call limit on agent runs; add stuck detection

chiang-daniel:
Agent runs were capped at a hard-coded 30 tool calls per turn, which blocked legitimate tool-heavy agents. A design partner's discovery agents hit the cap during normal retries. The limit dated to the original tool-calling loop and was never tuned or documented.

This removes the cap entirely. Agent runs now make as many tool calls as the task needs. In its place, a stuck detector watches for the one provably wasteful pattern: the model repeating the exact same tool call and getting the same result. After 3 identical rounds it injects a warning the model can see and act on. After 5 it stops the run, but only when the repeated results are errors, so an agent polling a status endpoint is never killed. Distinct calls in any volume never trigger it. The thresholds and warn-then-stop shape follow shipped implementations in Gemini CLI, Cline, and OpenHands.

Supporting changes: context window exhaustion now surfaces as a clear message instead of 'An unexpected error occurred', unmapped provider 400s and content policy blocks get accurate messages, error mapping now works on the streaming path (it previously only applied to blocking runs), per-round message copying is no longer quadratic so long runs stay fast, and the injected warning is marked kiln_injected in saved traces so trace consumers can distinguish it from real user input.

Tested at three layers: unit tests on the detector and copier, mocked-model tests through both loop paths, and a live paid regression test where gpt-4.1-mini drives 35 chained tool calls past the old cap (marked paid, runs only with --runpaid, verified twice).

Total spend on a runaway agent remains bounded by the provider account. Docs for that guidance are drafted and need a home on docs.kiln.tech.

Fixes KIL-783.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
@@ -40,6 +40,10 @@ from kiln_ai.adapters.model_adapters.base_adapter import (
     Usage,
 )
 from kiln_ai.adapters.model_adapters.litellm_config import LiteLlmConfig
+from kiln_ai.adapters.model_adapters.tool_loop_guards import (
+    IncrementalMessageCopier,
+    RepeatedToolCallDetector,
+)
 from kiln_ai.datamodel.datamodel_enums import InputType
 from kiln_ai.datamodel.json_schema import (
     close_object_schemas,
@@ -66,7 +70,6 @@ from kiln_ai.utils.open_ai_types import (
 )
 
 MAX_CALLS_PER_TURN = 10
-MAX_TOOL_CALLS_PER_TURN = 30
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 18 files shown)


libs/core/kiln_ai/adapters/model_adapters/litellm_adapter.py
```diff
@@ -66,7 +70,6 @@
 )
 
 MAX_CALLS_PER_TURN = 10
-MAX_TOOL_CALLS_PER_TURN = 30
```
chiang-daniel, line 69: This is the bulk of the change.

libs/core/kiln_ai/adapters/model_adapters/tool_loop_guards.py
```diff
@@ -0,0 +1,279 @@
...
+
+        if self._count >= STUCK_NUDGE_THRESHOLD and not self._nudged:
+            self._nudged = True
+            # A plain user message so every provider surfaces it to the model.
+            # `kiln_injected` marks it as ours: stripped before the provider call, kept
+            # in the trace so the run can tell it apart from real user input.
+            return ChatCompletionUserMessageParamWrapper(
+                role="user",
+                content=(
+                    f"You have called {tool_names} with the same arguments "
```
leonardmq, line 134: This is orchestration-level guardrails, which should not be baked into the adapter because it is opinionated and happening in the blackbox. This would obfuscate a model being really bad, and would be weird to bake into evals and so on; also has impact on costs and so on so user should at least have some visibility into which guardrails are on or off. Which guardrails work best would also be something that auto-optimizer could come up on its own ideally

leonardmq (changes requested):

chiang-daniel:
Closing per team decision: shipping the minimal cap raise in #1678 instead of full limit removal with stuck detection. The research and implementation here stay on the branch for reference if the unbounded direction is revisited. Note for anyone reading later: this PR also contained fixes unrelated to the limit itself (context window exhaustion and other provider 400s currently surface as 'An unexpected error occurred', and the streaming path bypasses error mapping entirely). Those gaps remain on main and are worth extracting separately.

PR #1708, opened 2026-08-19
get_tool_results: match OpenAI-style role "tool" messages

chiang-daniel:
KilnEvalHelpers.get_tool_results filtered the trace for role "tool_result", but Kiln traces store tool results as OpenAI-style role "tool" messages. On any real trace the helper returned an empty list, so a code judge using it saw tool calls with no results and scored zero.

Two-line fix: the filter now matches role "tool" alongside the tool_result shapes, with a test pinning the OpenAI-format case (tool result recovered by tool_call_id from a realistic trace).

Repro: create a Code judge whose scorer calls KilnEvalHelpers.get_tool_results(trace), run Test Judge against any task run whose trace contains tool calls. Before: always []. After: the tool messages.

This is the minimal release-line fix. The fuller scorer-helper set in #1592 contains the same correction and merges cleanly over this.

### DIFF
--- a/libs/core/kiln_ai/adapters/eval/eval_helpers.py
+++ b/libs/core/kiln_ai/adapters/eval/eval_helpers.py
@@ -66,13 +66,18 @@ class KilnEvalHelpers:
 
     @staticmethod
     def get_tool_results(trace: list[dict[str, Any]] | None) -> list[dict[str, Any]]:
-        """Return all tool-result entries from a trace."""
+        """Return all tool-result entries from a trace.
+
+        Kiln traces store tool results as OpenAI-style ``role: "tool"``
+        messages; ``tool_result``-shaped entries are also matched.
+        """
         if not trace:
             return []
         return [
             entry
             for entry in trace
-            if entry.get("role") == "tool_result" or entry.get("type") == "tool_result"
+            if entry.get("role") in ("tool", "tool_result")
+            or entry.get("type") == "tool_result"
         ]
 
     # -- Tool-call matching -------------------------------------------------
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/adapters/eval/eval_helpers.py
```diff
@@ -66,13 +66,18 @@ def get_assistant_messages(trace: list[dict[str, Any]] | None) -> list[str]:
...
+        Kiln traces store tool results as OpenAI-style ``role: "tool"``
+        messages; ``tool_result``-shaped entries are also matched.
+        """
         if not trace:
             return []
         return [
             entry
             for entry in trace
-            if entry.get("role") == "tool_result" or entry.get("type") == "tool_result"
+            if entry.get("role") in ("tool", "tool_result")
```
tawnymanticore, line 79: nit: I think you can safely remove `tool_result` from the list here its not a role that can be matched on, its called `tool`

tawnymanticore (approved):

libs/core/kiln_ai/adapters/eval/eval_helpers.py
chiang-daniel: Agreed it's dead — nothing produces role "tool_result" (the only "tool_result" strings in core are the sandbox bridge's IPC message types, not trace entries). Left it in this minimal fix to keep the release-line diff two lines; removing it in #1592, which carries the same list in the fuller helper set.

chiang-daniel: Removed in Kiln-AI/Kiln#1592 as of 67a1c4a75 — role match is `== "tool"` only, the `type` defensive match stays, and the test now pins that a role-shaped "tool_result" entry does not match.

PR #1719, opened 2026-08-21
[REVIEW ONLY — do not merge] The conflict resolution for #1718, on its own

leonardmq:
Not a real PR. It exists so the resolution in **#1718** can be read in the browser.

The base branch is the merge **`git` produced on its own** — 63 files with conflict markers still in them, committed unresolved on purpose. The head is the same tree #1718 ships. So the Files changed tab here is exactly what I decided, with main's 424 upstream commits subtracted out: **71 files, +1,097 / −5,364**.

It is `git show --remerge-diff 15d03b069` rendered as a PR.

## Reading it

Conflict markers on the left side of every hunk are the base branch being unresolved — that is the point, not damage.

Three shapes account for nearly all of it:

- **Import unions** — both sides added to the same block. Mechanical.
- **Deletions of agi's superseded copy** — main's version of the same thing won, so agi's is removed. This is most of the −5,364: the pre-splits read layer (`eval_input_ids_in_filter`, the old `SplitItemResolver`), `_build_trace_reuse_index` and its helpers, the duplicated code-eval examples, and the tests that pinned the removed behaviour.
- **The two ports**, which are the part worth real attention:
  - `eval_runner.py` — the multi-turn drive lane onto `TraceIndex`, plus the `TraceKey` variant in `trace_index.py` / `task_run.py`.
  - `eval_api.py` — the split-query read layer onto `Eval.splits`, and execution identity onto `scored_run_id`.

## Behaviour changes to check

Called out in #1718's description too, repeated here because this is where they are visible:

- `EvalTaskInput.from_trace` accepts multi-turn synthetic sources; main refused them.
- Skip reasons for multi-turn: `missing_drive_config` and `missing_trace` where main said `incompatible_input_shape`.
- The copilot dataset splitter keeps agi's 50/25/25 over main's `MIN_GOLDEN_EXAMPLES` top-up.
- Trace reuse across racing writers is "a stable pick", not "lowest id".
- An EvalInput-backed eval shows an explanation instead of a dead-end "Add Eval Data" button.

Close this and delete both branches once #1718 is reviewed.


leonardmq:
Hard to review this way too — replacing it with a written summary of the resolution. Closing; branches deleted.

PR #1727, opened 2026-08-24
fix: restrict call_kiln_api to API paths

leonardmq:
## What does this PR do?

`call_kiln_api` builds its URL from the server root, and that server serves more than the API: the SvelteKit web app is mounted at `/`, and FastAPI's own `/docs`, `/redoc` and `/openapi.json` answer too.

`url_path` was validated for a leading slash and nothing else. So a mistyped path did not come back as a JSON 404. It came back as a web page, and the model got thousands of tokens of HTML in its context instead of an error it could act on.

`url_path` must now start with `/api/`, or be exactly `/ping`.

That pair is not a new rule. Every production route already lives under `/api/` apart from `/ping`, and `test_all_routes_have_tags` already uses exactly that pair as its definition of an API route.

## Rejected, with a reason the model can act on

Where the fix is obvious, the error names it:

```
url_path must start with '/api/', or be exactly '/ping'. Got: '/projects/123'.
Did you mean '/api/projects/123'? Correct paths are in the endpoint documentation.
```

The exceptions are listed from `ALLOWED_EXACT_PATHS` rather than hardcoded, so the message cannot drift from the guard. Stating the rule alone would read as "`/ping` is impossible", which would stop a caller retrying the one path outside `/api/` that does work.

Where the fix is not obvious, the error does not guess. `/api/openapi.json` is no more callable than `/openapi.json`, so a doc or asset route gets the rule and a pointer to the endpoint documentation, not a misleading suggestion. A dot in the final segment is read as a filename and suppresses the hint for the same reason.

A near miss on an exact allowed path is corrected to that path: `/ping/` is told `Did you mean '/ping'?`, not `/api/ping/`, which would just be a second wrong call. A genuinely unknown path like `/pings` is not a near miss and keeps the ordinary prefix hint.

## Dot segments

A prefix check on the string we are handed is not a check on the path that gets sent. httpx resolves dot segments, so a path starting with `/api/` could leave as something else:

```
/api/../docs                -> /docs
/api/a/../../openapi.json   -> /openapi.json
```

The second is the exact payload this guard exists to stop. Percent-encoded forms (`%2e%2e`, `%2E%2E`) are left encoded by httpx, so whether they escape depends on who decodes them downstream; they are rejected too, as is a dot segment hidden behind an encoded slash (`%2e%2e%2f`).

Segments are decoded one at a time and split again on any slash the decoding revealed, rather than decoding the whole path at once. So an encoded slash is not a separator in its own right — httpx keeps those encoded, and `/api/x%2Fy` stays one segment under `/api/`.

Dot segments get their own error message. Telling the caller a path "must start with `/api/`" reads as nonsense when it plainly does.

A test pins httpx's normalisation, so the reason for the rule is checked rather than just the rule.

## Why this is worth a guard rather than a doc line

The instruction already exists in agent prompts, and agents still get it wrong. In one internal optimisation loop, 34 recorded traces have a `<!doctype html>` body sitting in an agent's context, and 21 calls fetched `/openapi.json` despite a prompt line that forbids exactly that.

Two things caused it: a dropped `/api` prefix, and a wrong path segment. Both are ordinary model slips, and neither is fixable by asking more firmly.

## Notes

- The rejection happens before the HTTP call, so no page is fetched at all.
- The `url_path` parameter description is untouched. A caller asking for an endpoint is not wondering about web app routes, and the guard states the rule when a call breaks it. The tool file diff is purely additive.
- The check ignores any query string when deciding, so a path with both problems reports the prefix, not the query string.
- No response size cap. That would truncate the file downloads this tool is meant to serve, such as document and dataset downloads.
- Existing tests used `/test` as a path. They now use `/api/test`.

## The same guard, twice

`api_mcp` in the experiments repo holds a duplicated copy of this tool — it imports only the `KilnTool` base class and repeats the body, so it never picked any of this up. It had the same hole, including the dot-segment one, and Kiln-AI/experiments#28 ports this guard across verbatim.

That PR is where the field evidence above comes from. It is independent of this one, so the two can merge in either order; the copy is what should go, eventually.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -275,9 +275,13 @@ def _disallowed_path_message(url_path: str) -> str:
         and not looks_like_a_file
     )
     hint = f" Did you mean '/api{path_only}'?" if suggest else ""
+    # Name the exceptions. The rule alone reads as "/ping is impossible",
+    # which would stop a caller retrying the one path that does work.
+    exceptions = ", ".join(f"'{path}'" for path in sorted(ALLOWED_EXACT_PATHS))
     return (
-        f"url_path must start with '{API_PATH_PREFIX}'. Got: '{url_path}'."
-        f"{hint} Correct paths are in the endpoint documentation."
+        f"url_path must start with '{API_PATH_PREFIX}', or be exactly "
+        f"{exceptions}. Got: '{url_path}'.{hint} "
+        "Correct paths are in the endpoint documentation."
     )
 
 
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -265,16 +265,23 @@ def _disallowed_path_message(url_path: str) -> str:
             "endpoint documentation."
         )
 
-    # Suggest the prefixed form only where it could plausibly be the fix. A
-    # dot in the final segment means a filename, not an endpoint, and
-    # "/api/openapi.json" is no more callable than "/openapi.json".
-    looks_like_a_file = "." in path_only.rsplit("/", 1)[-1]
-    suggest = (
-        not path_only.startswith("/api")
-        and path_only not in _NON_API_PATHS
-        and not looks_like_a_file
-    )
-    hint = f" Did you mean '/api{path_only}'?" if suggest else ""
+    # A near miss on an exact allowed path — a trailing slash, say — is
+    # corrected to that path. Prefixing it instead would name a second wrong
+    # call: '/api/ping/' is no more a route than '/ping/' is.
+    near_miss = path_only.rstrip("/")
+    if near_miss in ALLOWED_EXACT_PATHS:
+        hint = f" Did you mean '{near_miss}'?"
+    else:
+        # Suggest the prefixed form only where it could plausibly be the fix.
+        # A dot in the final segment means a filename, not an endpoint, and
+        # "/api/openapi.json" is no more callable than "/openapi.json".
+        looks_like_a_file = "." in path_only.rsplit("/", 1)[-1]
+        suggest = (
+            not path_only.startswith("/api")
+            and path_only not in _NON_API_PATHS
+            and not looks_like_a_file
+        )
+        hint = f" Did you mean '/api{path_only}'?" if suggest else ""
     # Name the exceptions. The rule alone reads as "/ping is impossible",
     # which would stop a caller retrying the one path that does work.
     exceptions = ", ".join(f"'{path}'" for path in sorted(ALLOWED_EXACT_PATHS))
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -2,6 +2,7 @@
 
 import json
 from typing import Any
+from urllib.parse import unquote
 
 import httpx
 import jq
@@ -211,18 +212,52 @@ For SSE endpoints (text/event-stream), the tool consumes the stream until it clo
         return ToolCallResult(output=json.dumps(result, ensure_ascii=False))
 
 
-def _is_allowed_path(url_path: str) -> bool:
-    """True when url_path addresses the Kiln REST API rather than the web app.
+def _path_without_query(url_path: str) -> str:
+    """The path portion alone.
 
-    The query string is ignored here: a url_path carrying one is rejected by
-    its own check, and stripping it first keeps this error about the path.
+    The query string is ignored by the path checks: a url_path carrying one is
+    rejected by its own check, and stripping it first keeps those errors about
+    the path.
     """
-    path_only = url_path.split("?", 1)[0].split("#", 1)[0]
+    return url_path.split("?", 1)[0].split("#", 1)[0]
+
+
+def _has_dot_segment(path_only: str) -> bool:
+    """True when any segment is '.' or '..', encoded or not.
+
+    A prefix check alone is not enough, because the string we validate is not
+    the path that gets sent. httpx resolves dot segments, so '/api/../docs'
+    leaves as '/docs' and '/api/a/../../openapi.json' leaves as
+    '/openapi.json' — both past a guard that only looked at the '/api/' at the
+    front.
+
+    Segments are decoded one at a time rather than decoding the whole path,
+    so an encoded '%2e%2e' is caught while an encoded slash inside a single
+    segment does not split into new ones (httpx leaves those encoded too).
+    """
+    return any(unquote(segment) in {".", ".."} for segment in path_only.split("/"))
+
+
+def _is_allowed_path(url_path: str) -> bool:
+    """True when url_path addresses the Kiln REST API rather than the web app."""
+    path_only = _path_without_query(url_path)
+    if _has_dot_segment(path_only):
+        return False
     return path_only in ALLOWED_EXACT_PATHS or path_only.startswith(API_PATH_PREFIX)
 
 
 def _disallowed_path_message(url_path: str) -> str:
-    path_only = url_path.split("?", 1)[0].split("#", 1)[0]
+    path_only = _path_without_query(url_path)
+    if _has_dot_segment(path_only):
+        # A separate message: "must start with /api/" reads as nonsense for a
+        # path that plainly does start with it.
+        return (
+            "url_path must not contain '.' or '..' path segments, encoded or "
+            f"not. Got: '{url_path}'. Such a path is resolved before it is "
+            "sent, so it can land outside the API — '/api/../docs' is sent as "
+            "'/docs'. Write the full path instead. Correct paths are in the "
+            "endpoint documentation."
+        )
     last_segment = path_only.rsplit("/", 1)[-1]
     # Suggest the prefixed form only where it could plausibly be the fix.
     # Offering "/api/openapi.json" would send the model somewhere just as wrong.
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -65,7 +65,7 @@ For SSE endpoints (text/event-stream), the tool consumes the stream until it clo
                 },
                 "url_path": {
                     "type": "string",
-                    "description": "API path, starting with '/api/'. No query string — pass query args via query_params. Correct paths are in the endpoint documentation.",
+                    "description": "API path with no query string — pass query args via query_params. Correct paths are in the endpoint documentation.",
                 },
                 "query_params": {
                     "type": "object",
--- a/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
+++ b/libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
@@ -65,7 +65,7 @@ For SSE endpoints (text/event-stream), the tool consumes the stream until it clo
                 },
                 "url_path": {
                     "type": "string",
-                    "description": "API path, starting with '/api/' (only '/ping' is allowed outside it). No query string — pass query args via query_params. Web app routes and doc routes such as '/openapi.json' or '/docs' are rejected. Correct paths are in the endpoint documentation.",
+                    "description": "API path, starting with '/api/'. No query string — pass query args via query_params. Correct paths are in the endpoint documentation.",
                 },
                 "query_params": {
                     "type": "object",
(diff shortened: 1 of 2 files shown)


libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -199,6 +211,37 @@ async def run(  # type: ignore[override]
...
 
 
+def _is_allowed_path(url_path: str) -> bool:
+    """True when url_path addresses the Kiln REST API rather than the web app.
+
+    The query string is ignored here: a url_path carrying one is rejected by
+    its own check, and stripping it first keeps this error about the path.
+    """
+    path_only = url_path.split("?", 1)[0].split("#", 1)[0]
+    return path_only in ALLOWED_EXACT_PATHS or path_only.startswith(API_PATH_PREFIX)
```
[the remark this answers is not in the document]
leonardmq: Good catch, and it was worse than described. Fixed in b154d7f3a.

I verified against real httpx rather than the simulation (which used `ALLOWED_EXACT_PATHS = {"/health", "/docs"}`; the actual set is `{"/ping"}`, though that does not change the finding):

```
/api/../docs                -> /docs
/api/a/../../openapi.json   -> /openapi.json
/api/%2e%2e/docs            -> sent as-is, still encoded
```

The second one is the exact payload this guard exists to stop — the full OpenAPI schema is the largest context dump we recorded in the original incident. So the prefix check was validating the string it was handed, not the path that gets sent.

Fix: reject any segment that is `.` or `..`, before the prefix check. Segments are decoded individually rather than decoding the whole path, so `%2e%2e` is caught while an encoded slash inside one segment does not split into new segments — httpx keeps those encoded, and such a path stays under `/api/`.

Dot segments get their own error message. The existing "must start with /api/" reads as nonsense for a path that plainly does.

Both forms are in `TestDotSegments`, along with `/api/./docs`, `/api/%2E%2E/docs`, and `/api/a/../../openapi.json`. I also added `test_httpx_really_would_have_sent_these_elsewhere`, which pins httpx normalisation itself — if that behaviour ever changes, the test fails and the rule gets reconsidered rather than cargo-culted. Two negative cases guard against over-rejecting: `/api/%2Foutside` and `/api/files/report..txt` both still pass.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -55,7 +65,7 @@ def _build_parameters_schema() -> dict[str, Any]:
                 },
                 "url_path": {
                     "type": "string",
-                    "description": "API path with no query string — pass query args via query_params. Correct paths are in the endpoint documentation.",
+                    "description": "API path, starting with '/api/' (only '/ping' is allowed outside it). No query string — pass query args via query_params. Web app routes and doc routes such as '/openapi.json' or '/docs' are rejected. Correct paths are in the endpoint documentation.",
```
leonardmq: Remove the part about `Web app routes and doc routes such as '/openapi.json' or '/docs' are rejected` - tool caller is not even wondering about that at all. Remove mention of `/ping` as well - caller probably doesn't care

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -199,6 +212,71 @@ async def run(  # type: ignore[override]
...
+    # Suggest the prefixed form only where it could plausibly be the fix.
+    # Offering "/api/openapi.json" would send the model somewhere just as wrong.
+    suggest = (
+        not path_only.startswith("/api")
+        and path_only not in _NON_API_PATHS
+        and "." not in last_segment
+    )
+    hint = f" Did you mean '/api{path_only}'?" if suggest else ""
+    return (
+        f"url_path must start with '{API_PATH_PREFIX}', or be exactly "
```
leonardmq, line 289: Simplify this - still no need to mention `openapi.json` or `/docs`

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -199,6 +212,71 @@ async def run(  # type: ignore[override]
...
+            "'/docs'. Write the full path instead. Correct paths are in the "
+            "endpoint documentation."
+        )
+    last_segment = path_only.rsplit("/", 1)[-1]
+    # Suggest the prefixed form only where it could plausibly be the fix.
+    # Offering "/api/openapi.json" would send the model somewhere just as wrong.
+    suggest = (
+        not path_only.startswith("/api")
+        and path_only not in _NON_API_PATHS
+        and "." not in last_segment
```
leonardmq: Why the `.`

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Done in 9d2739c4b. Now just:

> API path, starting with '/api/'. No query string — pass query args via query_params. Correct paths are in the endpoint documentation.

Both mentions dropped. `/ping` is still allowed by the tool, just not advertised — it has its own endpoint doc, so a caller that wants it finds it there rather than here.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Done in 9d2739c4b. The prefix error is now the rule, the suggestion, and where to look:

```
url_path must start with '/api/'. Got: '/projects/123'. Did you mean '/api/projects/123'? Correct paths are in the endpoint documentation.
```

The dot-segment error got the same treatment. Added `test_prefix_error_names_only_the_rule_and_the_fix`, which asserts neither string reappears.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: It was a file-extension test, but you had to reverse-engineer that from the `.`, which is a fair complaint. Renamed in 9d2739c4b:

```python
# Suggest the prefixed form only where it could plausibly be the fix. A
# dot in the final segment means a filename, not an endpoint, and
# "/api/openapi.json" is no more callable than "/openapi.json".
looks_like_a_file = "." in path_only.rsplit("/", 1)[-1]
```

It only gates the "Did you mean" hint, never the accept/reject decision. Without it, `/openapi.json` and `/_app/immutable/start.js` get told to retry at `/api/openapi.json`, which is just as dead. `_NON_API_PATHS` covers the extensionless ones (`/`, `/docs`, `/redoc`, `/scalar`) that this test would miss.

Covered by `test_no_suggestion_for_something_that_looks_like_a_file`.

If you would rather drop the suggestion entirely, both this and `_NON_API_PATHS` go with it and the message becomes one line. I kept it because a dropped `/api` prefix is the most common form of this mistake, and naming the fix is what stops the retry loop.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -199,6 +212,71 @@ async def run(  # type: ignore[override]
...
+    the path that gets sent. httpx resolves dot segments, so '/api/../docs'
+    leaves as '/docs' and '/api/a/../../openapi.json' leaves as
+    '/openapi.json' — both past a guard that only looked at the '/api/' at the
+    front.
+
+    Segments are decoded one at a time rather than decoding the whole path,
+    so an encoded '%2e%2e' is caught while an encoded slash inside a single
+    segment does not split into new ones (httpx leaves those encoded too).
+    """
+    return any(unquote(segment) in {".", ".."} for segment in path_only.split("/"))
```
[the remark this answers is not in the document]
leonardmq: Hardened in 9d2739c4b, but the premise did not hold here, so recording what I measured.

I ran these against the actual Kiln server rather than reasoning about it:

```
url_path                     sent as                      status  content-type       bytes
/api/%2e%2e%2fdocs           /api/%2e%2e%2fdocs           404     application/json      23
/api/%2e%2e/docs             /api/%2e%2e/docs             404     application/json      23
/api/../docs                 /docs                        200     text/html            935
/openapi.json                /openapi.json                200     application/json  556716
```

httpx leaves `%2f` encoded, and uvicorn/starlette does not decode-and-normalize it, so `/api/%2e%2e%2fdocs` already returned a 23-byte JSON 404 — the outcome this guard exists to produce. The literal form on row 3 is the one that actually escaped, and that was the previous commit. This tool only ever talks to the local Kiln server (base URL comes from config as `http://127.0.0.1:<port>`), so there is no proxy in the path to decode it differently.

I did not take the prescribed fix. Decoding the whole path before splitting makes `%2F` a separator, which over-rejects a legitimate encoded slash inside a single path segment — `/api/x%2Fy` is one segment to httpx and stays under `/api/`.

Instead each segment is decoded and then split again on any slash the decoding revealed:

```python
return any(
    part in {".", ".."}
    for segment in path_only.split("/")
    for part in unquote(segment).split("/")
)
```

`%2e%2e%2f` is rejected; `/api/x%2Fy` and `/api/files/report..txt` still pass. Both encoded-slash forms are in `TestDotSegments`, and the two negative cases guard the over-rejection.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Gone entirely in 0572395fc. Reverted to the wording on `main`:

> API path with no query string — pass query args via query_params. Correct paths are in the endpoint documentation.

So this PR no longer touches the parameter schema at all, and the tool file diff is now purely additive — 82 insertions, no deletions.

The guard still enforces the rule; the error states it only when a call breaks it.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
leonardmq: Good catch — I over-trimmed this one. Added back in 62aa81f8e:

```
url_path must start with '/api/', or be exactly '/ping'. Got: '/projects/123'. Did you mean '/api/projects/123'? Correct paths are in the endpoint documentation.
```

Your reasoning holds: the prefix rule on its own reads as "/ping is impossible", so a caller that legitimately wants it would not retry.

The exceptions are rendered from `ALLOWED_EXACT_PATHS` rather than hardcoded into the string, so the message cannot drift from the guard if that set ever changes.

The `/openapi.json` and `/docs` lecture stays gone, and the parameter description still says nothing about any of this. Added `test_prefix_error_names_the_allowed_exceptions`, which iterates the set, so a future trim of this message fails a test rather than quietly reintroducing the problem.

libs/core/kiln_ai/tools/built_in_tools/kiln_api_call_tool.py
```diff
@@ -199,6 +212,79 @@ async def run(  # type: ignore[override]
...
+        and not looks_like_a_file
+    )
+    hint = f" Did you mean '/api{path_only}'?" if suggest else ""
+    # Name the exceptions. The rule alone reads as "/ping is impossible",
+    # which would stop a caller retrying the one path that does work.
+    exceptions = ", ".join(f"'{path}'" for path in sorted(ALLOWED_EXACT_PATHS))
+    return (
+        f"url_path must start with '{API_PATH_PREFIX}', or be exactly "
+        f"{exceptions}. Got: '{url_path}'.{hint} "
+        "Correct paths are in the endpoint documentation."
```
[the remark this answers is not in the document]
leonardmq: Confirmed and fixed in 44a148763. Reproduced before touching anything:

```
/ping/   -> Did you mean '/api/ping/'?
/ping//  -> Did you mean '/api/ping//'?
```

Both name a second wrong call, which defeats the point of suggesting at all.

A path that reaches an entry in `ALLOWED_EXACT_PATHS` once trailing slashes are stripped now gets that path back:

```
/ping/        -> Did you mean '/ping'?
/ping//       -> Did you mean '/ping'?
/pings        -> Did you mean '/api/pings'?
/projects/123 -> Did you mean '/api/projects/123'?
```

Keyed off `ALLOWED_EXACT_PATHS` rather than hardcoding `/ping`, so it follows the set.

**One part I did not take: excluding descendants.** `/ping/status` still gets `/api/ping/status`. It is not a near miss — no amount of slash-trimming reaches `/ping` — it is simply an unknown path, and the prefix guess is exactly as good there as it is for `/foo/bar`. Suppressing the hint for anything under `/ping/` would remove a useful suggestion to avoid a wrong one that is no more wrong than the general case.

Regression tests: `test_near_miss_on_an_exact_path_is_corrected_to_that_path` (parametrised over `/ping/` and `/ping//`, asserting `/api/ping` does *not* appear) and `test_a_genuinely_unknown_path_still_gets_the_prefix_hint`, which pins the boundary so a future tightening cannot silently swallow the normal case.

chiang-daniel (approved):

tawnymanticore (approved):

PR #1732, opened 2026-08-26
Skill bundles: atomic create with resource files, full-directory clone, resource APIs

leonardmq:
## What does this PR do?

Adds an API for creating full skills — SKILL.md plus `references/` and `assets/` files (nested folders, binary assets) — and makes clone copy the whole bundle.

### Architecture

- **Atomic, install-once bundles.** New `kiln_ai/datamodel/skill_bundle.py`: a bundle is staged in a hidden dir outside `skills/` (same filesystem), fsynced, then committed with a single `os.rename` — a skill either doesn't exist or exists fully formed, and is never mutated after install. Iteration = clone; there are deliberately no file-edit endpoints. Names may repeat across skills (coexisting versions differ by id); identity is always the id.
- **Validation accumulates every failure** (path safety, Windows portability, case/normalization collisions, UTF-8 references, size/count caps) so a caller fixes the bundle once, not one round trip per defect.
- **API** (`skill_api.py`): `POST /skills` accepts optional `files` (utf-8 or base64, MIME-wrapped base64 tolerated) · `POST /skills/{id}/clone` copies the full directory server-side (streamed, hand-added files and empty scaffold dirs included) · `GET /skills/{id}/resources` lists files · `GET /skills/{id}/resource_content` reads one (binary-safe, 10MB cap). All `ALLOW_AGENT`, annotations regenerated.
- **Runtime reuse:** `SkillTool` resource loading delegates to `Skill.read_resource_text` — one implementation of prefix routing and containment.
- **Hygiene:** crash-orphaned staging debris is swept at desktop startup; `.skill_staging/` is git-sync-ignored; symlinked staging roots are rejected (create) and never followed (sweep).

### UI behavior

1. **Skill created via API with assets/references:** the UI doesn't break — list and detail pages only render name/description/body, so resource files are simply not shown (the "open folder" button is how users see them today).
2. **Clone via UI:** the clone form now calls the new clone endpoint, so `references/` and `assets/` are copied (previously silently dropped). Name/description/body edits in the form apply to the clone.

### Bug fixes baked into this PR

- **Repo-wide error-handler crash:** a lone UTF-16 surrogate in rejected request input crashed `kiln_server`'s error handlers while UTF-8-encoding the response, turning 4xx responses into 500s on any endpoint. Both the validation handler and the shared `HTTPException` handler now sanitize recursively (nested dict/list details included).
- **UI clone dropped resource files:** cloning a skill through the web UI silently lost everything under `references/` and `assets/`.

### Manual testing

Stress-tested end-to-end against a live dev server (seeded sandbox project), API + real browser:

- Create SKILL.md-only skill; create a full bundle (nested `references/api/deep/`, unicode filename `café-menu.md`, a real PNG, a CSV); verify every file round-trips byte-identically through the resource endpoints.
- Unicode/CJK/emoji in body and descriptions — rendered correctly on list and detail pages.
- Three coexisting `ticket-api` versions (one archived) — list page and detail pages all render correctly.
- Near-cap bundle: 51 files / ~1MB installs instantly; no staging debris left behind.
- Error paths: one 422 accumulating bad path + traversal attempt + invalid base64; oversized file 422; traversal read 422; missing resource 404.
- Clone via the UI with an edited body: verified on disk that all 6 resources copied (nested dirs, unicode filename, byte-identical PNG), SKILL.md regenerated with the edited body, fresh `skill.kiln`.
- Create via the UI with a duplicate name (succeeds, new id); archive/unarchive from the detail page.
- Path traversal battery (API level): `..` traversal, absolute-path join, encoded slashes, symlinked files and directories pointing outside the skill — all rejected on read and write paths.

<b>Screenshots</b> (from the stress test; hosted on the <code>claude/pr-1732-screenshots</code> assets branch — never merge it, safe to delete after this PR closes)

**Skills list — three coexisting `ticket-api` versions, one archived:**

skills list

**Detail page of an API-created bundle skill (unicode body; resource files invisible by design):**

skill detail

**Clone form, prefilled from the bundle skill:**

clone form

**Final state after the UI create/clone/archive flows:**

final list

### Provenance (now wired in)

`KilnArtifactProvenance` (#1565) has since landed on the base branch, and merging it in was where all three conflicts lived — so the follow-up is done here rather than later:

- `POST /skills` takes optional `provenance` alongside `files`. Lineage is validated **before** anything is staged, so a bad `derived_from_ids` returns 400 with no half-installed bundle and no staging debris.
- `POST /skills/{id}/clone` takes the same optional `provenance`. It is **not** stamped automatically and the source's provenance is never copied — a clone's provenance describes the clone, so the caller supplies it (including the `derived_from_ids` link back to the source).
- The clone form keeps calling the clone endpoint (so `references/` and `assets/` are still copied) while stamping the lineage the base branch introduced: `{origin: "human", derived_from_ids: [source_id]}` on clone, `{origin: "human"}` on a fresh create.

## Related Issues

Groundwork for skill authoring by agents (auto research).

## Checklists

- [x] Tests have been run locally and passed (`checks.sh` green)
- [x] New tests have been added to any work in /lib (~100 new tests)

## Related PRs

- #1565 — Artifact provenance (shared `KilnArtifactProvenance` submodel). Merged into the base branch; create/clone here now stamp lineage (see above).

https://claude.ai/code/session_016hr8tm9wQRGYZyTPpcy1kr


libs/core/kiln_ai/datamodel/skill.py
```diff
@@ -112,6 +124,86 @@ def _read_resource(self, base_dir: Path, relative_path: str) -> str:
...
 
+    def _resource_base_dir(self, prefixed_path: str) -> tuple[Path, str]:
+        """Split a 'references/…' or 'assets/…' path into (base_dir, relative_path)."""
+        prefix, sep, relative_path = prefixed_path.partition("/")
+        if prefix not in RESOURCE_DIR_NAMES or not sep or not relative_path:
+            raise ValueError(
+                f"Resource path must start with 'references/' or 'assets/' followed by a filename: {prefixed_path!r}"
+            )
+        base_dir = (
+            self.references_dir() if prefix == "references" else self.assets_dir()
```
leonardmq: Do not fall back on the `assets` dir by default - which dir the call refers to must be explicit and precise. If we get an unexpected value here, we should throw explicitly. `assets` is not a fallback. Would cause bizarre behaviors when you factor in backward compatibility between Skills on the FS and Kiln desktop app version - e.g. we add `scripts/` later, but all the calls default to `assets/` and not found.

libs/core/kiln_ai/datamodel/skill.py
```diff
@@ -112,6 +124,86 @@ def _read_resource(self, base_dir: Path, relative_path: str) -> str:
...
+        prefixed_path must start with 'references/' or 'assets/'. Raises ValueError
+        for invalid paths or path traversal, FileNotFoundError if missing, and
+        ResourceTooLargeError if max_bytes is set and the file exceeds it (checked
+        while reading, so a file growing mid-read cannot slip past the cap).
+        """
+        base_dir, relative_path = self._resource_base_dir(prefixed_path)
+        resolved = self._resolve_resource(base_dir, relative_path)
+        try:
+            if max_bytes is None:
+                return resolved.read_bytes()
```
leonardmq, line 181: How about read text; if a Skill folder contains an arbitrary payload, it could be malicious, possibly safer to read text? And throw if not text - e.g. image.

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,494 @@
...
+                " (an archived skill has this name)" if skill.is_archived else ""
+            )
+            return (
+                f"skill name {name!r} already exists{archived_note} — skills are "
+                "install-once; pick a new name (a new version is a new skill)"
+            )
+    return None
+
+
+def _sweep_stale_staging(staging_root: Path) -> None:
```
leonardmq, line 209: Can we call that on startup of the desktop app to prevent staging from accumulating abandoned drafts here?

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,494 @@
...
+            continue
+        valid_paths.append(path)
+        total_bytes += len(content)
+        if len(content) > MAX_RESOURCE_FILE_BYTES:
+            errors.append(
+                f"file too large: {path!r} is {len(content)} bytes (cap {MAX_RESOURCE_FILE_BYTES})"
+            )
+        # The skill tool decodes references as UTF-8 at runtime — reject at
+        # install time, not at runtime. Assets may be binary.
+        if path.startswith("references/"):
```
leonardmq, line 163: What is the reason for doing this only for `references/`? Doesn't the same risk exist for the base SKILL.md too?

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,494 @@
...
+
+    Lenient to unreadable sibling skill files (e.g. written by a newer Kiln
+    version and synced in) — those can't be name-checked but must not make
+    every create fail.
+    """
+    existing, _load_errors = Skill.all_children_of_parent_path_with_errors(
+        project.path, readonly=True
+    )
+    for skill in existing:
+        if skill.name == name:
```
leonardmq: I believe we allow duplicate Skill names, otherwise we could not have coexisting versions of the same Skill - e.g. `call_api` V1, versus `call_api` V2. We only reject the same skill name within the same run config because that would not make any sense. But skills existing with the same name seems to be necessary. No?

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,494 @@
...
+    """Flush staged writes to stable storage before the commit rename, so a
+    power loss after install can never leave truncated files behind the
+    'fully formed' promise. Best-effort throughout."""
+    for dirpath, _dirs, files in os.walk(root):
+        for name in files:
+            _fsync_best_effort(os.path.join(dirpath, name))
+        _fsync_best_effort(dirpath)
+
+
+def _remove_staging_root_if_empty(staging_root: Path) -> None:
```
leonardmq, line 268: Any concurrency risk due to collaboration? For example, if two users are syncing the Kiln Project over git, and they both have drafts. How does the conflict get resolved, if there is a conflict at all? Should the staging dir be committed to git or added to a .gitignore we could bake into the Skill folder?

app/desktop/studio_server/skill_api.py
```diff
@@ -137,22 +272,125 @@ async def get_skill_content(
     @app.post(
         "/api/projects/{project_id}/skills", tags=["Skills"], openapi_extra=ALLOW_AGENT
     )
-    async def create_skill(
+    def create_skill(
```
leonardmq: Did you remove the `async` merely because you use a thread lock and it would not kick in on an async method (not sure if that is how the two combine), or is it just because we don't use any async calls within this method? If the latter, then I'd rather you keep it `async` (all our route controllers are `async` usually)

app/desktop/studio_server/skill_api.py
```diff
@@ -13,10 +23,46 @@
...
+# file I/O and fsyncs don't block the event loop; this lock restores the
+# serialization that makes the install-once name check race-free in-process.
+_skill_write_lock = threading.Lock()
+
+
+class SkillFileParam(BaseModel):
+    """A resource file to include in a skill bundle."""
+
+    path: str = Field(
+        description="Path within the skill bundle, starting with 'references/' or 'assets/'. Forward slashes only."
```
leonardmq, line 51: Is the `path` here validated to prevent traversal or random reads to arbitrary files?

libs/core/kiln_ai/datamodel/skill.py
leonardmq: Fixed in 974d71a: `_resource_base_dir` is now an explicit `if references / elif assets / else raise` — no default dir. If `RESOURCE_DIR_NAMES` grows (e.g. `scripts/`) without a reader being wired, calls fail loudly with "has no reader configured" instead of silently reading `assets/`.

---

libs/core/kiln_ai/datamodel/skill.py
leonardmq: I'd push back on this one. Binary-safe read is what makes binary assets (images, templates) servable at all — the bundle format explicitly allows binary under `assets/`, and this method backs the `resource_content` endpoint, which base64-encodes binary into JSON. Text-only would make any skill with an image asset unreadable through the API.

On the safety angle: the bytes are returned as data (base64 in a JSON response), never executed or interpreted server-side, the read is capped at 10MB, and path containment blocks reaching outside the skill dir. A malicious payload in a skill folder is no more dangerous read-as-bytes than sitting on disk. Text files still come back as text (`encoding: "utf-8"`); binary falls back to base64. The runtime `skill()` tool remains text-only via `read_resource_text`, unchanged.

Happy to restrict if you still prefer — but it would mean dropping binary asset support from the API.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
leonardmq: Done in 974d71a: new `sweep_stale_skill_staging(project_file_paths)` in `skill_bundle.py`, called from the desktop app's lifespan startup over `Config.shared().projects`. Best-effort, never raises, and keeps the 24h age gate so it can't touch another process's live install. Also removes the empty staging root afterwards.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
leonardmq: No change needed, I believe: SKILL.md can't hit this failure mode. Its body arrives as a JSON string field (`body: str`), so it's guaranteed decodable text by the time it reaches us (lone surrogates are rejected at request validation), and `save_skill_md` writes it as UTF-8. The `references/` check exists because those arrive as raw bytes (possibly base64-decoded binary) yet the runtime `skill()` tool decodes them as UTF-8 when serving them to a model — so we reject at install time rather than failing at runtime. `assets/` are deliberately exempt since they're allowed to be binary.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
leonardmq: You're right — removed in 974d71a. Run configs reference skills by id (`load_skills_from_tool_ids` resolves via `from_ids_and_parent_path`), so name is not identity and coexisting versions can share one. "Install-once" now means immutability only. Tests updated to assert same-name coexistence (create and clone).

One caveat worth knowing: `SkillTool` keys its lookup dict by name (`{s.name: s}`), so if two same-named skills were attached to the *same* run config, one would silently shadow the other at runtime — the per-run-config duplicate rejection you mentioned is what guards that, so it stays load-bearing.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
leonardmq: Low risk, and now gitignored (974d71a). Staging is transient: each install stages under a uuid-named dir and either renames it away or deletes it within the same call — it only persists after a hard crash, and the startup/install sweeps clean that up. Two collaborators can't collide: staging dir names are per-install uuids, and committed skill dirs are keyed by random 12-digit ids, so a git merge would just see two unrelated new directories.

It should definitely not be committed — I added `.skill_staging/` to the git-sync `.gitignore` (`OS_GITIGNORE` in `git_sync/clone.py`), which also back-fills the pattern into existing clones via the missing-patterns appender.

---

app/desktop/studio_server/skill_api.py
leonardmq: It was deliberate, not just style: plain `def` makes FastAPI run the handler in its threadpool so the install's file I/O + fsyncs don't block the event loop (and yes — a `threading.Lock` inside `async def` would block the loop, so the two went together; the lock restored the serialization that async handlers get for free).

Reverted to `async` in 974d71a per your preference and repo convention, and dropped the lock — with duplicate names now allowed, its main purpose (racing the name-uniqueness check) is gone, and async no-await handlers serialize naturally. Tradeoff to be aware of: a worst-case install (500 files, slow/synced disk) briefly blocks other requests; bundles are capped at 2MB so in practice it's sub-second.

---

app/desktop/studio_server/skill_api.py
leonardmq: Yes. The `path` goes through `Skill.read_resource_bytes` → prefix must be exactly `references/` or `assets/`, then `_resolve_resource` does `target.resolve()` + `resolved.relative_to(base_dir)` containment. That blocks `..` traversal, the pathlib absolute-join override (`base / "/etc/passwd"`), and symlink escapes (resolve follows the link, the real target fails containment — file and directory symlinks both). Covered by tests (`test_resource_content_traversal_rejected`, `test_read_resource_bytes_traversal`, and the tool-side traversal tests), and I additionally verified empirically with a battery including encoded slashes, absolute paths, and symlinked files/dirs pointing outside — all rejected.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,488 @@
...
+    project folders. Best-effort: never raises, never touches a staging dir
+    younger than STALE_STAGING_AGE_SECS (it could be another process's live
+    install).
+    """
+    for project_file_path in project_file_paths:
+        try:
+            staging_root = Path(project_file_path).parent / STAGING_DIR_NAME
+            if staging_root.is_dir():
+                _sweep_stale_staging(staging_root)
+                _remove_staging_root_if_empty(staging_root)
```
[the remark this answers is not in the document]
leonardmq: Valid — fixed in da0400f. The sweep now skips a symlinked staging root (`is_dir() and not is_symlink()`), and the create path also rejects a symlinked `.skill_staging` with a clear 422 instead of staging through it. Tests cover both: the sweep leaves a symlink target's aged dirs untouched, and create refuses to stage.

---

app/desktop/studio_server/skill_api.py
leonardmq: Does that work with Windows too if we only allow forward slashes? We support both windows and unix so should double check this. And do we validate these are relative paths? Since I think we stitch them together with the base path later on when persisting them, right?

leonardmq: Both good questions — one was fine, one was a real gap you caught. Fixed in a92d139.

**Forward slashes on Windows: yes, they work.** pathlib's `WindowsPath` accepts `/` as a separator for joins and resolution, so `references/foo/bar.md` reads correctly there. Backslash forms in the query param are rejected by the prefix parse (`references\x` doesn't split as `references` + `/`), and a Windows-style traversal like `assets/..\..\x` that survives parsing is still caught by the `resolve()` + `relative_to()` containment, which runs with the OS's own separator semantics. On the write side, backslashes are rejected outright for uploaded files.

**Relative validation before stitching: yes — and your instinct found a hole.** We rejected leading-`/` paths, but Windows drive forms (`C:/evil`, drive-relative `C:evil`) and leading backslashes pass a naive leading-slash check — and on Windows, joining `staging_dir / "C:/evil"` *replaces* the base entirely. Both validators (inline files and clone copy paths) now reject these explicitly and platform-independently (a deliberate regex rather than `os.path.isabs`, which only detects drive forms when running on Windows — validation must behave identically on every OS). Regression tests cover `C:/x`, `C:x`, and UNC forms. Practical exposure was narrow — inline uploads already required a `references/`/`assets/` prefix, so this mainly hardens the clone copy-path lane and the core API surface — but it's now closed at the validator level on both.

---

libs/core/kiln_ai/datamodel/skill_bundle.py
```diff
@@ -0,0 +1,499 @@
...
+        ) from None
+    if staging_root.is_symlink():
+        # Never stage through a symlink: the atomic-rename guarantee (and the
+        # cleanup paths) assume the real directory lives in the project folder.
+        raise SkillBundleValidationError(
+            [
+                f"{STAGING_DIR_NAME!r} in the project folder is a symlink — "
+                "remove it and retry"
+            ]
+        )
```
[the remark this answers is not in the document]
leonardmq: Valid — fixed in a92d139, same treatment as the staging-root case: after the `mkdir(parents=True, exist_ok=True)`, a symlinked `skills/` directory is rejected with a clear 422 before the commit rename can follow it. Regression test creates the symlink, asserts the create fails, and asserts nothing escaped through the link target.

---

leonardmq:
Merged `leonard/assistant-subagents` in (e190c2f) — the branch is mergeable again.

All three conflicts were in the same place: artifact provenance (#1565) has landed on the base, and it collides with the skill create/clone path this PR rewrites. Rather than pick a side, both features are kept and the "small follow-up" noted in the description is done here:

- **`skill_api.py`** — create keeps atomic bundle installation *and* accepts `provenance`. Lineage is validated **before** anything is staged, so a bad `derived_from_ids` returns 400 with no half-installed bundle and no staging debris left behind. The clone endpoint gets the same optional field.
- **`skill_bundle.py`** — `create_skill_with_files` / `clone_skill` take an optional provenance and stamp it on the skill they commit. A clone never inherits its source's provenance: a clone's provenance describes the clone, so the caller supplies it (including the `derived_from_ids` link back to the source). No auto-stamping, matching how every other create endpoint treats provenance.
- **`skill_form.svelte`** — clone still calls the dedicated clone endpoint (so `references/` and `assets/` are copied) while stamping the lineage the base branch introduced: `{origin: "human", derived_from_ids: [source_id]}` on clone, `{origin: "human"}` on a fresh create.

`api_schema.d.ts` regenerated; agent policy annotations verified current. 6 new tests cover the merged paths (bundle create with provenance, rejected lineage installs nothing, clone lineage recorded / absent / unknown-sibling 400). `checks.sh` green, full Python suite 8518 passed.

**One thread still needs your call**, not a code change: [binary-safe `read_resource_bytes`](https://github.com/Kiln-AI/Kiln/pull/1732#discussion_r3861943429) — you suggested text-only reads; I pushed back because that would drop binary asset support from the resource endpoint. Also left open is [why the UTF-8 check is `references/`-only](https://github.com/Kiln-AI/Kiln/pull/1732#discussion_r3861972915), answered but unconfirmed. The Windows-absolute-path thread is fixed and resolved.

---

PR #1734, opened 2026-08-27
Reuse MCP sessions without running a task

chiang-daniel:
## What does this PR do?

Promotes Kiln's MCP session-scoping idiom into a first-class API: `mcp_session_scope()`, an async context manager that gives any code block MCP session reuse without needing a task run in flight.

### Why

We found the gap during the eval builder v2 work. MCP session reuse is keyed off the agent run id, and only task execution sets it — so any code outside a run pays a fresh connect, initialize, and teardown per tool call. The spec builder describes a task's tools to the copilot before any run exists: a task with 4 tools on one local MCP server spawns that server 4 times per request, when one connection has all the answers. The code tool test-run endpoint hit the same wall earlier and worked around it by hand-rolling a run id plus cleanup. That workaround is now copy-pasted at five sites, and every copy risks leaking the server subprocess until app restart if cleanup is missed, since the session manager has no reaper.

There is no user-visible bug today and nothing here is ship-blocking. It is prevention for a class of latency and leak problems that has now bitten twice.

### The win

A small, contained change with real benefit. The helper puts the idiom in one place: it joins an open scope without owning its teardown, otherwise it creates a run id and guarantees cleanup on success and exception, keyed to the id it created. The five hand-rolled sites now use it with no behavior change, and new tests cover the helper's guarantees directly and pin cleanup at each migrated site.

The benefit going forward: any code that needs MCP access gets session reuse (one connection per server instead of one per tool) and guaranteed cleanup with a single line, instead of invoking a task or re-implementing the idiom — first up, the eval builder's capability collection, which shaves the per-tool server spawns off every copilot request.

## Checklists

### DIFF
--- a/libs/core/kiln_ai/tools/mcp_session_manager.py
+++ b/libs/core/kiln_ai/tools/mcp_session_manager.py
@@ -16,6 +16,12 @@ from mcp.client.streamable_http import streamablehttp_client
 from mcp.shared.exceptions import McpError
 
 from kiln_ai.datamodel.external_tool_server import ExternalToolServer, ToolServerType
+from kiln_ai.run_context import (
+    clear_agent_run_id,
+    generate_agent_run_id,
+    get_agent_run_id,
+    set_agent_run_id,
+)
 from kiln_ai.utils.config import Config
 from kiln_ai.utils.exhaustive_error import raise_exhaustive_enum_error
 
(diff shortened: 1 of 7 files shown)


libs/core/kiln_ai/tools/mcp_session_manager.py
```diff
@@ -501,6 +508,40 @@ def clear_shell_path_cache(self):
         self._shell_path = None
 
 
+@asynccontextmanager
+async def mcp_session_scope() -> AsyncGenerator[None, None]:
```
chiang-daniel, line 512: This is the main change.

leonardmq (approved):

PR #1735, opened 2026-08-27
Reject duplicate skill names at run config creation; surface save errors on the run page

leonardmq:
## What does this PR do?

Two small, related fixes around run configs and skills:

1. **Fail fast on duplicate skill names in a run config.** The agent loads skills by name at runtime, so a run config attaching two same-named skills fails only when it's used (the adapter raises). `POST /api/projects/{p}/tasks/{t}/run_configs` now validates up front and returns a 422 naming the colliding skills — the user finds out at creation, with a clear message, instead of at inference time. The runtime check remains as backstop for configs created outside the API.
2. **Surface save errors from the run page's "Save current options" action.** Its handler was `try/finally` with no `catch`, so any failed save spun briefly and silently did nothing. The component already declared and rendered `save_config_error`; the handler now sets it (and clears it on the next attempt), so server errors — including the new 422 — display under the Run Configuration dropdown.

### Manual testing

Verified on this branch against a live dev server:

- **API — `validate_unique_skill_names` kicks in:** `POST /run_configs` with two same-named skills attached → `HTTP 422` with body `{"message":"Multiple selected skills share the same name: ticket-api. Skills are loaded by name at runtime, so each skill attached to a run config must have a unique name."}`. Same request with a single skill → `HTTP 200`.
- **UI — error path:** selected two same-named skills in the run page's Skills dropdown, clicked "Save current options" → the 422 message renders under Run Configuration (previously: nothing happened).
- **UI — recovery:** deselected one, saved again → error clears, the new config ("Ruffled Putter") is created and selected, and the inline action flips to "Set as task default".

<details>
<summary><b>Screenshots</b> (captured on this branch; hosted on the <code>claude/run-config-validation-assets</code> branch — never merge it, safe to delete after this PR closes)</summary>

**Skills selector allows picking two same-named skills (versions of one skill):**

skills selector

**Save current options with both selected — the validation error is now shown:**

save error

**After deselecting one — save succeeds, error cleared:**

save success

</details>

## Related Issues

None.

## Checklists

- [x] Tests have been run locally and passed (`checks.sh` green)
- [x] New tests have been added to any work in /lib (endpoint test covering reject + accept cases)

https://claude.ai/code/session_016hr8tm9wQRGYZyTPpcy1kr

### DIFF
--- a/app/desktop/studio_server/eval_api.py
+++ b/app/desktop/studio_server/eval_api.py
@@ -6,6 +6,7 @@ from typing import Annotated, Any, Dict, List, Set, Tuple, Type, TypeVar
 
 from fastapi import FastAPI, HTTPException, Path, Query, Request
 from fastapi.responses import StreamingResponse
+from kiln_ai.adapters.adapter_registry import load_skills_from_tool_ids
 from kiln_ai.adapters.eval.base_eval import (
     DEFAULT_SYSTEM_PROMPT,
     build_default_llm_judge_prompt,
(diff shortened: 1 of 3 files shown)


chiang-daniel:
What happens to existing tools that are already duplicated? Do we continue to let them error out when they run?

chiang-daniel (approved):

leonardmq:
> What happens to existing tools that are already duplicated? Do we continue to let them error out when they run?

Yes

PR #1739, opened 2026-08-30
test: remove obsolete test_structured_input_gpt_4o_mini

Anai-Guo:
## Problem

`test_structured_input_gpt_4o_mini` cannot run. It calls the shared helper with
three arguments:

```python
# libs/core/kiln_ai/adapters/model_adapters/test_structured_output.py:525
@pytest.mark.paid
async def test_structured_input_gpt_4o_mini(tmp_path):
    await run_structured_input_test(tmp_path, "llama_3_1_8b", "groq")
```

but the helper has four required parameters:

```python
# same file, line 419
async def run_structured_input_test(
    tmp_path: Path, model_name: str, provider: str, prompt_id: PromptId
):
```

so the test raises `TypeError: run_structured_input_test() missing 1 required
positional argument: 'prompt_id'` before it reaches any provider call. Because
it is `@pytest.mark.paid`, `conftest.py:157` skips it unless `--runpaid` is
passed, which is why it has stayed green in normal runs — but it fails for
anyone running the paid suite (including the pre-release smoke path).

## Which value to pass

The file's other two call sites both pass `"simple_prompt_builder"`:

| line | test | args |
|---|---|---|
| 525 | `test_structured_input_gpt_4o_mini` | 3 — **missing `prompt_id`** |
| 534 | `test_all_built_in_models_structured_input` | 4, `"simple_prompt_builder"` |
| 565 | `test_all_built_in_models_structured_input_mocked` | 4, `"simple_prompt_builder"` |

The mocked test at line 565 is this test's twin — same model (`llama_3_1_8b`)
and same provider (`groq`) — so `"simple_prompt_builder"` keeps the paid and
mocked variants testing the same thing.

## Verification

The helper's signature was parsed out of the file with `ast`, rebuilt as a stub,
and all three call sites replayed through `inspect.Signature.bind`, so the two
working call sites act as a control group:

```
def run_structured_input_test(tmp_path, model_name, provider, prompt_id)   @ line 419

before:
  line  525  3 positional  in test_structured_input_gpt_4o_mini
            -> TypeError: missing a required argument: 'prompt_id'
  line  534  4 positional  in test_all_built_in_models_structured_input      -> binds
  line  565  4 positional  in test_all_built_in_models_structured_input_mocked -> binds
  1 of 3 call sites cannot bind

after:
  line  525  4 positional  -> binds
  line  536  4 positional  -> binds
  line  567  4 positional  -> binds
  0 of 3 call sites cannot bind
```

## Notes

- Test-only change, 1 call site. No library code touched.
- Aside, not changed here: the test is named `test_structured_input_gpt_4o_mini`
  but actually exercises `llama_3_1_8b` on `groq`. Happy to rename it in this PR
  if you'd like, but I kept the change to the failure itself.

### DIFF
--- a/libs/core/kiln_ai/adapters/model_adapters/test_structured_output.py
+++ b/libs/core/kiln_ai/adapters/model_adapters/test_structured_output.py
@@ -521,11 +521,6 @@ async def run_structured_input_task(
 
 
 @pytest.mark.paid
-async def test_structured_input_gpt_4o_mini(tmp_path):
-    await run_structured_input_test(tmp_path, "llama_3_1_8b", "groq")
-
-
-@pytest.mark.paid
 @pytest.mark.ollama
 @pytest.mark.parametrize("model_name,provider_name", get_all_models_and_providers())
 async def test_all_built_in_models_structured_input(


libs/core/kiln_ai/adapters/model_adapters/test_structured_output.py
```diff
@@ -522,7 +522,9 @@ async def run_structured_input_task(
 
 @pytest.mark.paid
 async def test_structured_input_gpt_4o_mini(tmp_path):
```
leonardmq, line 524: This test is outdated - we have `test_all_built_in_models_structured_input` that is parametrized that replaces this hardcoded model one.

Could you please just delete this whole `test_structured_input_gpt_4o_mini` test?

leonardmq (changes requested):
Thanks for the contribution - the test this PR touches is obsolete and superseded by one that parametrize over many models, so we should just delete this one (instead of adding this prompt builder arg). Can you please delete it?

Anai-Guo:
Thanks @leonardmq — done. I've dropped the `prompt_id` change and the PR now simply deletes `test_structured_input_gpt_4o_mini`, since `test_all_built_in_models_structured_input` already covers it parametrized across models.

The branch is rebased on current `main`, so the diff is just the deletion.

leonardmq (approved):

scosman:
thanks!

PR #1762, opened 2026-09-10
Revert nltk 3.10.3 bump

chiang-daniel:
## Why

`Test Count And Coverage` went red on main at `25d1b9a` after the dependency batch landed. The cause is the nltk bump in #1742.

nltk 3.10.3 added a path-security guard that refuses to open any corpus file with more than one hard link (CWE-59). llama_index ships its stopwords corpus inside `site-packages`, and uv installs by hard-linking from its cache on Linux, so the file arrives with `st_nlink=2` and nltk rejects it:

```
PermissionError: Security Violation [pathsec.open]: refusing multiply-linked file
'.../llama_index/core/_static/nltk_cache/corpora/stopwords/english' (st_nlink=2);
a hardlink can point at an outside-root inode (CWE-59)
```

That failed `test_fixed_window_chunker_concrete_chunker`. Every other workflow stayed green, including the full suite on Python 3.10 through 3.13 — `Test Count And Coverage` is the only job that runs the tests serially (`-n0`), and the guard only trips there.

## What this does

Reverts the merge of #1742, putting nltk back to 3.10.0 in `uv.lock`. Nothing else changes.

## Trade-off

This gives back the CVE-2026-12252 / CVE-2026-12841 fixes that shipped in 3.10.3. That is a real cost, taken deliberately to get main green now rather than leave trunk red while the upstream conflict is worked out.

## Follow-ups

- The nltk / llama_index hard-link conflict is unresolved and will come back the next time Dependabot opens this bump. Re-landing needs it fixed first.
- Worth investigating separately: `test_count.yml` still runs `coverage run -m pytest ... -n0`, the pattern #1742's failure depends on. `coverage.yml` moved off it to `pytest --cov` under xdist. Bringing the two in line may remove the serial-only exposure entirely.

## Priority

Medium. Trunk is red on one job; the library itself is unaffected and the other ten workflows are green.

https://claude.ai/code/session_01T6NC2nGHWRxyHeo81YkWch

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2312,7 +2312,7 @@ wheels = [
 
 [[package]]
 name = "nltk"
-version = "3.10.3"
+version = "3.10.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "click" },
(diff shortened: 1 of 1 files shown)


chiang-daniel:
@coderabbitai full review

leonardmq (approved):

PR #1765, opened 2026-09-11
feat(compare): radar scale toggle, hideable usage axes, row renames

tawnymanticore:
## What does this PR do?

**TLDR:** The compare radar chart said nothing with one or two run configs, because every axis was scaled to the selection and a missing score was drawn as zero. This PR adds a Relative / Full Scale toggle, plots only the scores every plotted config has, turns the cost, latency and token rows into hideable axes, and lets the user hide or rename any row. It replaces #1628 with the same feature on top of the current `main`, moves the chart maths into tested modules, and fixes three defects the new tests exposed. Targets `main`. There are no companion branches.

Before this PR, the radar chart on the compare page scaled each axis to the highest value among the selected run configs. One run config sat on the outer ring of every axis. A score with no result was plotted at the centre, which reads the same as a score of zero. Only eval sections had a hide control. After this PR, the chart has a Relative / Full Scale toggle, plots only the scores every plotted run config has a result for and reports the omitted count under the title, charts cost, latency and tokens as relative axes that the user can hide from the table, and lets the user rename any row for display. The chart logic lives in pure modules with unit tests, and the option object that reaches ECharts is asserted by a test.

**Implementation**

- Feature, unchanged from #1628 (first commit, byte-identical to that branch): the scale toggle, the shared-score filter with the omitted count, the five usage axes with "bigger means less" scoring, the hide and rename controls on every table row, and the URL parameters `hidden_metrics` and `metric_labels` that persist them.
- Chart data layer: `lib/utils/radar_chart_data.ts` holds score normalisation, axis scaling, usage-axis inversion, the shared-key filter, label resolution and the tooltip trim. The component computes chart data once per update instead of twice. Its script drops from 629 lines to 374. The whole file is 430 lines; on `main` today it is 417 lines without the feature.
- Shared constants: `lib/utils/compare_metric_keys.ts` holds the cost section id and the five usage keys. The page and the chart import them instead of repeating string literals.
- Typed score range: the page's score-range lookup takes `TaskOutputRatingType` and ends with `assertNever`, so a new rating type fails the type check instead of returning `null`. The lookup and the axis-max builder move into `compare_view.ts`.
- Tests: 33 component tests mock ECharts and assert on the option object that reaches `setOption`. 32 unit tests cover the data module. `compare_view.test.ts` grows from 20 to 26 cases. This is the first unit test of an ECharts component in the repo.
- Fix: score labels, run config names and model names are HTML-escaped in the tooltip and the legend. Labels come from eval score names and from the `metric_labels` URL parameter, so a shared link could inject markup.
- Fix: two run configs with the same name are separate series. Before, they collapsed into one legend entry that showed the first config's model, cost and scores.
- Fix: the note under the title counts omitted axes with the correct word. It counted usage axes together with eval scores and called all of them "scores".
- Fix: the radar card stays on the page when hidden rows leave fewer than three axes, and its empty state says how to get axes back. Before, the card disappeared with no message.
- Fix: the radar centre and radius change in the wide layout so the left-hand axis labels fit. Before, "Input Token Efficiency" was clipped to "…en Efficiency" at the card edge.
- Fix: the hide and rename controls have accessible names, and the rename input has a 60-character limit.
- Fix: the chart redraws when the prompt list becomes `undefined` after it loaded. Before, the guard in the reactive statement stopped the redraw.
- Rename: the page's hide state is named for what it hides, sections rather than evals, because the cost section is hideable too. The URL parameters keep their names. The section type is defined once, in `compare_metric_keys.ts`, and the chart's `comparisonFeatures` prop uses it: same shape, one field fewer.
- Copy: the chart subtitle and the tooltip header cover the usage axes. `escape_html` also escapes the single quote.
- Page-level, carried over from #1628: `updateURL()` passes `noScroll` and `keepFocus` to `goto()`, so a state change no longer jumps the page to the top. The first table column grows from 200 px to 240 px to fit the rename control.

**Before you merge**

- A user sees these changes: the scale toggle, the hide and rename controls on every row, the usage axes on the radar, and the omitted-score count. The old page hid eval sections only.
- In the wide layout the radar is drawn smaller (radius 85 % to 70 %, centre 32 % to 36 %) so the axis labels fit. The compact layout for one or two run configs is unchanged.
- Hiding rows no longer removes the radar card. The card shows an empty state with the ways to get axes back.
- Not included: skipping `lower_is_better` or `informational` scores from the radar. `EvalOutputScore` has no `direction` field on `main`.

<details>
<summary><b>How the refactor was checked</b></summary>

The first commit reproduces the #1628 branch tip `ba08b006e` on `main`; the five changed files are byte-identical (`git diff` between the two is empty for those paths). A test harness was committed on top of it before any refactor. The refactor commits were then gated three ways:

1. The option object the component hands to ECharts, serialised with its formatter functions invoked, for six fixtures. The dump after each refactor commit is byte-identical to the dump at the baseline (`diff -r` is empty).
2. Full-page screenshots of six page states on the seeded fixture project, rendered from a checkout at the baseline commit and from the final commit, on separate ports. Two consecutive renders of the baseline are pixel-identical, so the comparison has no noise. At `e9020ebab`, after the three refactor commits and the first four fixes, all six states are pixel-identical to the baseline (0 pixels differ). At the final commit the differences are confined to the radar plot area in the three-config states (the resize for the labels) and to one subtitle line in the one- and two-config states, measured by the bounding box of the differing pixels.
3. The harness and the existing `compare_view` tests pass unchanged through the refactor commits. The three fix commits each change exactly the assertions they say they change.

Commits, in order:

- `d199675e9` feat(compare): radar chart scale toggle, hideable usage axes, no false zeros, renamable rows — #1628 squashed, byte-identical in its five files
- `0271edcbf` test(compare): characterise the compare radar chart option at baseline — 25 cases, no component change
- `1559d98a2` refactor(compare): extract the radar chart data layer into a tested module — golden identical
- `611c863af` refactor(compare): share the cost section id and usage keys between page and chart — golden identical
- `7d3351b6b` refactor(compare): type the score range lookup and move it into compare_view — golden identical
- `5c278e1b9` fix(compare): escape score labels in the radar tooltip — golden identical (no fixture carries markup characters); one harness assertion changed
- `f12bc6163` fix(compare): keep run configs that share a name apart on the radar — golden changes only in the duplicate-names fixture; one harness assertion changed
- `e3712603f` fix(compare): count omitted axes, not scores, under the radar title — DOM text only
- `e9020ebab` fix(compare): name axes consistently in the radar empty state — DOM text only
- `45b5bfbbb` fix(compare): keep the radar card when hidden rows leave too few axes — DOM only, golden identical
- `db31ec158` fix(compare): make room for the radar axis labels — golden changes only in `radar.center`, `radar.radius`, `axisName.width`
- `3ece384c1` test(compare): assert the radar option shape instead of a tautology — tests only
- `4bb7c8945` fix(compare): label the hide and rename controls for assistive technology — golden identical
- `a4498758d` refactor(compare): name the hide state for what it hides — golden identical
- `bc245af22` fix(compare): redraw the radar when the prompt list loads — golden identical; one new harness case
- `96ac35cfa` fix(compare): update the radar copy for usage axes — golden changes only in tooltip header text

</details>

<details>
<summary><b>Screenshots</b></summary>

Hosted on the `mike/compare-radar-legibility-v2-assets` branch. Never merge that branch. The team can delete it after this PR closes.

**Three run configs, relative scale. The table has hide and rename controls on every row; the radar plots one eval axis and five usage axes:**

three run configs, relative scale

**Full scale. The eval axis uses its own 0 to 1 range, and the axis labels fit:**

three run configs, full scale

**The same view on the #1628 branch, for comparison. The left labels are clipped:**

before: labels clipped

**A row renamed to "Escalation" through the rename control. The name changes on the table, the axis, and the tooltip:**

renamed row

**Two usage rows hidden from the table. They leave the radar, and the "Show Section" menu lists them:**

hidden usage rows

**A single run config defaults to full scale with an area fill:**

single run config

</details>

## Related Issues

Replaces #1628. No tracked issue.

## Checklists

- [x] Tests have been run locally and passed (`uv run ./checks.sh --agent-mode` exits 0 in a clean worktree; `npm run check` reports 0 errors and the same 38 warnings as `main`; `misspell` is not installed locally, so CI covers that check)
- [ ] New tests have been added to any work in /lib — not applicable, this PR changes the web UI only

### DIFF
--- a/app/web_ui/src/lib/components/chart_no_data.svelte
+++ b/app/web_ui/src/lib/components/chart_no_data.svelte
@@ -1,3 +1,8 @@
+<script lang="ts">
+  export let title: string = "No Data Available"
+  export let message: string = "Create and run evals to see a comparison chart."
+</script>
+
 <div class="flex items-center justify-center h-[300px]">
   <div class="text-center">
     <svg
@@ -13,9 +18,7 @@
         d="M9 19v-6a2 2 0 00-2-2H5a2 2 0 00-2 2v6a2 2 0 002 2h2a2 2 0 002-2zm0 0V9a2 2 0 012-2h2a2 2 0 012 2v10m-6 0a2 2 0 002 2h2a2 2 0 002-2m0 0V5a2 2 0 012-2h2a2 2 0 012 2v14a2 2 0 01-2 2h-2a2 2 0 01-2-2z"
       />
     </svg>
-    <p class="text-lg font-medium">No Data Available</p>
-    <p class="mt-1 text-gray-500">
-      Create and run evals to see a comparison chart.
-    </p>
+    <p class="text-lg font-medium">{title}</p>
+    <p class="mt-1 text-gray-500 max-w-md">{message}</p>
   </div>
 </div>
--- a/app/web_ui/src/lib/components/compare_chart.svelte
+++ b/app/web_ui/src/lib/components/compare_chart.svelte
@@ -15,17 +15,10 @@
   } from "$lib/utils/run_config_formatters"
   import ChartNoData from "./chart_no_data.svelte"
   import { formatLatency } from "$lib/utils/formatters"
-
-  // Type for comparison features (same as parent page)
-  type ComparisonFeature = {
-    category: string
-    items: { label: string; key: string }[]
-    has_default_eval_config: boolean | undefined
-    eval_id: string
-  }
+  import type { ComparisonSection } from "$lib/utils/compare_metric_keys"
 
   // Props
-  export let comparisonFeatures: ComparisonFeature[]
+  export let comparisonFeatures: ComparisonSection[]
   export let getModelValueRaw: (
     modelKey: string | null,
     dataKey: string,
--- a/app/web_ui/src/lib/components/compare_radar_chart.svelte
+++ b/app/web_ui/src/lib/components/compare_radar_chart.svelte
@@ -96,7 +96,7 @@ Cost, latency and token axes score each run config against the others, so they s
   $: notShownNote =
     omittedAxisCount > 0
       ? `Not shown: ${omittedAxisCount} ${
-          omittedAxisCount === 1 ? "score" : "scores"
+          omittedAxisCount === 1 ? "axis" : "axes"
         } without results for every selected run config. See the table above.`
       : null
 
--- a/app/web_ui/src/lib/components/compare_radar_chart.svelte
+++ b/app/web_ui/src/lib/components/compare_radar_chart.svelte
@@ -142,7 +142,10 @@ Cost, latency and token axes score each run config against the others, so they s
     }
   }
 
-  function buildLegendSubtext(config: TaskRunConfig): string {
+  function buildLegendSubtext(
+    config: TaskRunConfig,
+    promptList: PromptResponse | null,
+  ): string {
     const parts: string[] = []
     if (isMcpRunConfig(config.run_config_properties)) {
       const toolName =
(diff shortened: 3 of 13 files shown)


scosman (changes requested):
Not sure I like "a dozen pencils" UI, or an edit UI that doesn't save the edit. We should chat through the UI design.

PR #1774, opened 2026-09-16
Multi-turn tasks and the eval builder

chiang-daniel:
## What does this PR do?

Brings the multi-turn line to main. This branch is the hub two pieces of work were built on, and it lands them together because the second depends on the first.

**Multi-turn tasks** (Leonard, reviewed and merged into this branch in #1409, approved by Steve). Tasks can run as conversations: a chat-style run page, forking from an assistant message, cumulative usage across turns, and the run chain model underneath.

**The eval builder** (#1731, in review now, merging into this branch). A spec-driven eval builder for single-turn and multi-turn tasks: generated data, real traces, a human-calibrated LLM judge, and a save that deals the data into test, train, val and golden. Until #1731 merges into this branch, the diff here is the multi-turn half only.

## Sequencing

- kiln_server side is Kiln-AI/kiln_server#258, on staging now. It has to be on prod before this merges, or the desktop calls routes that are not there.
- #1731 merges into this branch first, then this PR becomes the single merge to main.
- After main: publish the synthetic-user model flag in remote config; KIL-807 and KIL-808 ship as their own PRs.

## Testing

Trial merge against main is conflict-free. CI here runs the hub against today's main; #1731 carries its own 13 checks and the end-to-end walk record.


chiang-daniel:
Closing: Steve asked for a single PR into main, so #1731 now targets main directly and carries both halves. Nothing here that #1731 does not.

PR #1778, opened 2026-09-16
[Eval V2 Kiln, review-only 1/7] Datamodel: score pointers, splits, synthetic user, claim reviews

chiang-daniel:
**Review-only.** Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with `libs/core/kiln_ai/datamodel` reverted, so "Files changed" is exactly the datamodel: 11 source files (+304 / −90) and 4 test files. Details are inline on each hunk.

## How it works

The builder's five steps each write one kind of record into the project on disk. Nothing is copied between records: a score points at the conversation it scored, a review sits under the conversation it grades. Existing evals and judges load exactly as before; the new fields are optional with defaults.

```mermaid
flowchart TB
  Eval["Eval: the spec, plus splits (which tag is test / train / val / golden)"]
  EvalConfig["EvalConfig: one judge; V2 judges get typed settings"]
  EvalInput["EvalInput: one generated case (persona, goal, first message, drive settings)"]
  TaskRun["TaskRun: one conversation, trace inside, synthetic user spend beside the agent's"]
  EvalRun["EvalRun: one score; points at the run it scored"]
  ClaimReview["ClaimReview: your agree / disagree per claim and your verdict"]
  Eval --> EvalConfig
  Eval --> EvalInput
  EvalInput -- "driven at eval time, once per run config" --> TaskRun
  EvalConfig -- "scores" --> EvalRun
  EvalRun -- "dataset_id or eval_input_id" --> TaskRun
  TaskRun -- "child record" --> ClaimReview
```

- **Step 4, generated cases** write an `EvalInput` each. Multi-turn: the synthetic user (`SyntheticUserInfo`), the first message, and `MultiTurnDriveConfig` (which model plays the user, how many turns). Stamped once, never edited.
- **The conversations that ran** are ordinary `TaskRun`s with the trace inside. New: `synthetic_user_usage`, the user model's spend, kept apart from the agent's.
- **The judge's pass or fail** is an `EvalRun`. V2 records are pointers (`dataset_id` or `eval_input_id`), never a copy of the conversation. A judge that reads the whole trace cannot save a score without one.
- **Step 5, the claim review** is a `ClaimReview` stored under the conversation: the judge's claims, your grade on each, your verdict. Re-running the judge never overwrites it.
- **Save** deals items into splits by tag; the eval's `splits` field names the tag per split. Bad tags are rejected on save and on load.
- **The judge itself** is an `EvalConfig`; only V2 judges parse `properties` into typed settings, legacy dicts stay dicts.

Everything else is small and labelled inline: a filter-order fix, a readonly passthrough, a code-tool validation, one constant, a dead branch, exports.

## Why this PR is first

These records are what users' evals are made of from the first release onward. A wrong field here means a migration on their files later; a wrong behaviour anywhere else is a fix and a release. Read this one line by line; the other six can be skimmed.

## How to comment

Mark **Blocker** (correctness, data loss, security, back-compat) or **Later**. Blockers are fixed on `dchiang/eb-v2-merge` in one round and this view is regenerated. Inline markers: **Bug** = fix it; **Decision made** = built this way on purpose, say yes or push back; anything else is a note.


libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -470,7 +475,15 @@ def validate_scores_against_output_scores(
     """
 
     def _is_numeric(v: object) -> bool:
-        return isinstance(v, (int, float)) and not isinstance(v, bool)
+        # NaN compares False against every range bound, so it passes every check
```
chiang-daniel, line 478: **Bug fix in score validation: a `NaN` score passed every check and corrupted the saved file.**

**What this is:** `EvalScores` is the dict of scores a judge gives one run, keyed by the eval's output score names. Each output score declares a type (`five_star` 1 to 5, `pass_fail`, `pass_fail_critical`, `custom`) and this helper checks each value against its type's range. It returns problem strings rather than raising; callers decide what to do.

**The bug:** `NaN` is a float, and `NaN` compares `False` against every bound, so a `NaN` score passed as valid. It then serialized to JSON `null`, and the saved eval run failed to load.

**The fix:** a score must be a finite number. The `OverflowError` branch covers ints too large to become a float.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -532,15 +546,94 @@ class UserMessage(BaseModel):
     text: str
 
 
+class SyntheticUserInfo(BaseModel):
```
chiang-daniel, line 549: **New model: the synthetic user who plays the other side of a multi-turn conversation. Needed so a multi-turn eval item can be re-driven at eval time from the item alone.**

An `EvalInput` is one item of an eval's generated dataset; its `data` is `SingleTurnEvalInputData` (one user message) or `MultiTurnSyntheticEvalInputData` (below). This class is the multi-turn item's user: `persona` (who they are), `goal` (what they want), `behavior_guidance` (optional, how they act). The driver renders the user model's system prompt from these fields at eval time. `extra="allow"` keeps fields a newer generator adds through a load-save round trip on an older build.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -532,15 +546,94 @@ class UserMessage(BaseModel):
...
+    the item reusable across evals.
+
+    first_message may be None; such items carry no seed to open a
+    conversation with, so the eval runner skips them instead of re-driving.
+    """
+
     type: Literal["multi_turn_synthetic"] = "multi_turn_synthetic"
     first_message: UserMessage | None = None
-    synthetic_user_info: dict[str, JsonValue] = {}
+    synthetic_user_info: SyntheticUserInfo
```
chiang-daniel, line 612: **New field, and a decision made: drive settings live on the item, nowhere else.**

**Why:** a multi-turn item is a replication recipe: `first_message`, `synthetic_user_info`, and this `drive_config` (which model plays the user, on which provider, the ceiling on assistant turns). The runner re-drives it with the run config under test as the agent and this user held constant, so comparing two run configs varies only the agent.

**Decision made:** the settings are stamped when the item is minted and never edited; traces are keyed to the item, so a changed setup means new items and existing traces stay valid. An item with no `drive_config` is skipped with `SkippedReason.missing_drive_config`, never driven with a guessed config. The wrap serializer omits an unset `drive_config`, so such items are not rewritten with a `null` on resave.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -571,6 +664,17 @@ class EvalInput(KilnParentedModel):
         description="Tags for filtering eval inputs.",
     )
 
+    @model_validator(mode="after")
```
chiang-daniel, line 667: **Decision made: eval input tags follow the same rule as task-run tags.**

**What this is:** `EvalInput.tags` is how items are dealt into splits. The eval's `splits` name a tag per split (test, train, val, golden) and the runner selects items by tag. Every tag filter in the app treats an empty or space-containing tag as unselectable.

**Decision made:** reject such a tag when the item is validated, exactly as `TaskRun.tags` already does. A rejected write beats an item that silently belongs to no split.

**Note:** this validator has no loaded-from-file exemption, so a file with a bad tag fails to load rather than quietly missing from every split.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -624,12 +728,24 @@ def from_trace(
             trace_data = [dict(msg) for msg in trace.trace]
 
         if isinstance(source, EvalInput):
-            if not isinstance(source.data, SingleTurnEvalInputData):
-                raise ValueError("EvalTaskInput only supports single-turn EvalInput")
             reference_data = source.reference
             # The item's own text, not the trace's: an EvalInput is the canonical
             # statement of the input, and the adapter may have reserialized it.
-            task_input = source.data.user_message.text
+            if isinstance(source.data, SingleTurnEvalInputData):
```
chiang-daniel, line 734: **New case: how a multi-turn item becomes the judge's input.**

**Why:** a judge scores a bundle built from the item and its trace; for a multi-turn case nothing built that bundle, so every multi-turn item was skipped as unsupported before it reached a judge.

`EvalTaskInput` is what a V2 judge receives: `trace`, `task_input` (the canonical input text), `reference_data` (ground truth, if any) and the final message; `from_trace` builds it from the `TaskRun` holding the trace plus the item it came from (an `EvalInput` or a dataset `TaskRun`). For a multi-turn `EvalInput`, `task_input` is the item's `first_message`, since the rest of the exchange is in the trace. An item minted without a first message yields `None`; the runner skips such items before reaching here.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -861,35 +985,49 @@ def validate_record_mode(self) -> Self:
...
-        if parent_eval_config and parent_eval_config.config_type == EvalConfigType.v2:
+            # Pointer mode: the trace lives on the referenced TaskRun, and
+            # validate_record_mode already forbids inline copies here. The checks
+            # below are about data carried on this record, so none of them apply.
             return self
         parent_eval = parent_eval_config.parent_eval() if parent_eval_config else None
         if not parent_eval:
             return self
 
+        evaluation_data_type = parent_eval.evaluation_data_type
```
chiang-daniel, line 999: **Bug fix: a judge that reads the whole conversation could save a score with no conversation attached.**

**The bug:** the validator returned early for every V2 config, so single-turn full-trace runs were saved without their trace.

An `EvalRun` is a score record. V1 records copy the output inline (and, for full-trace evals, the conversation in `task_run_trace`); V2 records are pointers (`dataset_id` or `eval_input_id` name the `TaskRun` that holds the trace; `validate_record_mode` above enforces nothing inline). Three fields classify a run: `evaluation_data_type` on the eval (`full_trace` or `final_answer`), `eval_config_eval=True` (a calibration run over a human-rated item), `skipped_reason` (no output).

**Now:** a real scoring run, not skipped, of an eval whose judge reads the trace, with no trace attached, refuses the write, unless the record is being loaded from disk. Then V2 returns early because the remaining checks are about inline output, which V2 never has.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -903,6 +1041,12 @@ def validate_eval_run_types(self) -> Self:
             raise ValueError(
                 "task_run_config_id must be set if eval_config_eval is false"
             )
+        if self.eval_config_eval and self.dataset_id is None:
```
chiang-daniel, line 1044: **New invariant: a calibration run must score a dataset item.**

**What this is:** an `EvalRun` with `eval_config_eval=True` is a calibration run. It scores an item that already carries a human rating so the judge's score can be compared against the human's. `dataset_id` and `eval_input_id` are the two pointer fields naming what was scored; exactly one is set.

**Why:** only dataset `TaskRun`s carry human ratings. Generated `EvalInput`s do not.

**What changed:** before this, a calibration run pointing at an `EvalInput` could be written and would never take part in calibration.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -972,33 +1116,32 @@ class EvalConfig(KilnParentedModel, KilnParentModel, parent_of={"runs": EvalRun}
         default=EvalConfigType.g_eval,
         description="This is used to determine the type of eval to run.",
     )
-    properties: V2EvalConfigProperties | dict[str, Any] | None = Field(
+    properties: dict[str, Any] | V2EvalConfigProperties | None = Field(
```
chiang-daniel, line 1119: **Bug fix in how judge-config properties load, and a decision made on the union's order.**

`EvalConfig` is one judge; `config_type` is `g_eval` or `llm_as_judge` (legacy, `properties` an arbitrary dict) or `v2` (`properties` a dict on disk whose `type` key selects a typed class in the `V2EvalConfigProperties` discriminated union).

**The bug:** the previous before-validator shallow-copied the dict back onto itself, a no-op, so legacy dicts only loaded because of the union's order.

**Decision made:** the union lists `dict` first, so a raw dict always stays a dict, even one whose keys match a typed shape; V2 configs are parsed into the typed union explicitly in `dispatch_properties_parsing` below.

libs/core/kiln_ai/datamodel/task_run.py
```diff
@@ -106,6 +109,15 @@ class TaskRun(
             "cost fields equal those of `usage`."
         ),
     )
+    synthetic_user_usage: Usage | None = Field(
```
chiang-daniel, line 112: **New field: the synthetic user's spend, kept apart from the agent's. Needed because an eval-driven conversation pays a third model, the user, whose spend had nowhere to go.**

A `TaskRun` carries `usage` (the agent's tokens and cost) and `cumulative_usage` (the same summed along a multi-turn chain). This field holds the user model's spend as a full `Usage`, so tokens can be split per provider against an invoice, not only a cost.

**`None` when:** ordinary runs, a drive that reused an existing trace (no driver call happened), and a provider that reported nothing. Never a zeroed `Usage`, so an unmeasured drive cannot read as a free one.

libs/core/kiln_ai/datamodel/task_run.py
```diff
@@ -119,6 +131,15 @@ class TaskRun(
         description="Set when this run was generated by an eval. Names the eval dataset item it was generated for. None for ordinary dataset runs. Runs with this set are excluded from Task.runs() by default, so they do not appear on dataset surfaces.",
     )
 
+    @model_validator(mode="after")
```
chiang-daniel, line 134: **Crash fix: traces with content parts could not be copied, and the whole request failed.**

**What this is:** `TaskRun.trace` is the conversation in OpenAI message format. A message's content is either a string or a list of content parts (text, images). Pydantic validates the `Iterable`-typed content of the wrapper types into a lazy, single-use iterator.

**The bug:** that iterator cannot be deep-copied. The model cache deep-copies every load that is not readonly, so a bulk load of runs with content-part traces (the eval results page joining rows to traces) raised and failed the whole request. It also lost data silently: content parts on user and tool messages serialized as empty lists.

**The fix:** materialize the content at validation, so every later copy is safe and the parts persist.

libs/core/kiln_ai/datamodel/task_run.py
```diff
@@ -60,6 +62,7 @@ class TaskRun(
     KilnParentModel,
     parent_of={
         "feedback": Feedback,
+        "claim_reviews": ClaimReview,
```
chiang-daniel, line 65: **New child collection: a reviewer's claim grades, stored under the run they grade.**

**What this is:** a `TaskRun` is a parent of child records stored in subfolders beside it; `feedback` was the only one. `ClaimReview` joins it: the judge's claims about this run and the reviewer's agree or disagree on each (see `claim_review.py`).

**Why here:** as its own record under the run rather than fields on the `EvalRun`, re-running a judge never overwrites a human's grades, and judge-prompt refinement can read the full review for any run. The record itself is commented on `claim_review.py`.

libs/core/kiln_ai/datamodel/task.py
```diff
@@ -247,11 +247,14 @@ def runs(
         copy every run regardless.
         """
         runs = self._runs(readonly=readonly)  # type: ignore[attr-defined]
+        # Eval filter first: an excluded eval-generated run must not count as a parent
```
chiang-daniel, line 250: **Bug fix in `Task.runs()` filtering: an eval-generated run could hide a curated one.**

**What this is:** `Task.runs()` is the default in-process view of a task's runs. It applies two filters:
- drop eval-generated runs (`eval_source` set): eval byproducts, not curated data
- drop intermediate runs of a multi-turn chain (any run that is another run's `parent_task_run_id`), so only chain leaves show

**The bug:** the intermediate filter ran first. An eval-generated run chained onto a dataset run made that dataset run look like an intermediate, so it was dropped; the eval run was then dropped by the eval filter; the curated run vanished from the default view.

**The fix:** eval filter first, so an excluded run cannot make anything else disappear.

libs/core/kiln_ai/datamodel/code_tool.py
```diff
@@ -136,8 +137,20 @@ def validate_function_name(cls, v: str) -> str:
         return v
 
     @model_validator(mode="after")
-    def validate_parameters_schema(self) -> Self:
+    def validate_parameters_schema(self, info: ValidationInfo) -> Self:
         validate_schema_dict(self.parameters_schema, require_object=True)
+
```
chiang-daniel, line 142: **Bug fix from the review of the branch, not eval builder scope: a keyword-named parameter made a code tool uncallable.**

**What this is:** a `CodeTool` is a user-written Python `run()` function plus a JSON schema naming its parameters. When the model calls the tool, the sandbox calls `run()` with the arguments as keyword arguments keyed by the schema's names.

**The bug:** a parameter named after a Python keyword (`from`, `class`) cannot be a `run()` parameter, so such a tool could never be called. The web helper that renamed those parameters in the generated stub hid the problem rather than fixing it.

**The fix:** reject the name when the tool is authored. Files already on disk are exempt so existing projects open. Soft keywords (`match`, `case`, `type`, `_`) are legal identifiers and stay allowed.

libs/core/kiln_ai/datamodel/task_output.py
```diff
@@ -16,6 +16,15 @@
     from kiln_ai.datamodel.task import Task
 
 
+TASK_OUTPUT_SCHEMA_ERROR_PREFIX = (
```
chiang-daniel, line 19: **Refactor: one constant for a sentence three places depend on.**

**What this is:** when a task has an output schema and the model's JSON fails it, `TaskOutput` validation raises a `ValueError` beginning with this sentence. Retry classification in the adapters decides a run is worth retrying by matching that sentence, and the MCP adapter raises it too.

**What changed:** the string lives in one constant so the three cannot drift. No wording change.

libs/core/kiln_ai/datamodel/basemodel.py
```diff
@@ -767,14 +767,17 @@ def from_id_and_parent_path(
...
-        cls: Type[PT], ids: Set[str], parent_path: Path | None
+        cls: Type[PT], ids: Set[str], parent_path: Path | None, readonly: bool = False
     ) -> Dict[str, PT]:
         """
         Bulk equivalent of from_id_and_parent_path, much faster for large collections.
 
         It picks out the matching models from the directory only once. This avoids
         doing individual costly lookups that scan the whole directory in scenarios
         where we need to iterate over a large collection of models (e.g. bulk tagging).
+
```
chiang-daniel, line 778: **Passthrough of an existing flag: the bulk loader learns `readonly`.**

**What this is:** every model load goes through a shared cache. A normal load returns a deep copy the caller may mutate; a `readonly` load returns the cached instance, and any mutation raises. Single loads and the collection accessors (`task.runs(readonly=True)`) already took the flag. This bulk loader did not, so callers paid a deep copy per model.

**Why now:** the eval results page joins every row to its trace through this loader, and with content-part traces the deep copy itself crashed (`task_run.py:134`).

**Safety:** default `False`, so every existing caller is unchanged.

libs/core/kiln_ai/datamodel/copilot_models/questions.py
```diff
@@ -182,3 +182,6 @@ class RefineSpecApiOutput(BaseModel):
 
     new_proposed_spec_edits: list[NewProposedSpecEditApi]
     not_incorporated_feedback: str | None
+    # Model-suggested, filename-safe name for the eval, derived from the issue.
```
chiang-daniel, line 185: **Additive optional field: the copilot suggests an eval name.**

**What this is:** `RefineSpecApiOutput` is what the copilot returns when it refines a spec from the user's answers to its clarifying questions: the proposed edits, plus feedback it did not incorporate.

**What changed:** it now also carries a suggested eval name, which the builder offers as the default. `None` when the model gives none.

libs/core/kiln_ai/datamodel/tool_id.py
```diff
@@ -129,11 +129,8 @@ def _check_tool_id(id: str) -> str:
 
     # Code tools must have format: kiln_tool::code::<code_tool_id>
     if id.startswith(CODE_TOOL_ID_PREFIX):
-        ct_id = code_tool_id_from_tool_id(id)
-        if not ct_id:
-            raise ValueError(
-                f"Invalid code tool ID: {id}. Expected format: 'kiln_tool::code::<code_tool_id>'."
-            )
+        # Raises ValueError on malformed IDs; the extracted ID isn't needed here.
```
chiang-daniel, line 132: **Dead branch removed.**

**What this is:** tool ids are prefixed strings (`kiln_tool::code::<id>` for code tools) and `_check_tool_id` validates the shape.

**Why safe:** `code_tool_id_from_tool_id` already raises on a malformed id, so the check on its return value could never fire.

libs/core/kiln_ai/datamodel/eval.py
```diff
@@ -1207,41 +1350,6 @@ class Eval(KilnParentedModel, KilnParentModel, parent_of={"configs": EvalConfig}
         description="The output of the task run to evaluate. Can be final answer, full trace, or None for V2 evals.",
     )
 
-    @model_validator(mode="before")
```
chiang-daniel, line 1210: **Dead code removed: a migration with nothing left to migrate.**

**What this is:** an `Eval`'s `splits` field names, per split (test, train, val, golden), the tag or dataset filter that selects its items. Two older single-purpose fields are still migrated into it on load.

**Why safe:** this validator did the same for a third key, `eval_input_filter_id`, which no shipped build ever wrote, so no user's project file carries it.

libs/core/kiln_ai/datamodel/claim_review.py
```diff
@@ -0,0 +1,52 @@
...
+    human_grade: Literal["agree", "disagree"] = Field(
+        description="The human's grade on this claim."
+    )
+    human_feedback: str | None = Field(
+        default=None,
+        description="Optional plaintext reason for the grade.",
+    )
+
+
+class ClaimReview(KilnParentedModel):
```
chiang-daniel, line 27: **New record: `ClaimReview`, a reviewer's grades on the judge's claim card for one run. Needed because Step 5 asks the reviewer to agree or disagree per claim and the judge refiner reads those grades; an `EvalRun` holds only a score.**

Stored as a child of the `TaskRun` it grades (`task_run.py:65`). Fields: the judge's `judge_score` (pass or fail) and `judge_reasoning`; the `overview` the reviewer read; `claims`, each with its `text`, `human_grade` (agree or disagree) and optional `human_feedback`; and `human_verdict`, the reviewer's overall call, taken from their grade on the verdict claim when the card carried one, otherwise asked directly.

**Decisions made:**
- One direction on every grade: agree means the judge got that decision right, so the refiner never re-derives polarity from the claim wording.
- The card text is copied in (overview, claim text), so a stored review still reads on its own after the judge prompt changes.
- Re-running the judge never touches a review; a second review of the same run is a second record.
- No link to the `EvalRun` or `EvalConfig` that produced the card, so a review cannot be traced to the judge version it graded. Blocker or Later?

sfierro (approved):

leonardmq (approved):

PR #1780, opened 2026-09-16
Fix git sync CORS test by clearing KILN_FRONTEND_PORT env var

scosman:
## What does this PR do?

Fixes the `test_git_sync_error_response_has_cors_headers` to respect `KILN_FRONTEND_PORT` environment variable before running the test. 

This would fail if KILN_FRONTEND_PORT wasn't 5173 

## Checklists

- [x] Changes are isolated to test code only

https://claude.ai/code/session_01PfYDkMKz3KRJq6uHtk7GYm


scosman:
Pushed a follow-up commit (`2f902e4`) refining the fix: instead of `monkeypatch.delenv("KILN_FRONTEND_PORT", ...)` to force the app's default port, the test now derives the expected origin from `KILN_FRONTEND_PORT` the same way `make_app()` does:

```python
frontend_port = os.environ.get("KILN_FRONTEND_PORT", "5173")
origin = f"http://localhost:{frontend_port}"
```

This passes under any ambient value of that env var (no global state mutation needed) and stays focused on what the test actually verifies — that CORS headers survive `GitSyncMiddleware`'s short-circuit — rather than incidentally asserting the default port is `5173`.

---

PR #1782, opened 2026-09-16
[Eval V2 Kiln, review-only 2/7] Eval runner and core: driving, scoring, trace reuse, retries

chiang-daniel:
**Review-only.** Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with this area reverted to the baseline, so "Files changed" is `libs/core` minus the datamodel (#1778) and the synthetic user package (3/7): 26 source files (+1430 / −290) and 24 test files. Details are inline on each hunk.

## How it works

"Run Eval" on the results page builds one job per (item, judge, run config), skips the ones that already have a saved score, and runs the rest with a small retry budget. Single-turn items are generated and scored as before. A multi-turn case from the builder is driven as a fresh conversation for each run config (the synthetic user held constant, the agent varying), persisted once as a standalone run, and every judge of that case reuses it through the trace index instead of paying again. Judges read the conversation through one transcript formatter. Retry decisions for every model call in the product now go through one classifier.

```mermaid
flowchart LR
  Run["Run Eval / Compare judges"] --> Collect["eval_runner: collect jobs, dedupe saved scores, drop stale skips"]
  Collect --> Job{"item kind"}
  Job -- "single-turn" --> Gen["generate once, judge"]
  Job -- "multi-turn case" --> Drive["drive a conversation per run config (synthetic user held constant)"]
  Job -- "stored conversation" --> Stored["judge the stored trace as-is"]
  Drive --> Index["trace_index: one paid conversation per (case, run config), reused by every judge"]
  Gen --> Index
  Index --> Judge["v2 judges: LLM judge, code eval, tool-call check, ..."]
  Judge --> Fmt["eval_trace_formatter: the transcript every judge reads"]
  Retry["retry_classification + async_job_runner: what retries, how long to wait"] -.-> Drive
  Retry -.-> Judge
```

- **`eval_runner.py`**: job collection with recoverable skips that stop counting once the reason is fixed; the multi-turn re-drive; the conversation completeness check; the readiness preflight.
- **`trace_index.py`**: the reuse table; traces are vetted before reuse; a corrupt file is skipped, not fatal.
- **`eval_trace_formatter.py`**: every part of a message now renders (text, reasoning, tool calls, tool results by name, structured answers); the old formatter dropped some.
- **`retry_classification.py`** (new), **`errors.py`**, **`async_job_runner.py`**: one retry classifier for every pipeline; typed parse errors; exponential backoff with jitter.
- **`slow_operation.py`** (new): a log-only watchdog that replaced the removed timeouts.
- **Judges and helpers**: fail-loud tightenings (non-finite scores, unknown template variables, g_eval on a model without logprobs); a dead flag removed from the chat formatters.
- **Tools**: a broken code-tool allowlist entry no longer takes down every tool; timeouts are a structural flag, not a text match.
- **Small**: a model-list flag for the synthetic-user picker; CSV import rolls back a half-saved multi-turn import; a constant for the structured-answer wrapper.

## Why this PR is second

This is the engine every eval runs on, single-turn included, and the retry and formatter changes touch every model call. A wrong policy here changes scores and costs for existing users without any UI being involved.

## How to comment


libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -66,15 +94,128 @@ def _calibration_item(job: EvalJob) -> TaskRun | None:
...
+        return any(
+            isinstance(part, dict) and bool(str(part.get("text") or "").strip())
+            for part in content
+        )
+    return False
+
 
-    Multi-turn items are skipped by V2 evals, which also keeps eval traces childless —
-    a property `TraceIndex._seed` relies on to find them.
+def conversation_health_problem(
```
chiang-daniel, line 122: **New check: is this conversation complete enough to judge?**

**Why:** a conversation that stopped early, ended on the user's turn, or was recorded against an older turn ceiling would be scored as if it were the whole exchange. The check runs on every fresh drive before saving and on every reuse from the trace index; a failing conversation is not saved (fresh) or not reused (stored), and is driven again.

**The rule:** exactly the required number of user messages (or fewer when the leaf carries the ended-by-user tag), non-empty, ending on an assistant message with text. Shape only, not whether the agent erred. Example: a 5-turn case whose user stopped after 3 turns passes with the tag; without it, it reads as truncated and is driven again.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -66,15 +94,128 @@ def _calibration_item(job: EvalJob) -> TaskRun | None:
...
+    with no provider keys, and the two fields it needs are static entries in that table.
+    Custom, fine-tuned and litellm models are not in the table, so they keep the
+    credentialed lookup. A config that cannot be resolved answers False: the drive
+    fails on the identical lookup before it spends anything, and one unresolvable
+    config must not block the run configs beside it.
     """
-    if isinstance(item, TaskRun):
-        return item.parent_task_run_id is not None
-    return isinstance(item.data, MultiTurnSyntheticEvalInputData)
+    try:
```
chiang-daniel, line 192: **Preflight for run configs, and a gap: a config it cannot resolve passes.**

**Why:** some prompt styles spend two user-role messages per agent turn (two-message chain of thought), so such a run config can never produce a conversation of the right shape; the preflight rejects it before any paid call. Built-in models are read from the static table so this works with no provider keys; custom and fine-tuned models still need the credentialed lookup.

**The gap:** any exception during resolution is logged and answered `False` ("compatible"), so a run config whose model cannot be resolved (custom model, missing key) passes and then pays for conversations that fail the completeness check, up to three per item per run config (`eval_runner.py:1040`). Suggest answering "unknown" and letting the readiness check name the config.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -174,7 +317,45 @@ def __init__(
         # Live, not precomputed like `already_run`: a trace persisted by one job has to be
         # visible to the next, whether that next job is running concurrently under a
         # different eval config or is this job's own retry (functional spec 4.2, 4.3).
-        self._trace_index = TraceIndex(self.task)
+        self._trace_index = TraceIndex(self.task, vet=self._build_trace_vet())
```
chiang-daniel, line 320: **New: a stored conversation is checked against its case before it is reused.**

**Why:** a case's turn ceiling can change and a stored conversation can be truncated; reusing it would score a conversation that no longer matches the case. The trace index is handed a checker built from the eval's multi-turn cases (case id to turn ceiling); on seeding and on every serve a mismatch is rejected and a fresh conversation is driven. `None` when the split has no multi-turn cases, so single-turn evals behave exactly as before.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -199,16 +380,29 @@ def collect_tasks_for_eval_config_eval(
 
         # already_run[eval_config_id][dataset_id]
         already_run: Dict[ID_TYPE, Set[ID_TYPE]] = {}
+        superseded: Dict[Tuple[ID_TYPE, ID_TYPE], List[EvalRun]] = {}
         for eval_config in self.eval_configs:
             already_run[eval_config.id] = set()
             for run in eval_config.runs(readonly=True):
-                already_run[eval_config.id].add(run.dataset_id)
+                # Only calibration records mark a golden item done (or supersede
```
chiang-daniel, line 387: **Bug fix in judge calibration: a golden item scored by a run-config eval no longer counts as calibrated.**

**The bug:** the calibration collector treated any saved score on a golden item as "already calibrated", including scores written by the run-config comparison. A golden item that had been evaluated as a test item was skipped from calibration and the judge's accuracy was computed on fewer items than it looked.

**The fix:** only records with `eval_config_eval=True` count; the rest of the collector splits saved records into "already run" and "superseded skip" (`eval_runner.py:479`).

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -242,19 +441,33 @@ def collect_tasks_for_task_run_eval(self) -> List[EvalJob]:
                 # `str | None`, so a run config whose file carries a null id would make
                 # None a real key and fold every calibration record into its set.
                 if (
-                    run.task_run_config_id is not None
-                    and run.task_run_config_id in already_run[eval_config.id]
+                    run.task_run_config_id is None
```
chiang-daniel, line 444: **Bug fix in run-config collection, and the same superseded-skip split.**

**The bug:** a saved score record with no run config id matched every job's `(judge, run config, item)` key, so jobs were skipped as already scored. The membership test is now a guard that requires the id. Then the same split as calibration (`eval_runner.py:479`): saved scores dedupe; saved skips dedupe only while their reason still holds.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -263,6 +476,135 @@ def collect_tasks_for_task_run_eval(self) -> List[EvalJob]:
             not in already_run[eval_config.id][run_config.id]
         ]
 
+    def _counts_as_already_run(self, run: EvalRun, eval_config: EvalConfig) -> bool:
```
chiang-daniel, line 479: **Decision made: a recoverable skip stops blocking an item once its reason is fixed. Plus the readiness check.**

**Why:** an eval run that cannot score an item saves a skip record so the next run does not retry it forever; but three reasons are fixable by the user, and the old dedupe kept skipping the item after the fix.
- `missing_drive_config`: stops counting once the case carries a drive config
- `incompatible_input_shape`: once the case carries a first message
- `type_not_available`: once the judge type has an adapter

Anything else stays deduped. A stale skip is deleted after its replacement job succeeds (`eval_runner.py:664`). Each recoverable skip is checked against every case in the split; fine at today's sizes.

**Also here:** `validate_multi_turn_drive_readiness` raises one error naming every drive problem at once (no stamped case, unknown user-model provider, a run config that is not an agent config, a two-message chain-of-thought config), so "Run Eval" fails before spending. Called from the desktop's run endpoint.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -291,24 +633,54 @@ async def run(self, concurrency: int = 25) -> AsyncGenerator[Progress, None]:
     async def run_job(self, job: EvalJob) -> bool:
         try:
             if job.eval_config.config_type == EvalConfigType.v2:
-                return await self._run_v2_job(job)
+                done = await self._run_v2_job(job)
             else:
-                return await self._run_legacy_job(job)
+                done = await self._run_legacy_job(job)
```
chiang-daniel, line 638: **What changed in the job wrapper: retries are logged as warnings, stale skips deleted on success.**

A `RetryableError` is re-raised untouched for the job runner (no traceback in the log; a retried transient is not a failure); any other error still fails the job. `superseded_tombstones` are deleted only when the job completed, so a failed replacement leaves the old skip in place.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -291,24 +633,54 @@ async def run(self, concurrency: int = 25) -> AsyncGenerator[Progress, None]:
...
                 # the underlying provider detail for the developer-facing error log.
-                raise RetryableError(str(_unwrap_kiln_run_error(e))) from e
+                raise RetryableError(str(unwrap_kiln_run_error(e))) from e
             logger.error(
                 f"Error running eval job for dataset item {job.item.id}: {e}",
                 exc_info=True,
             )
             raise
 
+    async def _delete_superseded_tombstones(self, job: EvalJob) -> None:
```
chiang-daniel, line 664: **New: deleting a superseded skip record, best-effort. Needed so a fixed item's old skip does not linger beside its new score.**

Each stale skip is deleted inside the save context; a deletion failure is logged and swallowed, since the new score already exists and the stale skip is simply re-superseded next run.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -393,15 +765,52 @@ async def _run_v2_job(self, job: EvalJob) -> bool:
...
             )
 
-        # Both skips come before `_resolve_trace`, so a job that can never be scored
-        # never pays for a generation.
-        if _is_multi_turn(job.item):
-            return await self._persist_skip(
-                job,
-                SkippedReason.incompatible_input_shape,
-                "V2 evals do not yet support multi-turn inputs",
+        if (
```
chiang-daniel, line 768: **Decision made: a multi-turn case is driven fresh for every run config.**

**As before this branch:** V2 evals skipped every multi-turn item with "not yet supported".

**What changed:** a builder-minted multi-turn case (persona, goal, first message, drive settings) is driven as a new conversation for each run config under comparison: the agent varies, the synthetic user is held constant, so a score difference is the agent's. Real conversations, real spend, once per (case, run config); every judge of that case then reuses the stored conversation.

**Guard:** only in run-config comparison mode. Judge calibration is scoped by the golden filter, which only addresses dataset runs, so the other combination never occurs.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -393,15 +765,52 @@ async def _run_v2_job(self, job: EvalJob) -> bool:
...
+            # golden filter and golden filters only address TaskRuns. There is
+            # no runtime handler for a hand-built job of that shape.
+            seed = (
+                job.item.data.first_message.text if job.item.data.first_message else ""
+            )
+            return await self._run_v2_multi_turn_synthetic_job(
+                job, evaluator, job.item, job.item.data, seed
             )
 
+        if isinstance(job.item, TaskRun) and job.item.parent_task_run_id is not None:
```
chiang-daniel, line 786: **Decision made: a stored multi-turn conversation is judged as-is, so its scores are the same for every run config.**

**What this is:** a dataset conversation (a chain of runs, the leaf carrying the whole trace) cannot be regenerated in one model call and carries no persona to re-drive from. Both modes judge the stored trace; no generation, no eval-source stamp (so the item stays on dataset pages).

**Consequence to confirm:** comparing run configs over stored conversations yields identical numbers per run config. That is by design (the conversation already happened), but the compare page shows it like any other comparison. An empty trace is a `missing_trace` skip.

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -535,40 +944,211 @@ async def _persist_judgment(
...
+                    "answered a second message",
+                    eval_input.id,
+                )
+            leaf_trace = drive_result.chain[-1].trace if drive_result.chain else None
+            # Gated on the turns that actually ran, not the item's ceiling: the drive
+            # is still in hand, so its own chain is the exact count the leaf's trace
+            # has to match. A conversation the synthetic user ended is shorter than
+            # the ceiling and still complete, while one whose trace lost turns the
+            # chain says it ran is a broken record either way.
+            problem = conversation_health_problem(leaf_trace, turns_run)
```
chiang-daniel, line 1040: **Decision made, and a cost to confirm: an incomplete drive is retried, up to three paid conversations.**

After a fresh drive the completeness check runs on the turns that actually ran (a user-ended conversation is shorter and still complete). A failure raises `RetryableError` so the job runner drives again, because saving an incomplete conversation would make every judge of the item reuse it.

**The cost:** two retries. A deterministic failure (a run config that always ends on a tool message, or emits two user messages per turn) pays for three full conversations per item per run config, each logged as a warning, then fails. No "same problem twice, stop" guard; the readiness check (`eval_runner.py:479`) is meant to catch these first, but its preflight can pass an unresolvable config (`eval_runner.py:192`).

libs/core/kiln_ai/adapters/eval/eval_runner.py
```diff
@@ -535,40 +944,211 @@ async def _persist_judgment(
...
+                raise RetryableError(
+                    f"The driven conversation for eval item {eval_input.id} is "
+                    f"incomplete ({problem}), so it was not saved."
+                )
+            return await self._persist_driven_conversation(
+                key, drive_result, seed=seed, ended_by_su=ended_by_su
+            )
 
-    return False
+        # A raising drive persists nothing, so the job's retry re-drives; a
```
chiang-daniel, line 1053: **New: how a driven conversation is saved.**

**Why:** the runner pays for the conversation once and every judge reuses it, so what is stamped here is all a later judge, the results page and the compare page have to go on.

One standalone, childless `TaskRun`: input is the seed, the output source is re-stamped with the run config id, `synthetic_user_usage` carries the user model's spend, `eval_source` names the eval, and the ended-by-user tag is written when the drive stopped early. `_conversation_usage` sums the chain's tokens, cost and latency, `None` when nothing was measured. A leaf with no output source raises before anything is saved, which discards the paid drive: the cheap structural failure is permanent while the expensive one retries (`eval_runner.py:1040`).

libs/core/kiln_ai/adapters/eval/trace_index.py
```diff
@@ -71,10 +74,20 @@ class TraceIndex:
...
+    `vet` is the caller's veto over reuse: it decides whether a candidate record is fit to
+    serve a key, which is knowledge the index does not have (a conversation's completeness
+    depends on the item asking for it). It answers None for a fit record, or a reason
+    string for an unfit one — the reason travels with the rejection so the index's own log
+    line carries it, rather than stranding it in a separate line the caller emitted. A
+    `vet` of None reuses on key identity alone.
     """
 
-    def __init__(self, task: Task):
+    def __init__(
```
chiang-daniel, line 86: **New: the trace index checks a stored conversation before serving it.**

**Why:** the index is how a retry after a scoring failure re-scores instead of regenerating, and how several judges share one paid conversation; a stale entry served here would be scored by every judge as if it fit.

It maps (source kind, source id, run config) to the file holding that generation. The new `vet` callback is the caller's rule for whether a candidate fits (`eval_runner.py:320`). At seed time a rejected file is skipped, logged at info, and a later fit candidate can still serve the key; the rejected file stays on disk, so it is re-read and re-logged on every seed and nothing ever reaps it.

libs/core/kiln_ai/adapters/eval/trace_index.py
```diff
@@ -146,6 +176,18 @@ def _load_indexed(self, key: TraceKey) -> TaskRun | None:
             )
             del self._paths[key]
             return None
+        except (json.JSONDecodeError, ValidationError, ValueError) as error:
```
chiang-daniel, line 179: **Bug fix: a corrupt trace file no longer fails every job on that key.**

A truncated or schema-shifted file raised out of `load_from_file` and every job wanting that trace failed. Now the entry is dropped with a warning and the trace is regenerated.

libs/core/kiln_ai/adapters/eval/trace_index.py
```diff
@@ -162,8 +204,48 @@ def _load_indexed(self, key: TraceKey) -> TaskRun | None:
...
+                "Indexed eval trace at %s failed vetting for %s (%s); regenerating",
+                path,
+                key,
+                rejection,
+            )
+            del self._paths[key]
+            return None
         return trace
 
+    def _vet_rejection(self, key: TraceKey, run: TaskRun) -> str | None:
```
chiang-daniel, line 224: **Decision made: a vet that raises counts as a rejection.**

A broken predicate must cost a regeneration, not take down every job. The cost: on the multi-turn path a regeneration is a whole paid conversation per case per run config, and a systematically raising vet regenerates all of them with nothing counting how often; the drive-side retry budget is at `eval_runner.py:1040`.

libs/core/kiln_ai/adapters/eval/eval_utils/eval_trace_formatter.py
```diff
@@ -15,54 +18,97 @@ class MessageDetails:
...
+        while it calls a tool — so each is emitted as its own block. Collecting
+        blocks rather than assigning to shared variables is what keeps them: an
+        earlier design assigned and emitted once per message, so the last branch
+        taken silently replaced the others and a tool call accompanied by a
+        sentence never reached the judge.
+        """
+        blocks: list[str] = []
+        for message in trace:
             message_details = EvalTraceFormatter.message_details_from_message(message)
+            role = message_details.role
```
chiang-daniel, line 42: **Bug fix in the transcript every judge reads: a message no longer loses parts.**

`EvalTraceFormatter` renders a trace as labelled blocks (`<user_message>`, `<assistant_message>`, `<assistant_reasoning_message>`, `<assistant_requested_tool_calls>`, `<tool_tool_message>`) for judge prompts and the claim builder.

**The bug:** one block per message, assigned in sequence, so an assistant message with text and tool calls lost the tool calls, and a tool result whose originating call could not be found was dropped entirely.

**The fix:** every facet of a message is its own block; a tool result is labelled `tool result from <name>`, or bare `tool result` when the origin is unknown, now kept rather than dropped. The origin lookup still logs at `error` when it fails, which contradicts the "acceptable, proceed" posture two lines below; downgrade to warning.

**Consequence to confirm:** judge-visible text changed for traces with tools or reasoning, so an existing judge's scores on such traces can drift. The test that pinned the old lossy behaviour was deleted on purpose.

libs/core/kiln_ai/adapters/eval/eval_utils/eval_trace_formatter.py
```diff
@@ -135,15 +184,45 @@ def formatted_tool_calls_from_message(
         if tool_calls is None:
             return None
 
-        tool_calls_description = ""
+        # The task_response wrapper is not a tool the model chose to call, so it
```
chiang-daniel, line 187: **What changed: tool calls render one per block, and the structured-answer wrapper is not a tool call.**

Calls were concatenated with no separator (one call's arguments ran into the next call's name); now joined by a blank line. The synthetic `task_response` call that structured output rides on is filtered out here and rendered instead as the answer inside `<assistant_message>` (line 220). A malformed tool call (no `function` key) still raises a raw `KeyError` on the non-Jinja paths; unchanged by this branch, a follow-up.

libs/core/kiln_ai/adapters/eval/eval_utils/v2_eval_helpers.py
```diff
@@ -26,14 +26,24 @@ def build_binary_scores(
...
 
-# Matches the `trace` identifier as a whole word (e.g. `trace`, `trace[-1]`,
-# `trace | length`) without matching substrings like `retrace` or `tracer`.
-_TRACE_REFERENCE_RE = re.compile(r"\btrace\b")
-
-
 def references_trace(expression: str) -> bool:
-    """True if a Jinja expression references the `trace` variable."""
-    return bool(_TRACE_REFERENCE_RE.search(expression))
+    """True if a Jinja expression references the `trace` variable.
```
chiang-daniel, line 30: **Bug fix: "does this expression read the trace?" is now answered by the parser, not a regex.**

`references_trace` gates whether a judge needs the trace loaded. The old word-boundary regex matched `trace` as data (a quoted string, a dict key, another variable's attribute) and forced a trace load or a `missing_trace` skip. Now the expression is parsed with Jinja's own parser and its undeclared variables inspected; a syntax error answers `False` and is reported by the extraction path. Note it imports `_expression_env`, a private name from `jinja_engine`; if that module's internals move, this silently answers `False` and a `missing_trace` skip turns into `extraction_failed`.

libs/core/kiln_ai/adapters/eval/registry.py
```diff
@@ -34,6 +34,17 @@
 }
 
 
+def v2_eval_type_available(eval_config: EvalConfig) -> bool:
```
chiang-daniel, line 37: **New probe: is this judge type runnable on this build?**

**Why:** an eval run that hits a judge type with no adapter (a newer eval opened on an older build) saves a "type not available" skip per item. Those skips must stop counting once a build that has the adapter opens the eval, so job collection asks this question for every such skip on every run. It has to be answered from the registry alone, without constructing a judge, or collection would pay the judge's setup per skip.

`v2_eval_type_available` does exactly that (`eval_runner.py:479` is the caller).

libs/core/kiln_ai/adapters/eval/sandbox_worker.py
```diff
@@ -101,6 +101,20 @@ def execute_scorer_bridged(
                 "stderr": captured_stderr.getvalue(),
             }
         )
+    except SystemExit as e:
```
chiang-daniel, line 104: **Bug fix in code-eval scoring: `sys.exit()` in a user scorer no longer kills the child silently.**

`SystemExit` is not an `Exception`, so the broad handler missed it; the child died before queueing a result and the user saw a generic "exited without result". Now it is reported as "User code exited via sys.exit(...) instead of returning scores". `KeyboardInterrupt` still propagates on purpose.

libs/core/kiln_ai/adapters/eval/v2_eval_code_eval.py
```diff
@@ -137,11 +138,23 @@ def _validate_scores(self, raw: dict[str, Any]) -> EvalScores:
                     f"Score '{key}' returned a bool. Use a float (e.g. 1.0 for pass, 0.0 for fail)."
                 )
             if isinstance(value, int):
-                value = float(value)
+                try:
```
chiang-daniel, line 141: **Decision made: a code-eval score must be a finite number.**

An int too large for a float raised a bare `OverflowError`; `NaN` and infinities passed and corrupted the saved record. Both now fail in the scorer's own error surface with the offending value in the message.

libs/core/kiln_ai/adapters/eval/v2_eval_llm_judge.py
```diff
@@ -141,6 +152,17 @@ async def evaluate(self, eval_input: EvalTaskInput) -> V2EvalResult:
                 skipped_detail=f"Template rendering failed: {e}",
             )
         except UndefinedError as e:
+            # A typo'd template variable and genuinely missing reference data
```
chiang-daniel, line 155: **Bug fix: a typo in the judge prompt template is reported as a template error, not a missing reference key.**

Both raised Jinja's `UndefinedError` and both were labelled `missing_reference_key`, so an authoring mistake read as missing data. Now the template's top-level variable names are compared to the input's fields; unknown names skip as `extraction_failed` with the names listed.

libs/core/kiln_ai/adapters/eval/v2_eval_llm_judge.py
```diff
@@ -162,7 +184,18 @@ async def evaluate(self, eval_input: EvalTaskInput) -> V2EvalResult:
 
         if props.g_eval:
             model_provider = built_in_models_from_provider(provider, model_name)
-            if model_provider is not None and not model_provider.supports_logprobs:
+            if model_provider is None:
```
chiang-daniel, line 187: **Decision made: a G-Eval judge on a model outside the built-in table now fails before the call.**

**As before:** a model absent from the table (Ollama, custom, fine-tuned) fell through the logprobs check and the paid judge call failed deterministically.

**What changed:** it raises a `ValueError` with the fix in the message ("use a built-in model that supports logprobs, or disable G-Eval"). Not retryable, so the eval job fails. This is the one hunk here that can break an existing saved judge config: any G-Eval judge pointed at such a model stops running until it is edited.

libs/core/kiln_ai/adapters/eval/v2_eval_tool_call_check.py
```diff
@@ -74,17 +78,14 @@ def _check(
         expected_tools: list[ToolCallSpec],
         match_mode: str,
         on_unexpected: str,
-    ) -> tuple[bool, bool]:
-        """Check tool calls against expectations.
-
-        Returns (passed, unexpected_fail).
-        """
+    ) -> bool:
```
chiang-daniel, line 81: **Dead code removed, plus a malformed-call guard.**

`_check` computed and documented an `unexpected_fail` flag its only caller discarded; it now returns a bare bool with identical behaviour on all three match modes. A tool call whose `function` is present but `None` no longer raises (same guard in `eval_helpers.py`, where `five_star` also rejects `NaN`).

libs/core/kiln_ai/adapters/retry_classification.py
```diff
@@ -0,0 +1,96 @@
...
+    """The innermost non-wrapper error.
+
+    The model adapter wraps provider exceptions in KilnRunError (to carry the
+    partial trace), whose own message is genericized user-facing text — so both
+    retry classification and error detail must use the underlying error. The
+    isinstance guard on `original` keeps a (contract-violating) None from
+    escaping as the result."""
+    while isinstance(e, KilnRunError) and isinstance(e.original, BaseException):
+        e = e.original
+    return e
```
chiang-daniel, line 28: **New module: the one place that decides whether a failed model call is worth retrying.**

**Why:** three flows (the eval runner, the synthetic-user batch, the builder's pipeline) each decided this on their own, so the same provider hiccup was retried in one and fatal in another. Now there is one answer.

`is_retryable_error` is true for rate limits, connection errors, timeouts (`litellm.Timeout` listed separately because it does not inherit from litellm's connection error), provider 5xx, JSON-schema failures, wrong-shape structured output, schema-mismatch `ValueError`s carrying the shared prefix, and an empty assistant reply. Content-filter refusals share that raise site and are deliberately not retried. `is_batch_fatal_error` (auth, permission, not-found, budget) tells a batch owner to stop the whole batch.

**Fragility:** two arms match on message text (the schema prefix and the empty-reply message); six of the thirteen tests call the real raise sites, but a copy edit to either message still silently disables that retry.

**Decision made:** `NotFoundError` is batch-fatal. Some providers return 404 transiently; a transient 404 then aborts the whole batch and deletes its partial chains.

libs/core/kiln_ai/adapters/errors.py
```diff
@@ -50,6 +52,18 @@ def __init__(
         self.error_type = type(original).__name__
 
 
+class StructuredOutputParseError(ValueError):
```
chiang-daniel, line 55: **New error type, and a timeout message that tells the truth.**

**Why:** when a model answers with JSON of the wrong shape the run should be retried, not failed, but that failure raised a plain `ValueError`, indistinguishable from any other. And a request that timed out was reported to the user as "could not connect", sending them to check the wrong thing.

`StructuredOutputParseError` subclasses `ValueError`, so every existing handler and user message is unchanged; only the retry classifier looks for it. `format_error_message` gets a `litellm.Timeout` branch before the connection bucket (a timeout inherits from the connection error, so it fell into that message).

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -275,16 +273,26 @@ async def _run_returning_run_output(
                 if self.output_schema is not None:
                     # Parse json to dict if we have structured output
                     if isinstance(parsed_output.output, str):
-                        parsed_output.output = parse_json_string(parsed_output.output)
+                        try:
```
chiang-daniel, line 276: **What changed in every model call: parse failures are typed, and the strategy decision moved out.**

Unparseable or wrong-shape JSON raises `StructuredOutputParseError` on both the streaming and non-streaming paths (same messages; the MCP adapter gets the identical treatment), so the retry classifier can tell wrong shape from any other `ValueError`. `build_chat_formatter` delegates the strategy choice to `chat_strategy_for_run` (a pure function the eval preflight can call without keys) and resolves the provider only when a chain-of-thought prompt exists; a plain-prompt run no longer looks the provider up here. Confirm nothing relied on that early lookup failing.

libs/core/kiln_ai/adapters/model_adapters/base_adapter.py
```diff
@@ -128,14 +134,6 @@ class AdapterConfig:
     """
     automatic_prompt_caching: bool = False
 
-    """
```
chiang-daniel, line 131: **Dead code removed: `forward_thinking_instructions`.**

The flag, its threading, the formatter branch and a `DeprecationWarning` steering engineers toward it all survived after the only consumer was removed. Gone here and in `chat_formatter.py`; the R1 formatter now silently ignores `thinking_instructions` (its docstring says so, a test pins it). Confirm you are content with a factory that accepts an argument one branch ignores.

libs/core/kiln_ai/adapters/chat/chat_formatter.py
```diff
@@ -410,16 +373,57 @@ def get_chat_formatter(
...
-                user_input,
-                thinking_instructions,
-                forward_thinking_instructions=forward_thinking_instructions,
-            )
+            return SingleTurnR1ThinkingFormatter(system_message, user_input)
         case _:
             raise_exhaustive_enum_error(strategy)
 
 
+def chat_strategy_for_run(
```
chiang-daniel, line 381: **New: the "which prompt strategy will this run config use" decision, pulled out of the model adapter as a pure function.**

**Why:** before a multi-turn eval spends money, "Run Eval" has to reject run configs whose prompt style spends two user-role messages per agent turn (chain of thought sent as its own message), because those conversations can never have the right shape. That decision lived inside the model adapter, which cannot be built without provider credentials. Pulled out, the preflight (`eval_runner.py:192`) answers the question from the run config alone: no adapter, no keys, no call.

**What moved:** `chat_strategy_for_run(cot_prompt, tuned_strategy, reasoning_capable)` is the exact logic the adapter used; `is_two_message_cot_strategy` says whether a strategy is one of the two-message kinds. Both side-effect free; the adapter now calls the first.

libs/core/kiln_ai/adapters/ml_model_list.py
```diff
@@ -374,6 +375,7 @@ class KilnModelProvider(BaseModel):
     tuned_chat_strategy: ChatStrategy | None = None
     supports_doc_extraction: bool = False
     suggested_for_doc_extraction: bool = False
+    suggested_for_synthetic_user: bool = False
```
chiang-daniel, line 378: **New flag: which models the synthetic-user picker suggests. Needed because the builder's user-model picker had no way to rank models that are fast, cheap and read a whole conversation each turn.**

`suggested_for_synthetic_user` is one boolean read by the provider API; it is set on 16 provider entries (the 16 one-line data hunks below). Which 16 is a curation call for the model-list owner.

libs/core/kiln_ai/tools/base_tool.py
```diff
@@ -48,6 +48,10 @@ class ToolCallResult(BaseModel):
         default=None,
         description="Human-readable error message, set when is_error is True.",
     )
+    timed_out: bool = Field(
```
chiang-daniel, line 51: **New field: a tool timeout is a flag, not a substring. Needed because the sandbox bridge classified nested-call timeouts by matching "timed out" in the result text, so any failure whose message mentioned a timeout was misclassified.**

`ToolCallResult.timed_out` is set by the code tool when its execution limit fires; the bridge reads it.

libs/core/kiln_ai/tools/sandbox_bridge.py
```diff
@@ -401,7 +398,17 @@ async def name_map(self) -> dict[str, list[ToolId]]:
 
         name_map: dict[str, list[ToolId]] = {}
         for tool_id in self._allowlist:
-            fn_name = await self._canonical_tool_name(tool_id)
+            # A broken allowlist entry (deleted tool, bad config) must not take
```
chiang-daniel, line 401: **Bug fix, and a contract change to confirm: one broken allowlist entry no longer disables every tool.**

**The bug:** building the dispatch map resolved every allowlisted tool; a deleted RAG config threw and took the whole nested-tool surface down.

**The fix:** a failing entry is logged and skipped from the dispatch map. In `tools_info` (line 448) it is listed under its raw id with `(unavailable: <reason>)` instead of re-running the failing resolution.

**Inconsistency:** the two surfaces now disagree: the model sees the entry in `list_tools()` but calling it produces a dispatch error, not the unavailable reason it was shown. The raw exception text also goes into a description the model reads.

libs/core/kiln_ai/utils/async_job_runner.py
```diff
@@ -1,12 +1,36 @@
...
 import logging
+import random
 from dataclasses import dataclass
 from typing import AsyncGenerator, Awaitable, Callable, Generic, List, TypeVar
 
 logger = logging.getLogger(__name__)
 
 T = TypeVar("T")
 
+# Each retry waits in a window this many times wider than the last, so a short
```
chiang-daniel, line 11: **Decision made: retries wait with exponential backoff and full jitter.**

Each retry draws uniformly from a window four times wider than the last (0 to 1 s, then 0 to 4 s, ...), so callers that failed together do not retry in lockstep. Attempt counts are unchanged.

**Policy constant:** `RETRY_BACKOFF_FACTOR = 4.0` is uncapped; the comment says raising `max_retries` beyond 2 or 3 means lowering it or adding a ceiling. That tripwire lives in a comment, not in code (at 5 retries the last window is 0 to 1024 s).

libs/core/kiln_ai/utils/slow_operation.py
```diff
@@ -0,0 +1,50 @@
...
+import contextlib
+import logging
+from typing import AsyncIterator
+
+logger = logging.getLogger(__name__)
+
+# Well beyond a healthy conversation drive or single task run, so a firing
+# means something worth investigating (a trickling stream, a stuck tool
+# loop), not a slow-but-fine case. Logging only; nothing is cancelled.
+DEFAULT_SLOW_LOG_THRESHOLD_SECONDS = 600.0
```
chiang-daniel, line 18: **Decision made: a log line replaced the timeouts.**

**What this is:** `log_if_slow` warns once when the wrapped block is still running after 600 s and never cancels it. It wraps the batch drive, the single-turn pipeline run and the eval-time re-drive. The per-turn drive budget and the single-turn run budget were removed in the same change because every firing they produced killed a healthy run.

**What remains:** the turn ceiling, the tool-call cap, and the model client's per-request timeout. A provider that trickles bytes on a non-streaming body is bounded by nothing but that log line.

**Note:** the `finally` uses `gather(return_exceptions=True)` rather than suppressing `CancelledError`, so a cancellation aimed at the enclosing task still propagates and a cancelled batch stops spending.

libs/core/kiln_ai/utils/jinja_engine.py
```diff
@@ -50,19 +50,62 @@ def _fromjson(value: object) -> Any:
         raise JinjaExtractionError(f"value is not valid JSON: {e}") from e
 
 
+def _format_trace(value: Any) -> str:
```
chiang-daniel, line 53: **New filter: `format_trace`, the transcript block baked into judge prompts. Needed so a judge template can write `{{ trace | format_trace }}` and get the same transcript the claim builder sees.**

Renders a trace through the formatter (`eval_trace_formatter.py:42`). A non-list, or a non-dict member (reported with its index and type), raises an extraction error; any formatter exception is also converted to a per-item `extraction_failed` skip, so a formatter regression shows as a wave of skips, not a crash.

libs/core/kiln_ai/utils/open_ai_types.py
```diff
@@ -189,17 +196,20 @@ def _materialize_lazy_content(trace: Iterable[Any]) -> None:
...
     ``content`` as ``str | None`` (the function role), so no tighter signature type
     checks, and ``typing.cast`` is banned repo-wide. The guard below still means only
     a message actually holding a lazy iterator is ever written to.
 
-    One side effect, not a defect: a lazy ``content`` matched the union's ``generator``
-    branch, so ``TaskRun.model_dump_json`` emitted no warning for it. A materialized
-    ``list`` matches neither ``str`` nor ``generator``, so each later dump of the same
-    object emits one ``PydanticSerializationUnexpectedValue`` warning. The bytes
-    written are unchanged - this is stderr noise only.
+    Now that TaskRun runs this at validation, it also changes what ``save_to_file``
```
chiang-daniel, line 205: **Docstring, pointing at a persistence change made elsewhere.**

`materialize_lazy_content` is now public and runs at `TaskRun` validation (the datamodel view, 1/7), which means content parts on user and tool messages survive `save_to_file` where they previously serialized as empty lists. This docstring is the only place in this area that says so. `TASK_RESPONSE_TOOL_NAME` (line 38) names the structured-answer wrapper call; the same literal exists in two TypeScript files with no cross-language pin.

libs/core/kiln_ai/utils/dataset_import.py
```diff
@@ -264,13 +264,40 @@ def import_csv(
...
+    All rows are validated before any are persisted to files, and a save
+    failure rolls back the runs already saved (best effort), so a failed
+    import leaves no partial data behind."""
 
     if task.turn_mode == TurnMode.multiturn:
         return _import_csv_multiturn(task, config)
     return _import_csv_single_turn(task, config)
 
 
+def _save_runs_with_rollback(runs: list[TaskRun]) -> None:
```
chiang-daniel, line 276: **Bug fix in CSV import: a failed multi-turn import no longer leaves a truncated conversation behind.**

**The bug:** rows were saved in a loop; an I/O failure mid-chain left a partial conversation on disk posing as a complete one, while the docstring claimed no partial imports.

**The fix:** `_save_runs_with_rollback` saves in order and, on failure, deletes every run it saved before re-raising.

**Check:** the rollback deletes in save order, parents first. If a child run's file lives under its parent's directory, the later child deletes fail and are logged as cleanup failures on exactly the case this exists for; deleting in reverse order avoids it. Not verified against `TaskRun.delete()`.

sfierro (approved):

leonardmq (approved):

PR #1784, opened 2026-09-16
[Eval V2 Kiln, review-only 3/7] Synthetic user: personas, conversation driving, early stop

chiang-daniel:
**Review-only.** Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with `libs/core/kiln_ai/synthetic_user` removed, so "Files changed" is exactly that package: 10 source files (+1519) and 9 test files, all new. Details are inline on each hunk.

## How it works

The synthetic user is the customer side of a multi-turn conversation. Step 4 gets a persona, a goal, behaviour guidance and a first message per case from the server; this package plays that user locally, with the user's own provider keys. Each turn: the agent under test answers, then a second model call, given the conversation with the roles swapped and a system prompt built from the persona, writes the user's next message. That repeats until the case's turn count, which is a ceiling, or until the user replies with a single control token meaning "my goal is met". Each agent turn is an ordinary `TaskRun` chained to the previous one; the last one is tagged so Step 5 and the eval runner can find the conversation. At eval time the same case is driven again in memory for each run config, with the synthetic user held constant so a comparison varies only the agent.

```mermaid
flowchart LR
  Case["case: persona, goal, guidance, first message"] --> Loop
  subgraph Loop["drive_case: one conversation"]
    T["agent turn (TaskRun, chained)"] --> Q{"turn < ceiling?"}
    Q -- yes --> SU["synthetic user reply (role-swapped view + persona prompt)"]
    SU -- "next message" --> T
    SU -- "'<DONE>'" --> End["stop early, tag the leaf"]
    Q -- no --> End
  end
  Batch["runner.py: Step 4 batch (queue of events, retries, cleanup)"] --> Loop
  Eval["eval_drive.py: eval-time re-drive per run config (in memory)"] --> Loop
```

- **`drive_loop.py`**: the loop above; the ceiling, the stop token, spend accounting.
- **`prompt.py`**: the system prompt the user plays from: persona, goal, guidance, then the house rules (conversation style, never invent records, accept a not-found reply, when it may end).
- **`driver.py`, `role_swap.py`**: one model call per user turn; the agent's turns become `user` messages and the user's become `assistant` so a chat model writes the user side; tool-only agent turns are hidden.
- **`runner.py`**: Step 4's batch: concurrency, retries on transient errors, per-case events for the live stream, tagging, cleanup of a failed case's partial chain, honest spend accounting across attempts.
- **`eval_drive.py`**: the eval-time re-drive; nothing touches disk, the eval runner persists one standalone `TaskRun`.
- **`models.py`, `parser.py`, `case.py`**: the stop token and the on-disk tag, the persona blob codec, the wire model.

## Why this PR is where it is

Every multi-turn conversation in the product, at Step 4 and at eval time, comes out of this loop. Its policies (a ceiling not a count, no timeouts, what counts as a finished conversation) shape what the judge sees. Nothing here is reachable from single-turn evals.

## How to comment

Mark **Blocker** or **Later**. Blockers are fixed on `dchiang/eb-v2-merge` in one round and this view is regenerated. Inline markers: **Bug** = fix it; **Decision made** = built this way on purpose, say yes or push back; anything else is a note. Prompt wording is the prompt owner's lane; the comments name what the prompt does, not how to word it.


libs/core/kiln_ai/synthetic_user/drive_loop.py
```diff
@@ -0,0 +1,208 @@
+"""Single-case drive loop for the multi-turn synthetic-user runner.
+
+`drive_case` alternates the target task adapter with the local
+`SyntheticUserDriver`, producing a chain of `TaskRun`s on the target side.
+`turns` is a ceiling: the loop stops early when the SU driver replies with
```
chiang-daniel, line 5: **Decision made: the case's turn count is a ceiling, not a length.**

**Why:** a synthetic user who already got what they came for and is made to keep talking produces filler, and the judge then scores the filler. Ending when the goal is met keeps the conversation honest; the count only bounds it.

**What this is:** `drive_case` runs one conversation: agent turn, then the synthetic user's reply, up to `turns` agent turns. A case minted with `turns=5` can end after 2, 3 or 4 when the user's goal is met (`drive_loop.py:187`). A conversation never runs longer than its ceiling.

**What a reviewer sees:** the same case driven under two run configs can come back with different lengths, so a score gap between two agents may be a shorter conversation rather than a worse agent. Nothing downstream normalises for length; the judge reads whatever was said. Say if you want a minimum length or a length note on the results page.

**Pinned by:** `test_drive_case_stops_when_su_returns_sentinel` and its siblings.

libs/core/kiln_ai/synthetic_user/drive_loop.py
```diff
@@ -0,0 +1,208 @@
...
+        chain.append(new_run)
+
+        # The SU only speaks when another target turn will consume its
+        # reply — an SU call after the final turn would be paid LLM output
+        # nobody reads. The SU driver does the role filtering / role swap /
+        # invariant checks itself; we pass the cumulative trace as-is.
+        su_message: str | None = None
+        su_cost = 0.0
+        su_ended_conversation = False
+        if turn < turns:
```
chiang-daniel, line 176: **Decision made: the synthetic user is only asked when another agent turn will read the reply.**

On the final allowed turn the loop does not call the user model: that reply would be paid output nobody reads. So a 5-turn conversation makes 5 agent calls and at most 4 user calls; a 1-turn case never calls the user model at all (it is a single-turn run with a persona attached). The per-turn hook still fires for the last turn with `su_message=None`.

libs/core/kiln_ai/synthetic_user/drive_loop.py
```diff
@@ -0,0 +1,208 @@
...
+            if turn_usage is not None:
+                # Usage.__add__ is None-graceful per field, so a turn that
+                # reports only cost doesn't erase another turn's token counts.
+                su_usage = turn_usage if su_usage is None else su_usage + turn_usage
+                if turn_usage.cost is not None:
+                    su_cost = float(turn_usage.cost)
+            # A reply that is nothing but the sentinel is the SU saying it is
+            # done. Whole-message match (after trimming the whitespace models
+            # add) so a user who quotes the sentinel mid-sentence still gets a
+            # real turn. The sentinel itself is dropped here: it is a control
```
chiang-daniel, line 187: **Decision made: how the user ends a conversation, and one gap.**

**What this is:** the user's reply is compared, after trimming whitespace, against the stop token `<DONE>` (defined in `models.py`). An exact match ends the case; the token is dropped, never sent to the agent and never written into a trace. A reply that mentions the token mid-sentence is an ordinary message.

**The gap:** an empty reply from the user model is not caught here. The seed message is guarded (`drive_loop.py:150`: an empty seed raises "so it does not surface as a confusing model-side error"), but a later empty reply is handed to the agent as the next user message unchanged. It comes from a model, so it is the likelier source of an empty string. Cheap to guard the same way, or to treat as a stop.

libs/core/kiln_ai/synthetic_user/drive_loop.py
```diff
@@ -0,0 +1,208 @@
...
+    against no invoice and split per model not at all.
+
+    None when no turn reported usage, rather than a zeroed Usage: an
+    unmeasured drive must not read as a genuinely free one.
+    """
+
+    chain: list[TaskRun]
+    su_usage: Usage | None
+
+    def ended_early(self, turns: int) -> bool:
```
chiang-daniel, line 91: **Decision made: "ended early" is inferred from the chain length; there is no stop-reason field.**

`ended_early(turns)` is `len(chain) < turns`. The docstring's argument: every other way out of the loop raises, so a short chain can only mean the user stopped it. That holds for the code as written, and the caller must pass the same `turns` the drive ran with (both callers do). The on-disk record is a tag on the leaf (`models.py:23`), not a field. Fine as long as no future exit path returns a partial chain without raising.

libs/core/kiln_ai/synthetic_user/models.py
```diff
@@ -0,0 +1,49 @@
...
+from kiln_ai.datamodel.datamodel_enums import ModelProviderName
+from kiln_ai.datamodel.eval import SyntheticUserInfo as SyntheticUserInfo
+
+VisibleMessageRole = Literal["user", "assistant"]
+
+# The exact message the SU sends, alone and with nothing else, to end a
+# conversation before the turn ceiling is reached. It lives in this leaf module
+# so the drive loop that acts on it and the prompt that teaches it can share
+# one string without importing each other.
+EARLY_STOP_SENTINEL = "<DONE>"
```
chiang-daniel, line 23: **New constants: the stop token and its on-disk counterpart.**

- `EARLY_STOP_SENTINEL = "<DONE>"`: what the user model says, alone, to end a conversation. Lives in this leaf module so the prompt that teaches it and the loop that acts on it share one string.
- `TAG_SU_ENDED_CONVERSATION`: the tag written on the last run of a conversation the user ended. The eval runner's completeness check reads it to tell a deliberately short conversation from a truncated one (a drive that died mid-way) long after the drive is gone.

`SyntheticUserDriverConfig` below carries the user model and provider and deliberately no temperature (the persona and guidance carry style).

libs/core/kiln_ai/synthetic_user/prompt.py
```diff
@@ -0,0 +1,70 @@
...
+
+_OPENING = (
+    "You are playing the role of a user interacting with an AI assistant. "
+    "Stay in character — respond as the user would, not as the assistant. "
+    "Your entire output is the user's next message, verbatim and nothing "
+    "else: no narration, no meta-commentary, no quotes, no labels like "
+    '"User:". Just the message text the user would type.'
+)
+
+_CONVENTIONS = f"""## Conversation style
```
chiang-daniel, line 44: **Decision made: the house rules every synthetic user plays by. This is the whole behavioural contract.**

**What this is:** the system prompt is persona, goal, optional guidance, then this block. Read it in full; it is short.

**Three rules a reviewer should know:**
- Never invent identifiers, records, names or dates; ask the assistant what exists or to create it. When the assistant says something does not exist, that is the truth: do not insist, even if the guidance says press. These two outrank persona and guidance.
- When it may end: only after the assistant has replied at least twice, only when the goal is fully and specifically met, never after a refusal, deflection, partial answer, question or mistake, and only by replying with the stop token alone. Guidance asking for more wins over ending.
- Stay engaged and vary approach when the assistant is vague, evasive or wrong (the line just above the two world-state rules; the docstring explains the ordering resolves the tension).

**Consequences:** with `turns=2` the user is asked once, after the assistant's first message, where ending is forbidden, so a 2-turn case always runs both turns. In a captured run about half the cases stopped early. The block is pinned verbatim by a test, so any wording change edits the test too.

**Whose call:** prompt wording is the prompt owner's; what to confirm here is the policy, not the sentences.

libs/core/kiln_ai/synthetic_user/driver.py
```diff
@@ -0,0 +1,146 @@
...
+)
+from kiln_ai.synthetic_user.prompt import render_system_prompt
+from kiln_ai.synthetic_user.role_swap import role_swap
+from kiln_ai.utils.open_ai_types import (
+    ChatCompletionMessageParam,
+    ChatCompletionSystemMessageParam,
+)
+
+
+def _is_tool_dispatch_only(msg: ChatCompletionMessageParam) -> bool:
```
chiang-daniel, line 30: **Guard: agent turns that are only tool calls are hidden from the user model.**

**What this is:** the agent's trace can contain assistant turns with no text, only tool calls (a lookup, then a second turn with the answer). After the role swap those would become blank user messages. `content` is checked falsy, not `is None`, because the adapter emits `''` as well as `None` for such turns.

**Edge to confirm:** `respond` then requires the last visible turn to be an assistant turn (line 115). An agent turn that ends on a tool-only dispatch with no final text is filtered out, the invariant fails, and the case dies as `unexpected_error`. Whether an agent can end a turn that way depends on the tool-call cap and the model; not seen in captured runs, not tested.

libs/core/kiln_ai/synthetic_user/driver.py
```diff
@@ -0,0 +1,146 @@
...
+        self._run_config = KilnAgentRunConfigProperties(
+            model_name=driver_config.model_name,
+            model_provider_name=driver_config.model_provider_name,
+            prompt_id="simple_prompt_builder",
+            structured_output_mode=StructuredOutputMode.default,
+            tools_config=ToolsRunConfig(tools=[]),
+        )
+        self._adapter = adapter_for_task(self._task, self._run_config)
+
+    async def respond(
```
chiang-daniel, line 86: **Decision made: the user model's spend is returned as a full `Usage`, or `None` when unmeasured.**

The user's turns are never persisted, so this return value is the only place the user model's tokens exist. A full `Usage` can be split per model against an invoice and repriced; a cost alone cannot. `None` means the provider reported nothing, distinct from a zeroed `Usage` that would read as a free call. The loop sums it per conversation; the datamodel stores it beside the agent's spend.

libs/core/kiln_ai/synthetic_user/role_swap.py
```diff
@@ -0,0 +1,61 @@
...
+"""
+
+from kiln_ai.utils.open_ai_types import (
+    ChatCompletionAssistantMessageParamWrapper,
+    ChatCompletionMessageParam,
+    ChatCompletionUserMessageParam,
+)
+
+
+def role_swap(
```
chiang-daniel, line 23: **Guard: the role swap fails loud on anything but plain text user and assistant turns.**

Eval-frame `user` becomes `assistant` and vice versa so a chat model writes the user's side. A system or tool role, or non-string content, raises. The driver filters roles first, so a raise here means an invariant broke upstream.

**Edge:** the driver only drops assistant turns with falsy content; an assistant turn carrying list content (content parts) would pass the filter and raise here, failing the case. Whether Kiln's adapters emit list content on the agent side is not established.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+    AsyncJobRunnerObserver,
+    RetryableError,
+)
+from kiln_ai.utils.git_sync_protocols import SaveContext, default_save_context
+from kiln_ai.utils.open_ai_types import ChatCompletionMessageParam
+from kiln_ai.utils.slow_operation import log_if_slow
+
+logger = logging.getLogger(__name__)
+
+# Module constants.
```
chiang-daniel, line 49: **Decision made: the batch limits, and which of them this module enforces.**

- `NUM_CASES_MAX = 200`: the largest batch. Not enforced here; the desktop routes validate against it (four `Field` bounds mirror it). 200 has not been driven live.
- `MAX_TURNS_DEFAULT = 5`, `CONCURRENCY = 4`: no rationale in code. The builder overrides concurrency with its own 10; 4 applies only to the standalone multi-turn generation route.
- `DRIVE_MAX_RETRIES = 2` on transient errors only (same posture as the eval runner), with exponential backoff and full jitter from a 1 s base.

**What to check:** whether you want the runner itself to refuse a batch over the cap, and whether one concurrency value should serve both callers.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+    case_timeout_seconds: float | None = None,
+) -> AsyncIterator[BatchEvent]:
+    """Drive `cases` concurrently against `target_task`, streaming progress.
+
+    Cases run through an AsyncJobRunner worker pool bounded by `concurrency`,
+    so the target-task LLM and the SU LLM aren't both hammered at unbounded
+    fan-out. Events from different cases interleave; ordering WITHIN a case
+    is `turn_completed`* → `case_completed | case_failed`.
+
+    `case_timeout_seconds` optionally bounds each case's drive; a case that
```
chiang-daniel, line 175: **Decision made: no timeout on a conversation by default.**

`case_timeout_seconds` defaults to `None`. Termination is structural: the turn ceiling, the per-turn tool-call cap, and the model client's per-request timeout. A slow case is logged by a 600 s watchdog, not killed. An earlier per-turn budget was removed because every firing it produced was a false positive on a healthy run. A caller can still set a budget, and then a timeout fails the case without retry.

**Trade-off:** a provider that trickles bytes forever is bounded only by the client's between-bytes timeout; the log line is the only signal.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+                    target_invoker=_target_invoker,
+                    su_driver=su_driver,
+                    turns=turns,
+                    on_turn=_on_turn,
+                ),
+                timeout=case_timeout_seconds,
+            )
+
+        ended_by_su = result.ended_early(turns)
+        if len(result.chain) == 1 and turns > 1:
```
chiang-daniel, line 473: **Decision made: a one-turn "multi-turn" conversation is logged, not rejected.**

A case with `turns > 1` whose user stopped after the agent's first reply (the prompt forbids it, but a model can) is a single-turn trace wearing a multi-turn label. It is warned about, then tagged, persisted, shown in Step 5 and judged like any other. Killing a paid drive over a rule the prompt states but cannot guarantee was judged to cost more than it saves. The eval runner mirrors this for re-drives. Say if you would rather reject it.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+                chain_run_ids=[
+                    str(r.id) if r.id is not None else "" for r in result.chain
+                ],
+                leaf_run_id=str(leaf.id) if leaf.id is not None else "",
+                total_turns=len(result.chain),
+                total_cost=_cumulative_cost(leaf) + result.su_total_cost,
+                discarded_attempts_cost=failed_attempt_spend.get(case_index, 0.0),
+            )
+        )
+    except _CaseFailure:
```
chiang-daniel, line 505: **Note: the pass-through failure arm skips spend banking and cleanup.**

`except _CaseFailure: raise` exists for the persona-blob parse failure, which happens before anything is persisted, so there is nothing to bank or delete. That is true today. Any future `_CaseFailure` raised after a turn persists would leave an untagged orphan chain on disk and lose that attempt's spend. Cheap to make the arm bank and clean like the others.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+        # Raised above before anything persisted (parse) — pass through.
+        raise
+    except asyncio.CancelledError:
```
chiang-daniel, line 508: **Guard: stopping a batch does not leave half-driven conversations on disk.**

Stop cancels in-flight cases mid-conversation. Their persisted turns would otherwise remain as untagged orphans (a chain is only discoverable through the leaf's tag, which is applied after a successful drive). The arm banks the attempt's spend, deletes the partial chain under `asyncio.shield` so the cancellation cannot interrupt the delete mid-chain, then re-raises.

**Note:** the banked spend of a cancelled case never reaches an event (the batch is ending), so "what did the stopped batch cost" is not answerable from the stream.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+        if is_retryable_error(e):
+            raise RetryableError(f"{type(cause).__name__}: {cause}") from e
+        raise _CaseFailure(
+            "unexpected_error",
+            f"{type(cause).__name__}: {cause}",
+            type(cause).__name__,
+        ) from e
+
+
+def _bank_attempt_spend(
```
chiang-daniel, line 579: **Decision made: a failed attempt's real spend is counted, then its chain is deleted.**

A transient failure retries the whole case (up to two retries). Each failed attempt's agent spend (the deepest persisted run's cumulative usage) plus its user-model spend is banked per case before the chain is deleted, and reported as `discarded_attempts_cost` on the completed event or folded into the failed event's total. Without this, a retried case would under-report by the cost of every attempt that was thrown away.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+    """
+    target_cost = 0.0
+    if persisted_runs:
+        target_cost = _cumulative_cost(next(reversed(persisted_runs.values())))
+    failed_attempt_spend[case_index] = (
+        failed_attempt_spend.get(case_index, 0.0) + target_cost + attempt_su_cost
+    )
+
+
+async def _delete_partial_chain(
```
chiang-daniel, line 600: **Decision made: chain cleanup is best-effort and never raises.**

Deletes newest-first so an interrupted cleanup cannot orphan a child before its parent. A failure to delete is logged with a traceback and swallowed: the terminal failure the caller is about to raise is the event that matters. Consequence: a failed cleanup leaves orphan runs with no user-visible signal. Confirm the trade.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+
+
+def _new_batch_tag() -> str:
+    """12-char hex tag from uuid4 — short enough to read; long enough to
+    avoid collisions across batches a user runs in the same session.
+    """
+    return uuid.uuid4().hex[:12]
+
+
+def _cumulative_cost(run: TaskRun) -> float:
```
chiang-daniel, line 736: **Note: a cost reader shaped by tests, lax in production.**

`_cumulative_cost` reads the run's rolled-up cost through two `getattr` calls defaulting to `0.0`, "defensive against fakes in unit tests". A real `TaskRun` always has the attribute, so in production this turns a genuinely missing cost into zero, which is the silent-zero the rest of the package argues against (`driver.py:86`). Suggest reading the field directly and letting a missing cost be `None`.

libs/core/kiln_ai/synthetic_user/runner.py
```diff
@@ -0,0 +1,770 @@
...
+# ───────────────────────── BatchEvent dataclasses ─────────────────────────
+
+
+@dataclass(frozen=True)
+class BatchStartedEvent:
+    batch_tag: str
+    num_cases: int
+
+
+@dataclass(frozen=True)
```
chiang-daniel, line 85: **Event contract for the live Step 4 stream, and two things to know.**

`TurnCompletedEvent` carries the turn index (1-based within the current attempt; a retried case restarts at 1), the agent run id, the user's next message, the cumulative cost, and a snapshot of the full trace so far. `CaseCompletedEvent` carries the chain, the leaf, the turn count, the cost, and the discarded attempts' cost. `CaseFailedEvent` carries a code, a message, the total across attempts, and the exception class name.

- `su_next_message=None` means either "final turn, no user call" or "the user ended the conversation"; the stream cannot tell the two apart. The UI shows the tag on the leaf afterwards.
- Every turn event ships the whole trace so far, so a 20-turn case sends a quadratic amount of data; fine at today's sizes, worth knowing at the 200-case cap.

libs/core/kiln_ai/synthetic_user/eval_drive.py
```diff
@@ -0,0 +1,88 @@
+"""Eval-time re-drive of one multi-turn synthetic case.
```
chiang-daniel, line 1: **Decision made: at eval time the conversation is driven again per run config, with the user held constant.**

The agent comes from the run config under test; the synthetic user comes from the case's stamped drive config (model, provider, turns). Nothing touches disk: the drive runs with saving off, and the eval runner persists one standalone `TaskRun` afterwards. Continuity rides the prior trace; no parent chaining. One drive per (case, run config) is shared by every judge through the trace index.

**Two things to know:** there is no per-turn progress or spend reporting on this path, so a long re-drive is invisible until it finishes (the 600 s watchdog is the only log); and because the ceiling allows early stops, two re-drives of one case can differ in length (`drive_loop.py:5`).

libs/core/kiln_ai/synthetic_user/parser.py
```diff
@@ -0,0 +1,60 @@
...
+
+def _extract(blob: str, tag: str) -> str | None:
+    """First match, non-greedy content. Returns trimmed content, or None."""
+    m = re.search(rf"<{tag}>(.*?)</{tag}>", blob, re.DOTALL)
+    if m is None:
+        return None
+    return m.group(1).strip()
+
+
+def parse_synthetic_user_info(blob: str) -> SyntheticUserInfo:
```
chiang-daniel, line 32: **Decision made: the persona blob is strict on persona and goal, lenient on guidance.**

The blob is the tagged text the server generates (`<persona>`, `<goal>`, `<behavior_guidance>`). Missing or empty persona or goal fails the case before any model call (`bad_synthetic_user_info`); missing guidance is fine. First match wins and nesting is unsupported, so a persona that itself contains `</persona>` is silently truncated at that point; the server neutralises closing tags on its side, so this is a defence in depth, not the only guard. Unknown tags are ignored so a newer generator can add fields.

sfierro (approved):

leonardmq (approved):

PR #1785, opened 2026-09-16
[Eval V2 Kiln, review-only 4/7] Desktop and server routes: builder API, batch plan, case generation, save

chiang-daniel:
**Review-only.** Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with `app/desktop` and `libs/server` reverted to the baseline, so "Files changed" is the desktop's routes and models plus the local server's: 32 source files (+6138 / −290) and 15 test files. The generated kiln_server client is excluded. Details are inline on each hunk.

## How it works

The web UI talks only to the desktop's local FastAPI server. For the builder that server does three things: it runs the paid work locally (driving conversations with the synthetic user, running single-turn inputs, running the judge over each result) and streams progress to the page as server-sent events; it proxies four calls to kiln_server (clarifying questions, judge authoring, the claim card, the judge refiner) and one to generate synthetic users; and at the end it persists the whole eval in one unit of work with rollback. Existing eval routes gained percentiles, an empty-golden-set guard, a drive-readiness check, and a CRUD set for generated cases that nothing calls yet.

```mermaid
flowchart LR
  UI["builder page"] --> EB["eval_builder_api: 3 SSE streams (drive+judge, run+judge, re-judge) + 4 calls"]
  UI --> SDG["multiturn_sdg_api: generate_cases (chunked)"]
  UI --> CP["copilot_api: question_spec, refine_with_answers, spec_with_copilot (the save)"]
  EB --> KS["kiln_server: build_claim_evidence, generate_judge_prompt, refine_judge_prompt"]
  SDG --> KS2["kiln_server: synthetic_user/generate"]
  CP --> KS3["kiln_server: question_spec, refine_spec_with_answers_and_name"]
  EB --> Core["libs/core: synthetic_user runner, eval adapters (2/7, 3/7)"]
  CP --> Save["copilot_utils: split, tag, mint EvalInputs, rate golden, rollback"]
  Evals["eval results / compare pages"] --> EA["eval_api: percentiles, golden guard, readiness check, eval_inputs CRUD"]
```

**How to read this one:** 18k lines, 12k of them tests; skip those. Read the inline comments in this order, they walk you through: `eval_builder_api.py` (the streams), then `copilot_api.py` and `copilot_utils.py` (the save and the split policy), then `eval_api.py`, then `libs/server`. The remaining files are import order and annotation JSON.

- **`eval_builder_api.py`** (new, 1645 lines): the three streams on one base class (drain loop, judge semaphore, retries, per-case failure isolation, batch-fatal abort, cancel on disconnect) and the four request/response routes.
- **`multiturn_sdg_api.py`** (new): case generation in chunks of 20, four in flight; plus a streaming route nothing calls.
- **`copilot_api.py`**: the save, now one unit of work with three rollback ledgers; task tools and skills enriched into the copilot payloads; a 404-only fallback to the older refine route.
- **`copilot_utils.py`**: the split policy (train 40 : val 25, golden capped at a quarter), tagging, minting generated cases as items, golden ratings, and their inverses.
- **`eval_api.py`**: score percentiles, calibration guards, `multi_turn_item_count`, and five `eval_inputs` routes with no caller.
- **`libs/server`**: the name-availability route with a suffix policy, chain positions from trace boundaries (chain-of-thought safe), a 409 when a conversation is deleted mid-continuation, 15 agent-policy annotations.
- **Also in this view**: ten identical `suggested_for_synthetic_user` lines in `provider_api.py`, a numerically identical cap rename in `batch_plan_api.py`, three OpenAPI tags, `conftest.py`.

## Why this PR is where it is

Everything the builder spends money on goes through these routes, and the save is the one write that persists a whole eval. It depends on the datamodel (#1778) and the runner (2/7) and is called by the builder pages (7/7). Read the two pipeline streams, the save, and the split policy first.

## How to comment


app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+
+logger = logging.getLogger(__name__)
+
+# The builder's two concurrency knobs, one per pipeline stage:
+#   DRIVE_CONCURRENCY  — concurrent drive units: SU drive loops on the
+#                        multi-turn pipeline, one-shot task runs on the
+#                        single-turn one (the most expensive stage either way).
+#   REVIEW_CONCURRENCY — concurrent review units: judge calls on the merged
+#                        pipelines and the judge_traces re-judge.
+DRIVE_CONCURRENCY = 10
```
chiang-daniel, line 134: **Decision made: fan-out and retry constants for the local pipelines.**

`DRIVE_CONCURRENCY = 10` conversations at once (the SDK runner defaults to 4 for the standalone route), `REVIEW_CONCURRENCY = 8` judge calls, two retries with 1 s jittered backoff on transient errors for runs and for judges. The judge legs that reach kiln_server are not retried here because the server retries them. Each has a sentence of rationale; none a measurement.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+REVIEW_CONCURRENCY = 8
+
+# The single-turn run stage's knobs — the same posture as the multi-turn
+# drive runner (shared retry classifier, retry count, backoff base). Like the
+# multi-turn drive, a run has no app-level timeout: termination is
```
chiang-daniel, line 139: **Decision made: no app-level timeout on a drive or a run.**

Termination rests on the adapter's tool-call cap and the model client's per-request timeout; a 600 s watchdog logs and cancels nothing. A per-case budget existed and was removed because every firing killed a healthy run. A provider that streams slowly forever is bounded only by the user closing the page (which cancels every in-flight call, `eval_builder_api.py:340`). The same decision appears in 2/7 and 3/7; this is where the builder inherits it.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+
+class ReplaceBatchTagsField(BaseModel):
+    """The delete-on-redrive half of a drive request, shared by both driving
+    streams (multi-turn multi_turn_pipeline, single_turn_pipeline) so the batch
+    lifecycle contract can't drift between them. Subclasses declare their
+    own `batch_tag`; the self-replacement guard reads it by name."""
+
+    replace_batch_tags: list[str] = Field(
+        default_factory=list,
+        max_length=20,
```
chiang-daniel, line 201: **Decision made: a drive names the batches it replaces, at most 20, never its own.**

`replace_batch_tags` is how stranded batches from earlier drives are deleted (delete-on-redrive, `eval_builder_api.py:487`). Capped at 20 with a tag-charset validator; a drive that lists its own batch is rejected because it would destroy its results as they were produced. The page appends a tag per drive and never prunes on failure, so the 21st drive after a run of failures is a 422 (the page is in 7/7).

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+        ),
+    )
+    judge: JudgeConfig
+
+    @field_validator("inputs")
+    @classmethod
+    def inputs_must_be_non_blank(cls, value: list[str]) -> list[str]:
+        # A blank input would run the task on nothing; reject the request up
+        # front instead of streaming a guaranteed per-case failure.
+        for input_text in value:
```
chiang-daniel, line 340: **New: one base class for the three streams, so a closed page stops every in-flight paid call.**

**Why:** driving, running and re-judging all stream progress to the page; if the user closes the tab, the conversations and judge calls already in flight must stop spending, and a crash in the producer must reach the page as one frame, not a hung stream. `JudgeStreamBase` owns the frame queue, the judge semaphore, `_review_tasks`, the counters, and `events()`: drain until the producer and every judge task finish, turn a producer crash into one `batch_failed` frame, and on consumer disconnect cancel the producer and every in-flight judge so abandoned calls stop spending. Subclasses supply only what to produce and how the judge sees a case. Frames are dumped with `by_alias=True` so citations keep their `from` key.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+                    tag,
+                )
+            except Exception:
+                logger.exception(
+                    "%s: failed to delete superseded batch %s",
+                    self._stream_name,
+                    tag,
+                )
+
+    async def _judge_case(
```
chiang-daniel, line 519: **Decision made: how one case fails, and what `total_cost` counts.**

A judge failure fails that case only (`case_failed` with the unwrapped provider error and its class name); an auth, permission, not-found or budget error aborts the whole batch once. `total_cost` on every frame is drive or run spend only: the judge's spend is never added, and the re-judge stream reports 0.0 although every case invokes a judge. The docstrings say "drive spend"; the wire field says `total_cost`. Rename or add the judge lane.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+        error_type: str | None = None,
+    ) -> None:
+        """One case died at `stage`; the batch continues without it.
+
+        `error_type` is the class name of the provider or unexpected exception
+        behind the failure (None on deterministic failures, whose `code` already
+        names them) — included in the log line as a grep-able `error_type=...`
+        field so local logs can be grouped by failure kind, not just the frame.
+        """
+        logger.exception(
```
chiang-daniel, line 622: **Bug: `logger.exception` outside an exception handler.**

`_fail_case` always logs with a traceback, but it is called from places that are not inside an `except` (a missing trace after a drive, the three re-judge reload failures, the job-runner observer that receives the error as an argument). Python then logs `NoneType: None` and the real exception never reaches the log. Pass the exception in, or use `logger.error` with `exc_info` where one exists.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+                    "missing_output", "The run produced no output to judge."
+                )
+            # Belt-and-braces tagging (normally a no-op — default_tags above
+            # already landed the tags in the run's own save). Inside the
+            # try: a failure here surfaces as case_failed, never a silent
+            # drop.
+            async with self._save_context():
+                tag_single_turn_drive_run(run, self._batch_tag)
+            case_cost = _run_cost(run)
+            self._total_cost += case_cost
```
chiang-daniel, line 1175: **Bug: a narrow cost double-count and an orphaned `case_driven` frame.**

The run's cost is added here; if anything between here and the judge dispatch raises or is cancelled (the `case_driven` emit is an `await`), the cleanup path deletes the run and adds its cost again. The client has already received `case_driven` with an id that no longer exists. Add the cost in one place, after the run is known to survive.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+        The default is the multi-turn reading — transcript I/O plus the full
+        trace — matching what the saved eval judges. The single-turn streams
+        override only where raw_input comes from; both arms judge a
+        transcript, so every override still returns one.
+        """
+        assert trace is not None  # multi-turn streams always carry a trace
+        raw_input, raw_output = transcript_io_for_trace(trace)
+        return raw_input, raw_output, trace
+
+    async def _delete_superseded_batches(
```
chiang-daniel, line 487: **Decision made: superseded batches are deleted after the new drive produced replacements, best-effort.**

Runs from earlier drives are deleted only when the leaf carries exactly the runner's tags, has no rating and is nobody's parent; the exact-set match fails closed. A failed delete is logged and the batch stays on disk with no frame to the user (same for a failed attempt's run on the single-turn path, which stays tagged and is then included in the eval save).

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+            pass
+        if self._any_case_driven:
+            # Replacement runs exist on disk — now the superseded batches
+            # can go. A run stage that produced nothing keeps them untouched.
+            await self._delete_superseded_batches(
+                self._input.replace_batch_tags,
+                lambda tag: delete_single_turn_batch_runs(self._task, tag),
+                self._save_context,
+            )
+
```
chiang-daniel, line 1108: **The single-turn run: the judge sees byte for byte what the saved eval will judge later.**

**Why:** if the review judged one rendering and the saved eval another, the calibration the reviewer did on Step 5 would not carry over. One input per case: structured inputs are parsed as JSON (a parse failure is deterministic, never retried), a fresh adapter per attempt with default tags so a cancelled run is still discoverable, `log_if_slow` instead of a timeout, and the judge reads the request's input string verbatim plus the transcript's output, which is byte-identical to what the saved eval will judge later.

app/desktop/studio_server/eval_builder_api.py
```diff
@@ -0,0 +1,1645 @@
...
+                "model_name": self._input.input_model_name,
+                "model_provider": self._input.input_provider.value,
+                "adapter_name": _SINGLE_TURN_ADAPTER_NAME,
+                "batch_tag": self._batch_tag,
+            },
+        )
+
+
+def connect_eval_builder_api(app: FastAPI):
+    @app.post(
```
chiang-daniel, line 1274: **New routes: the three streams and four calls, one caller each.**

`multi_turn_pipeline`, `single_turn_pipeline`, `judge_traces` (SSE, `@no_write_lock` because a lock would buffer the stream and break cancel-on-disconnect), `build_claims`, `preflight_model`, `author_judge`, `refine_judge`. The copilot key is checked first on the streams so a missing key fails before the user's own model spend; `refine_judge` and `build_claims` check it later inside the call. The module docstring still says only the claim builder reaches kiln_server; three routes do.

app/desktop/studio_server/api_models/eval_builder_models.py
```diff
@@ -0,0 +1,357 @@
...
+    """
+
+    raw_input: str
+    raw_output: str
+    eval_rubric: str
+    judge_reasoning: str
+    judge_score: JudgeScoreLiteral
+
+
+class BuildClaimsApiOutput(BaseModel):
```
chiang-daniel, line 115: **Wire contract for the review card, and one gap.**

`BuildClaimsApiOutput` says "one to eight claims" in its docstring but declares no bounds; the server enforces them, the desktop does not. `RefineJudgeApiInput.graded_traces` is capped at 50 with no rationale (the shipped path grades ten of forty). No string field on this surface has a size cap, in either direction, and every `case_judged` frame carries the complete trace. `PipelineCaseFailedEvent`'s comment still lists `case_timeout`, which can no longer occur.

app/desktop/studio_server/utils/eval_builder_utils.py
```diff
@@ -0,0 +1,456 @@
...
+
+@dataclass
+class JudgeVerdict:
+    """A judge's decision for one trace, in the shape the claim builder wants."""
+
+    judge_score: JudgeScoreLiteral
+    judge_reasoning: str
+
+
+def transcript_io_for_trace(trace: list[Any]) -> tuple[str, str]:
```
chiang-daniel, line 77: **Decision made: one canonical text for the judge, the claim builder and the citation highlighter.**

**Why:** a claim's citation is a character span; it only lands on the right words if the judge, the card builder and the highlighter all read the same rendering of the conversation. `raw_output` is the role-labelled transcript from the shared formatter, the same rendering the judge template's `format_trace` produces; `raw_input` is the first non-empty user message, or an empty string when there is none, silently. The judge prompt template raw-wraps the spec text so Jinja in a spec cannot inject, and states that the data blocks are data.

app/desktop/studio_server/utils/eval_builder_utils.py
```diff
@@ -0,0 +1,456 @@
...
+        properties=LlmJudgeProperties(
+            model_name=judge.model_name,
+            model_provider=judge.model_provider,
+            prompt_template=build_judge_prompt_template(judge.prompt, multi_turn),
+        ),
+        parent=eval_obj,
+    )
+
+
+def _reasoning_from_intermediates(
```
chiang-daniel, line 199: **Silent fallbacks to confirm: synthesized reasoning and the verdict-claim rule.**

When the judge model produced no reasoning, a placeholder sentence ("The judge returned a PASS verdict without an explicit reasoning trace.") is fed to the claim builder as `judge_reasoning`, indistinguishable from real reasoning downstream. The verdict claim is recognised by a prose prefix ("It passes" / "It fails") on the last claim only; nothing logs when no claim matches, so a wording drift on the server degrades silently.

app/desktop/studio_server/utils/eval_builder_utils.py
```diff
@@ -0,0 +1,456 @@
...
+            "task_instruction": task_instruction,
+            "raw_input": raw_input,
+            "raw_output": raw_output,
+            "eval_rubric": eval_rubric,
+            "judge_reasoning": judge_reasoning,
+            "judge_score": judge_score,
+        }
+    )
+
+    detailed_result = await build_claim_evidence_v1_copilot_build_claim_evidence_post.asyncio_detailed(
```
chiang-daniel, line 342: **Inconsistency: the three kiln_server calls here do not classify transport errors.**

The sibling proxies in `copilot_api.py` and `batch_plan_api.py` catch connection failures and raise a 502 with an actionable message; these three let an unreachable server surface as an unshaped 500, and their "unexpected response type" paths raise "Unknown error" without logging what came back.

app/desktop/studio_server/multiturn_sdg_api.py
```diff
@@ -0,0 +1,613 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Persona cases asked for per kiln_server call. kiln_server accepts at most
+# 50 cases per request and writes a request's whole batch in a single model
+# call, so the generated output grows with the count: 20 keeps each call
+# short (faster, and less of it lost to truncation and salvage) and well
+# under the server's bound. If that bound ever changes this must stay at or
+# below it.
+SU_CASES_PER_CALL = 20
```
chiang-daniel, line 87: **Decision made: synthetic users are generated in chunks of 20, four in flight.**

kiln_server accepts at most 50 cases per request and generates a request's batch in one model call; 20 keeps each call short; the old single request of 200 was a 422 against that cap. The server's 50 lives only in this comment. Each chunk's `scenario_index` is re-based from chunk-relative to plan-relative; a case with no index passes through untouched.

app/desktop/studio_server/multiturn_sdg_api.py
```diff
@@ -0,0 +1,613 @@
...
+                    case_scenarios=scenarios,
+                )
+            return [_case_dict_at_plan_offset(c, start) for c in chunk_cases]
+
+        try:
+            # gather keeps results in argument order, so the chunks stitch back
+            # in plan order. A chunk's typed failure IS the request's failure:
+            # it propagates out of gather and maps exactly as a single call's
+            # did, rather than shipping a partial batch.
+            by_chunk = await asyncio.gather(
```
chiang-daniel, line 482: **Decision made: one failed chunk fails the request, and the other chunks keep spending.**

`asyncio.gather` without `return_exceptions`: the first failure propagates and the request errors, but the sibling chunk calls keep running to completion (and billing) with their results discarded. A short chunk (salvage) is a plain 200 with fewer cases and nothing on the wire saying so (a test pins 44 of 45). A transport timeout escapes as a bare 500.

app/desktop/studio_server/multiturn_sdg_api.py
```diff
@@ -0,0 +1,613 @@
...
+            raise HTTPException(
+                status_code=502,
+                detail={
+                    "code": "upstream_invalid_output",
+                    "message": "Synthetic-user generator returned no cases.",
+                },
+            )
+        return GenerateCasesApiOutput(cases=cases)
+
+    @app.post(
```
chiang-daniel, line 510: **Dead route: `run_cases_batch` has no caller.**

Registered, annotated for agents, covered by about 30 tests; the builder drives through `eval_builder/multi_turn_pipeline`, which imports the library function directly. Remove, or say why it stays.

app/desktop/studio_server/copilot_api.py
```diff
@@ -110,23 +154,163 @@
 logger = logging.getLogger(__name__)
 
 
+async def copilot_passthrough_payload(
```
chiang-daniel, line 157: **Decision made: the studio enriches copilot payloads with the task's tools and skills, and never sends null for "unknown".**

When the client names a task, the studio reads one run config's tools and skills (names and descriptions only) and fills the payload; a client-supplied value wins; the studio-local ids are stripped before the body leaves. Absent means "not collected", empty means "the task has none". One route bypasses this: `generate_batch` (line 837) dumps the model directly and sends `task_tools: null`; no live impact because nothing calls it.

app/desktop/studio_server/copilot_api.py
```diff
@@ -110,23 +154,163 @@
...
 
     The client is responsible for building:
-    - definition: The spec definition string (use buildSpecDefinition on client)
-    - properties: The spec properties object (filtered, with spec_type included)
+    - definition: the spec definition string (buildSpecDefinition on client)
+    - properties: the spec properties object (filtered, with spec_type included)
     """
 
-    name: FilenameString
+    # Short limit: the name becomes the eval's EvalOutputScore.name (max 32)
```
chiang-daniel, line 307: **Bug fixes on the save request.**

The eval name is `FilenameStringShort` with a validator that it yields a JSON key (a name with no derivable key used to persist an eval every job then failed on); exactly one of the three synthesis paths must be present; the judge config is required (the calibrated judge ships). The comment above `task_sample` says the wizard arms omit it; the single-turn arm sends it.

app/desktop/studio_server/copilot_api.py
```diff
@@ -535,17 +904,57 @@ async def submit_question_answers(
 
         submit_input = SubmitAnswersRequestServerApi.from_dict(request.model_dump())
 
-        detailed_result = await refine_spec_with_answers_v1_copilot_refine_spec_with_answers_post.asyncio_detailed(
+        # Prefer the newer route that also returns a model-suggested eval name.
```
chiang-daniel, line 907: **Decision made: a 404-only fallback to the older refine route.**

`submit_question_answers` calls the newer `refine_spec_with_answers_and_name`; on a 404 only (the route not deployed) it falls back to the older route with a warning; any other status propagates. Well-scoped and labelled temporary, and it encodes a deploy-order dependency: kiln_server before the app.

app/desktop/studio_server/copilot_api.py
```diff
@@ -731,81 +1154,252 @@ async def create_spec_with_copilot(
         """
         task = task_from_id(project_id, task_id)
 
+        # Idempotency guard against re-submits after a completed save (the
```
chiang-daniel, line 1157: **Bug fix and decision: the save is one unit of work.**

A save whose name derives to the same eval tags as an existing spec is a 409 (before, "My Spec" and "my_spec" could share a tag namespace and each other's datasets; the race between two concurrent saves is documented as accepted for a single-user studio). Every reviewed reference must name a run of this batch, once, checked before anything is created. `persist_spec_save` runs off the event loop with three rollback ledgers (created models, tagged runs, rated runs) and reverses run mutations before deleting models. A `reference_answer` eval is refused with a 400 because the saved judge would never see the reference answer.

app/desktop/studio_server/utils/copilot_utils.py
```diff
@@ -30,24 +50,98 @@
...
+# advertises the resulting dataset size to the user off these.
 NUM_SAMPLES_PER_TOPIC = 20
 NUM_TOPICS = 15
-# Matches the golden-set goal the eval detail page holds users to
-# (MIN_GOLDEN_DATASET_SIZE). Every example above the reviewed ones is minted unrated, so
-# this is the hand-rating backlog a new spec starts with: a floor above the goal asks for
-# work the app never asks for again.
-MIN_GOLDEN_EXAMPLES = 12
+
+# Dataset split — the 50/25/25 spec (train / eval / golden). Golden is the
```
chiang-daniel, line 92: **Decision made: the split policy, and two encodings of one quarter.**

Golden is at most a quarter of the batch, drawn from rated runs only, never padded; the rest is dealt train 40 : val 25 by largest remainder. `GOLDEN_TARGET_FRACTION = 0.25` drives only the warning while the cap is computed from the weights, so they can drift; the cap floors, so any batch whose size is not a multiple of four trips the "below target" warning on a correctly sized split; a batch of one to three runs yields an empty golden set however much was reviewed, and the eval still saves. The train and val weights are hand-mirrored in `eval_generation_splits.ts`. The comment names a known leak: val runs share inputs with the test slice.

app/desktop/studio_server/utils/copilot_utils.py
```diff
@@ -61,6 +155,192 @@ def get_copilot_api_key() -> str:
     return api_key
 
 
+async def task_capabilities_for_task(
```
chiang-daniel, line 158: **Decision made: capability collection degrades to "not collected" on any failure.**

One MCP session per call; every tool in the run config is resolved (an MCP tool dials its server), per-tool failures are skipped, any other failure yields `(None, None)` with a warning. The user is never told the judge was authored without the task's tools. No count cap and no timeout on the collection; only an elapsed-ms log.

app/desktop/studio_server/utils/copilot_utils.py
```diff
@@ -246,70 +563,568 @@ def save_pending_feedback(self, task_run: TaskRun) -> None:
...
+    unit-of-work)."""
+    eval_inputs = build_single_turn_eval_inputs(
+        inputs, eval_tag, [f"{_TAG_PREFIX_SINGLE_TURN_DRIVE_BATCH}{batch_tag}"]
+    )
+    for eval_input in eval_inputs:
+        eval_input.parent = task
+    return eval_inputs
+
+
+def find_multi_turn_chain_leaves(task: Task, batch_tag: str) -> list[TaskRun]:
```
chiang-daniel, line 690: **Decision made: what delete-on-redrive may delete.**

A chain is deleted only when its leaf carries exactly the runner's tags, has no rating, and is nobody's parent, plus a whole-chain fork check; same rule for single-turn runs. Loads the whole run corpus once per pass, which is the expensive part of a re-drive on a large task.

app/desktop/studio_server/utils/copilot_utils.py
```diff
@@ -246,70 +563,568 @@ def save_pending_feedback(self, task_run: TaskRun) -> None:
...
+        current = set(leaf.tags or [])
+        if tag in current:
+            continue
+        leaf.tags = sorted(current | {tag})
+        leaf.save_to_file()
+        if tagged_out is not None:
+            tagged_out.append((leaf, {tag}))
+
+
+def build_multi_turn_eval_inputs(
```
chiang-daniel, line 959: **Decision made: generated cases are saved as items, not runs.**

Multi-turn: one `EvalInput` per driven case with the seed, the parsed persona and the drive config; an unparseable persona blob is a 422 before anything is written. Single-turn: inputs only, no output, because the runner produces a fresh output per run config at eval time. Each item is appended to a rollback ledger the moment it hits disk.

app/desktop/studio_server/eval_api.py
```diff
@@ -1629,6 +2033,187 @@ async def get_eval_default_judge_types(
                 break
         return result
 
+    @app.get(
```
chiang-daniel, line 2036: **New routes with no caller: create, read, update, delete for generated cases. They exist so an agent or API client can manage a builder-made eval's cases the way it manages dataset runs; the web UI does not use them.**

Five `eval_inputs` routes (list with a partial-load error count, get, create, patch reading `model_fields_set` so omit and null differ, delete with a 409 when a trace or score still names the item). Nothing in the web UI calls them; they are an agent and API surface from an upstream PR that is not on main. Agent policy: create allowed silently, patch needs approval, delete denied; the approval text for patch argues that tags decide the split, which applies to create too. The delete guard's docstring says it avoids a full load; it scans every run and every eval record.

app/desktop/studio_server/eval_api.py
```diff
@@ -2148,7 +2747,19 @@ async def run_eval_config_eval(
     ) -> StreamingResponse:
         """Run all eval configs against each other for calibration and stream progress via SSE. Used to check that eval configs produce consistent scores."""
         eval = eval_from_id(project_id, task_id, eval_id)
-        require_golden_set_or_422(eval)
+        golden_filter_id = require_golden_set_or_422(eval)
```
chiang-daniel, line 2750: **Bug fix with a gap: an empty golden set is refused before calibration.**

An empty set used to "complete" instantly with zero scores, which the UI read as success. The check is on the filter being empty; the message says "Add human-rated examples", but a golden set of unrated runs passes and produces the same vacuous result, and a small batch can mint exactly that (`copilot_utils.py:92`).

app/desktop/studio_server/eval_api.py
```diff
@@ -839,7 +1122,65 @@ def scored_trace_usage_for_run_config(
             ):
                 scored_run_ids.add(eval_run.scored_run_id)
     traces = load_task_children_by_id(TaskRun, task, scored_run_ids)
-    return {run_id: trace.usage for run_id, trace in traces.items()}
+    return {run_id: scored_trace_usage(trace) for run_id, trace in traces.items()}
```
chiang-daniel, line 1125: **Decision made: what a scored conversation's spend means.**

For a chain leaf, tokens and latency come from `cumulative_usage`, and only the cost is blended in from the synthetic user's usage; `None` when nothing was measured. `ScoreSummary` gains min, quartiles, p90 and max, all optional for older consumers; `multi_turn_item_count` on the result summary is required with no default. Calibration summaries now use calibration records only (a golden item's fresh-generation score was being correlated against its own human rating).

app/desktop/studio_server/eval_api.py
```diff
@@ -2075,6 +2660,20 @@ async def run_eval_config(
             save_context=build_save_context(request),
         )
 
+        # Surface drive-config/run-config incompatibilities as one 400 before
```
chiang-daniel, line 2663: **Decision made: readiness is checked once, before the stream opens.**

`run_eval_config` calls the runner's `validate_multi_turn_drive_readiness` and maps a `ValueError` to one 400 naming every problem, instead of N anonymous per-job errors mid-stream. Its own preflight can pass an unresolvable config (2/7, `eval_runner.py:192`).

libs/server/kiln_server/run_api.py
```diff
@@ -161,6 +161,37 @@ def _count_user_messages(trace: list[Any] | None) -> int:
     return sum(1 for m in trace if isinstance(m, dict) and m.get("role") == "user")
 
 
+def _position_chain_turns(
```
chiang-daniel, line 164: **Bug fix: chain positions come from trace boundaries, not user-message counts.**

Chain-of-thought strategies emit two user messages per turn, so counting them mispositioned every turn on those tasks (the fork handles landed on the wrong bubbles, 6/7). Each run is now anchored where its turn starts by trace-length monotonicity; a run whose trace does not extend its predecessor's marks the chain broken. `turn_index` changed meaning from absolute to chain-relative with no version bump; the one in-repo consumer was updated, an external client reading it as absolute would be wrong.

libs/server/kiln_server/run_api.py
```diff
@@ -762,12 +807,31 @@ async def run_task(
...
             prior_trace = parent_task_run.trace
 
-        return await adapter.invoke(
+        run = await adapter.invoke(
             input,
             prior_trace=prior_trace,
             parent_task_run=parent_task_run,
         )
 
+        # The conversation may have been cascade-deleted while the model was
```
chiang-daniel, line 816: **Bug fix: a conversation deleted mid-continuation is not resurrected.**

After the model call, the parent is re-checked; if it is gone, the just-saved run is deleted and a 409 says so. If that delete itself throws, the orphan survives and the 409 is replaced.

libs/server/kiln_server/spec_api.py
```diff
@@ -74,7 +76,89 @@ class SpecCreationRequest(BaseModel):
...
+        description="The candidate itself when free, else the nearest "
+        "available suffixed variant."
+    )
+    was_taken: bool = Field(
+        description="Whether the candidate collided with an existing spec "
+        "(and `name` is therefore a suffixed variant)."
+    )
+
+
+def resolve_available_spec_name(
```
chiang-daniel, line 92: **Decision made: an available spec name is found by suffixing, comparing derived tags.**

Candidates compare by the eval tag they derive to, not the raw string; a collision gets " 2", " 3", up to 99, then a 409. The 32-character cap is repeated as a literal beside the type that owns it. The request name is now `FilenameStringShort` so a long name is a 422 instead of a 500 inside eval construction.

libs/server/kiln_server/utils/agent_checks/annotations/post_api_projects_project_id_tasks_task_id_eval_builder_author_judge.json
```diff
@@ -0,0 +1,9 @@
+{
```
chiang-daniel, line 1: **Note: the annotation gate never ran on this branch.**

The CI check that catches a missing agent-policy annotation runs on pull requests only, so on a branch built from direct commits `author_judge` went un-annotated until it was caught by hand. Worth knowing for the next branch of this shape. `run_cases_batch` has an annotation for a route with no caller.

app/desktop/studio_server/synthetic_user/client.py
```diff
@@ -0,0 +1,202 @@
+"""Thin async wrapper over the vendored kiln_server SDK for `/generate`.
```
chiang-daniel, line 1: **New: the typed client for synthetic-user generation, with no retry.**

Error classification into request and server errors by status (2xx, 502, 500, 401 via the shared component, 422 rendered to one line) and a fallback by range; no retry because kiln_server retries transient provider failures internally. On an unmodelled status the response body is discarded, so an unparsed 2xx reads as `http_200` in a log.

app/desktop/studio_server/api_models/copilot_models.py
```diff
@@ -48,6 +80,40 @@ class SampleApi(BaseModel):
     model_config = {"populate_by_name": True}
 
 
+class ClaimReviewApi(BaseModel):
```
chiang-daniel, line 83: **New wire models for the save: what the builder sends when it saves an eval.**

**Why:** the save carries three things the old form never had: the task's tools and skills (so the copilot prompts see the real task), each reviewed conversation's claim grades (so they persist beside the golden rating and the refiner can read them later), and the driven synthetic cases (so the eval runner can re-drive them per run config).

- `TaskToolInfoApi` / `TaskSkillInfoApi` and the optional `task_tools` / `task_skills` on `TaskInfoApi`: `None` means not collected (prompts unchanged), `[]` means the task has none.
- `ClaimReviewApi` mirrors the persisted `ClaimReview` (1/7). `verdicts_must_agree` rejects a payload whose `user_says_meets_spec` disagrees with the review's `human_verdict`: both come from one derivation in the page, so a mismatch is a corrupt payload, refused before an answer key is written that contradicts the review saved beside it.
- `ReviewedChainApi`: a multi-turn verdict keyed by the leaf run id, the identity that rides from drive to save.
- `DrivenSyntheticCaseApi`: the seed, the tagged persona blob (parsed before anything persists) and the plan scenario index, recorded as a `scenario:{index}` tag on the minted item.

app/desktop/studio_server/utils/eval_builder_utils.py
chiang-daniel: @tawnymanticore two hunks here decide what your prompts see: this one (the single rendering the judge, the card builder and the citation highlighter share) and the one just below it (a placeholder reasoning fed to the claim builder when the judge gave none, and the verdict claim matched by a prose prefix).

sfierro (approved):

leonardmq (approved):

PR #1786, opened 2026-09-16
[Eval V2 Kiln, review-only 5/7] Shared UI: lib/ui and components the builder touched

chiang-daniel:
**Review-only.** The goal of this review: nothing on a shipped page got worse. Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with `app/web_ui/src/lib` and `app.css` reverted to the baseline, so "Files changed" is every shared control and helper the branch touched: 65 source files (+2153 / −853) and 42 test files. The generated `api_schema.d.ts` is excluded. Details are inline on each hunk.

## How it works

`lib/ui` and `lib/components` are the house controls every page is built from, and `lib/utils` the helpers behind them. The builder reused them wherever it could, and where it needed more it extended the control rather than forking it: a wider dialog, a header with an action slot, an output panel that can highlight a cited span, a chat trace that can be scrolled to a citation. Those extensions ride into every existing page that renders the control, which is what this PR is for. Three changes alter what an existing page shows without the page changing: the warning control lost its outer margin, chat bubbles render JSON as a code panel, and a global stylesheet rule hides closed nested dialogs.

```mermaid
flowchart LR
  Pages["every (app) page"] --> Warn["warning.svelte: margin removed, prop renamed"]
  Pages --> Dlg["dialog.svelte: extra_wide"]
  Pages --> Out["output.svelte: JSON stale-render fix, mark, slot"]
  Pages --> Chat["chat_trace.svelte: JSON panels, structured answer bubble, citation highlight"]
  Pages --> Sel["fancy_select.svelte: now fires change"]
  Pages --> CSS["app.css: closed nested dialogs leave layout"]
  Builder["builder (7/7)"] -.uses.-> Warn & Dlg & Out & Chat & Hdr["settings_header: actions and subtitle slots"]
  Evals["eval judge forms (lib/components/eval_types)"] --> Res["*_result.svelte: Pass/Fail badge fix"]
  Utils["lib/utils: sse, deadline_signal, data_guide_return, code_tool_helpers, link_builder"]
```

- **Reaches existing pages by itself**: warning margin (63 import sites, 20 without a compensating edit), the JSON code panel and the "Output" bubble on the two multi-turn run pages (Run, dataset run), the global dialog rule, `fancy_select` dispatching `change`, the centred success title, the model dropdown's empty state.
- **Bug fixes**: `output.svelte` stale JSON render (~29 pages); deterministic judge results never showed a Pass/Fail badge (7 files); the git-import empty-path guard; code-tool stub generation (reserved words, docstring escaping, Back wiping code).
- **New controls and helpers**: `json_span_map.ts`, `sse.ts`, `deadline_signal.ts`, `data_guide_return.ts`, `default_judge.ts`, `suggested_models.ts`, `conversation_animation.svelte`, `task_response_tool.ts`.
- **Cleanup**: `trace.svelte` drops five props nothing passed; a dead icon; a dead type alias; the `output_value_expression` prop removed from four judge forms (one leftover binding remains).
- **Also in this view, mechanical, each with its why inline**: 12 `tight` → `inline` call-site renames in 7 files (the prop was split in two, see the warning comment); three files where most changed lines are re-indentation after a wrapper was removed or restructured (`trace.svelte` 189 real lines of 409, `streaming_chat.ts` 31 of 185, `available_models_dropdown.svelte` 56 of 160); a byte-equivalent extraction (`llm_judge_form` → `suggested_models.ts`).

## Why this PR is where it is

Every control here is on shipped pages today, so a change here is a change for every user, not only builder users. Read the first bullet's list line by line; the rest can be skimmed with the inline notes.

## How to comment


app/web_ui/src/lib/ui/warning.svelte
```diff
@@ -10,7 +10,13 @@
     | undefined = undefined
   export let warning_icon: "exclaim" | "info" | "check" = "exclaim"
   export let large_icon: boolean = false
-  export let tight: boolean = false
+  // Warning owns no outer margin: the container it sits in owns the spacing
```
chiang-daniel, line 13: **Behaviour change on 63 pages: the warning control no longer carries an outer margin, and `tight` is now `inline`.**

**Where you see it:** any page with an inline warning: the delete and edit dialogs, the git-import steps, the fine-tune list, the generate pages, the tools pages, the results page.

**What this is:** the house inline advisory (icon plus message). Before, a plain warning added an 8 px top margin of its own; `tight` narrowed the icon gap.

**What changed:** the margin is gone (the container owns spacing, the same way every other form row is spaced; `outline` keeps its bottom margin). **Why the rename:** `tight` used to mean two things at once, drop the margin and narrow the icon gap, so callers inside a spacing column passed it just to avoid a double gap and got the narrow indent whether they wanted it or not. With the margin gone the flag means one thing, the indent, and is renamed `inline` to say so; the 57 `tight` → `inline={true}` call-site edits (12 in this view, 45 in 6/7, across 25 files) are that rename, nothing else. Thirty-five files were given container spacing in two follow-up commits.

**What to check:** 20 plain call sites in files no spacing commit touched lose 8 px: `delete_dialog`, `edit_dialog`, five git-import steps, the fine-tune list, four generate pages, three tools pages, the results page, `prompt_type_selector`, `oauth_install_step`. Several sit in flex columns with their own gap where nothing moves; the list is what to eyeball, not a list of regressions.

app/web_ui/src/lib/ui/warning.svelte
```diff
@@ -57,21 +63,27 @@
         return ""
     }
   }
+
```
chiang-daniel, line 66: **Bug fix: a warning whose colour arrives after mount kept its first colour.**

**Where you see it:** any warning whose colour depends on fetched data, such as the run page's model advisory.

`get_color()` was called from the template, and Svelte tracks the variables an expression names, not the function's inputs, so a colour that resolved after a fetch never re-rendered. The colour classes are now reactive values.

app/web_ui/src/app.css
```diff
@@ -22,6 +22,20 @@ a {
   @apply cursor-pointer;
 }
 
+/* A closed daisyUI modal stays in layout (display kept for its fades).
```
chiang-daniel, line 25: **Decision made: a global rule takes closed nested dialogs out of layout.**

**Where you see it:** any dialog that contains its own dialogs: the results page trace dialog, the jobs dialog, synth generation, Q&A, the RAG chunker, usage info.

**The bug it fixes:** a closed daisyUI modal stays in layout for its fade; nested inside another modal's box, the box's transform makes the closed overlay position against the box, inflating the scroll area by 1 rem: a permanent phantom scrollbar on every dialog whose content carries its own dialogs (the chat trace's tool and usage viewers).

**The cost:** every nested dialog app-wide loses its fade in and out (results page, jobs dialog, synth generation, Q&A, RAG chunker, usage info). Assumes `[open]` is the only open-state signal; a nested modal opened via `modal-open` or `:target` would be hidden outright. Nothing enforces that.

app/web_ui/src/lib/ui/trace/chat_trace.svelte
```diff
@@ -190,7 +409,28 @@
         <div class="group flex flex-col items-end" data-testid="chat-msg-user">
           <div class="rounded-xl bg-primary/10 px-4 py-3 max-w-[70%] text-sm">
             {#if content}
-              <ChatMarkdown text={content} />
+              {#if highlight && highlight.kind === "content" && highlight.trace_index === index}
```
chiang-daniel, line 412: **Behaviour change on the three run pages: a chat turn that is JSON renders as a code panel.**

**Where you see it:** the Run page and the dataset run page, on a multi-turn task only, on any chat turn whose text is JSON. Single-turn runs use the older trace view and are unaffected. Also the results page's new Trace dialog (6/7) and the builder.

**What this is:** the chat-bubble trace viewer. On shipped pages it renders only multi-turn conversations (the Run page and the dataset run page gate it on the task's turn mode); the results page's Trace dialog and the builder are new consumers on this branch.

**What changed:** user content, reasoning and assistant content each gain a branch: if the text parses as JSON and is not a bare string, it renders through `Output` (pretty-printed, highlighted) instead of markdown. A quoted string stays markdown; a bare number or array does not (tests pin both). A user who pastes `[1, 2, 3]` sees a code block.

**Decision made:** wanted on the two shipped multi-turn run pages, or only in the builder and the results dialog? No existing project data has a JSON chat turn on those pages; the change is pinned by tests, not seen live.

app/web_ui/src/lib/ui/trace/chat_trace.svelte
```diff
@@ -284,99 +559,175 @@
                 {@const tool_error = result
                   ? is_tool_error(result.message)
                   : false}
-                <div
-                  class="flex flex-col items-start"
-                  data-testid="chat-msg-assistant"
-                >
-                  <!-- svelte-ignore a11y-click-events-have-key-events a11y-no-static-element-interactions -->
+                <!-- A tool-CALL citation lands on the whole bubble (the call
```
chiang-daniel, line 562: **Behaviour change on the same pages: the structured answer reads as the answer, not as a tool call.**

**Where you see it:** the same two multi-turn run pages, for a multi-turn task with structured output. Single-turn runs are unaffected.

A structured-output task returns its answer through a synthetic `task_response` tool call. It rendered as a "Toolcall: task_response" expander; it now renders as a bubble headed "Output" with the arguments as JSON. Every structured-output multi-turn conversation on those two pages changes. Almost certainly right. No existing project data exercises it (no multi-turn task has structured output), so it is pinned by tests, not seen live.

app/web_ui/src/lib/ui/fancy_select.svelte
```diff
@@ -97,6 +97,11 @@
         }, 0)
       }, 0)
     }
+
```
chiang-daniel, line 100: **Behaviour change on every page with a fancy dropdown: it now fires `change`.**

**Where you see it:** any form with a fancy dropdown: the spec judge form, the judge builder, the code-eval form, the compare chart, the models page.

**What this is:** the house custom dropdown, behind every `FormElement` of type `fancy_select` or `multi_select`, the compare chart and the models page. It produced no native `change` event, so ancestor `on:change` handlers (unsaved-changes tracking) never saw a pick.

**What changed:** a bubbling `change` is dispatched after a selection, like a native `<select>`.

**What a user notices:** picking a judge model on the spec-judge form now marks it dirty, so the unsaved-changes confirm appears where it did not; the judge builder and the code-eval form react the same way. No test in this area pins it.

app/web_ui/src/lib/ui/completed.svelte
```diff
@@ -11,7 +11,6 @@
 <div class="flex flex-col items-center justify-center min-h-[60vh]">
   <Intro
     {title}
-    align_title_left={true}
```
chiang-daniel, line 14: **Behaviour change on five success screens: the title is centred.**

**Where you see it:** the five success screens: Fine Tune Created, Data Added, Prompt Optimization Job Started, both Data Guide Saved.

`align_title_left={true}` is removed from the shared success screen, so Fine Tune Created, Data Added, Prompt Optimization Job Started and both Data Guide Saved screens now centre their title like the builder's done screen. The control owner made this change.

app/web_ui/src/lib/ui/run_config_component/available_models_dropdown.svelte
```diff
@@ -38,6 +41,14 @@
   export let optional: boolean = false
   export let hide_optional_badge: boolean = false
   export let empty_label: string = "Select a model"
+  // Empty-dropdown affordance (fancy_select's built-in empty state). An
```
chiang-daniel, line 44: **Behaviour change on the run page, extractor create and the judge form: an empty state and a quieter advisory.**

**Where you see it:** the run page, extractor create and the judge form, when no provider is connected.

When no provider is connected the dropdown shows "Connect an AI provider" with a link to settings instead of an empty list; the model advisory is suppressed while the list is still loading so it cannot judge a model against a list that has not arrived; the wrapper gains a `gap-2`. Of the 160 changed lines, 56 are real; the rest are re-indentation under the restructured wrapper.

app/web_ui/src/lib/ui/dialog.svelte
```diff
@@ -11,7 +11,12 @@
   export let sub_subtitle: string | null = null
   export let sub_subtitle_link: string | null = null
   export let blur_background: boolean = false
-  export let width: "normal" | "wide" = "normal"
+  // Dialog width. "extra_wide" is for content that reads as a full page in
```
chiang-daniel, line 14: **New width: `extra_wide`. Needed because the builder's trace dialog shows a whole conversation, and the house dialog's widest setting truncated the bubbles.**

**Where you see it:** no shipped dialog changes width. The two consumers are new on this branch: the builder's trace dialog and the results page's Trace dialog (6/7).

`max-w-7xl` for a dialog that holds a whole conversation. Two consumers, both new on this branch: the builder's trace dialog and the eval results page's Trace dialog. No existing dialog moves from `wide` to `extra_wide`. The class-retention div carries the new class so Tailwind emits it.

app/web_ui/src/lib/ui/output.svelte
```diff
@@ -1,37 +1,90 @@
...
-      if (typeof json_output !== "string") {
-        formatted_json_html = JSON.stringify(json_output, null, 2)
-        formatted_json_html = hljs.highlight(formatted_json_html, {
-          language: "json",
-        }).value
-      }
-    } catch (_) {
-      formatted_json_html = null
-    }
+  // Guarded by the predicate, so the parse here cannot throw.
```
chiang-daniel, line 42: **Bug fix on ~29 pages, unadvertised: a JSON string no longer leaves the previous JSON on screen.**

**Where you see it:** any Output panel (about 29 pages) whose content is a JSON string.

**What this is:** the house read-only panel: pretty-prints JSON, otherwise raw text.

**The bug:** the old reactive block assigned the highlighted HTML only when the parse succeeded and was not a string, and cleared it only on a parse error. Content that parsed to a JSON string scalar (`"hello"`) hit neither branch, so the panel kept whatever the previous content had produced.

**The fix:** `printed_json` is derived in one expression that always resolves. Also new here: `is_non_string_json` exported for the chat viewer, a `mark` span rendered as `<mark>` in both the text and the JSON path (`json_span_map.ts` translates the span onto the pretty-printed layout), and a `transparent` background.

app/web_ui/src/lib/components/eval_types/contains_form.svelte
```diff
@@ -19,11 +19,6 @@
   export let reference_candidate_keys: string[] = []
   export let required_reference_fields: string[] = []
 
-  // Read-only mirror of the configured output source, surfaced to the parent so
```
chiang-daniel, line 22: **Cleanup with one leftover: `output_value_expression` removed from four judge forms, but the parent still binds it.**

**Where you see it:** the browser dev console on the judge builder and the spec-judge form.

The prop and its reactive mirror are gone from the contains, exact-match, pattern-match and set-check forms (and from the judge builder that bound it). `judge_config_fields.svelte`, unchanged on the branch, still declares the prop and `bind:`s it onto those four components on the judge builder and the spec-judge form. Svelte warns in dev about the unknown prop; the value is always null. Remove lines 35, 75 and 81 there too.

app/web_ui/src/lib/components/eval_types/contains_result.svelte
```diff
@@ -12,8 +12,12 @@
 
   $: props = extractV2Props(eval_config, "contains")
 
-  $: passed = scores.match === 1.0
-  $: has_score = "match" in scores
+  // Deterministic types emit one binary value per declared output score, keyed
```
chiang-daniel, line 15: **Bug fix on the results page: deterministic judge results never showed their Pass/Fail badge.**

**Where you see it:** the eval results page for a deterministic judge (contains, exact match, pattern match, set check and the rest).

The seven `*_result` components checked `"match" in scores`, but scores are keyed by the eval's own score names, so the check was always false and no badge rendered. They now pass when every present score is 1.0. Same 8-line fix in all seven files; commented once here.

app/web_ui/src/lib/utils/link_builder.ts
```diff
@@ -76,6 +77,7 @@ export function tool_set_type_label(type: ToolSetType): string {
     case "sandbox_code":
       return "Sandbox Code"
     default:
-      return type
+      // Compile error here means a new ToolSetType needs a label above.
```
chiang-daniel, line 80: **Decision made: an unknown tool-set type now throws instead of falling back to the raw string.**

**Where you see it:** the code-tool pages, if the server ever sends a tool-set type this build does not know.

`assertNever` makes a missing label a compile error, which is the house pattern. The type is generated from the server schema, so a server ahead of the client can send a value the client does not know and this now throws at render on the code-tool pages instead of showing the raw id. Confirm the trade for the kiln_server-backed path.

app/web_ui/src/lib/chat/streaming_chat.ts
```diff
@@ -557,108 +556,96 @@ export async function streamChat(options: StreamChatOptions): Promise<void> {
...
-                    `Tool execute API error ${postRes.status}: ${text || postRes.statusText}`,
-                  ),
-                )
-                return
-              }
-              const nextReader = postRes.body?.getReader()
-              if (!nextReader) {
-                onError(new Error("No response body from execute-tools"))
-                return
+      // A fresh line generator per reader: the stream is swapped after
```
chiang-daniel, line 559: **Refactor of the chat transport, made alongside the builder work.**

**Where you see it:** the /assistant page and the layout chat bar.

The hand-rolled SSE reader (decoder, byte buffer, line split, `data:` prefix check) is replaced by the shared `sse_data_payloads` generator. That is 31 real lines; the other ~150 lines in the diff are re-indentation from the removed inner `while` loop, nothing else. Parser behaviour was compared line by line and is identical; no test in this area changed for it. The `/assistant` page and the layout chat bar are the consumers. Worth a look from the chat's owner because nothing in the builder needed this file.

app/web_ui/src/lib/components/import/import_project.svelte
```diff
@@ -103,6 +103,14 @@
   }
 
   function redirect_if_missing_state(step: WizardStep): boolean {
+    // The local trust step depends on the component-local selected file path,
```
chiang-daniel, line 106: **Bug fix in git import, with a side effect to check.**

**Where you see it:** git import, going from the file step to the trust step.

An empty local path could reach the trust-confirm step (a failed file picker); the new guard sends the user back to the file step, and a second guard stops the import itself. The first guard routes into a pre-existing early return that happens before the `hashchange` listener is registered, so a deep link to that step leaves Back/Forward dead for that mount, and it wipes the "can't open file selector" message that caused it. Two questions from the earlier review-only PR on this file (#1654) were never answered.

app/web_ui/src/lib/utils/code_tool_helpers.ts
```diff
@@ -58,51 +58,6 @@ function jsonTypeToPython(prop: JsonSchemaProperty): string {
   }
 }
 
-const PYTHON_RESERVED_WORDS = new Set([
```
chiang-daniel, line 61: **Bug fixes in code-tool stub generation, seen before in #1655.**

**Where you see it:** the code-tool wizard: the generated stub, and Back from the code step.

Removed: the reserved-word renamer that produced `def run(from_param)` while the schema still said `from` (the tool could never be called; the datamodel now rejects such names, #1778). Fixed: docstring escaping (a description ending in `\` or containing `"` broke the stub). Added (line 159): `resolveStep2Code`, which decides the editor contents from the code state, not the wizard step, so Back no longer wipes authored code (the page half is 6/7). Open point from that review: a description-only edit still fires the "schema changed" hint.

app/web_ui/src/lib/components/eval_types/tool_call_check_form.svelte
```diff
@@ -136,6 +136,34 @@
...
+      return true
+    } catch {
+      return false
+    }
+  }
+
+  // Authoring-time check for one argument row. A value with no name, or a
+  // non-empty value that isn't valid JSON, is an error the author must fix --
+  // rather than being silently dropped or coerced to a raw string at save time.
+  function validate_arg_row(row: ArgRow): {
```
chiang-daniel, line 151: **Bug fix on the tool-call judge form: an argument value that is not JSON is an authoring error, not silently stored as a string.**

**Where you see it:** the tool-call judge form, the argument rows.

**Why:** the form let an author type a value that failed to parse and saved it as a raw string, so the check compared against a string the author never meant; a value with no name was dropped without a word. Both now show an inline error on the row and block save and test until fixed. An empty value still means "no value constraint".

sfierro (approved):

leonardmq (approved):

PR #1787, opened 2026-09-16
[Eval V2 Kiln, review-only 6/7] Existing pages: evals list, results, compare, generate, dataset

chiang-daniel:
**Review-only.** The goal of this review: the eval entry flow and the results page changed on purpose; every other page is the same or fixed. Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with these routes reverted to the baseline, so "Files changed" is every existing page the branch touched, plus the e2e spec and a few docs: 71 source files (+1402 / −2147) and 23 test files. The builder's own pages are 7/7. Details are inline on each hunk.

## How it works

These are pages that shipped before the branch. Most changes are one of four kinds: the eval-creation entry moved (one picker page, then the builder or the manual form; the old "choose your workflow" page and the manual form's Kiln Pro wizard are gone), the evals pages learned about conversations (a chat view on result rows, a banner where stored conversations score identically, no "add data" on builder-made evals), four pages that spun forever on a failed task load now show an error, and the generate lane got shared components and a Data Guide return path so the builder could reuse them. About 45 one-line hunks are a prop rename on the shared warning control; skip those.

```mermaid
flowchart LR
  Evals["Evals list: create / continue draft"] --> Pick["select_template (7/7)"]
  Pick --> Builder["builder (7/7)"]
  Pick --> Manual["spec_builder: manual form only (Pro wizard removed)"]
  Evals --> Detail["eval detail: no Add Eval Data on builder-made evals"]
  Detail --> Results["run_result: conversation rows + trace dialog"]
  Detail --> Compare["compare_run_configs: stored-conversation banner"]
  Gen["generate: shared batch form, Refine Plan, Data Guide return path"] -.-> Builder
  Run["dataset run page: fork fixes"]
  Load["fine_tune ×2, generate, prompt_optimization: error instead of endless spinner"]
```

- **Evals list**: the multi-turn block is gone; the create button reads "Continue Eval Draft" when a resumable draft exists (2 s peek); a Copilot check decides the destination. One finding: a failed Copilot check blanks the page.
- **Eval results**: rows with a renderable trace show a three-line preview and "View Full Trace" instead of inline input and output, for single-turn evals too; a dangling trace reads "Trace unavailable".
- **Compare pages**: a banner when stored multi-turn conversations are in the set; "Add Eval Data" replaced by text on builder-made evals (two guards left unreachable).
- **Entry flow**: `select_workflow` deleted; the manual form loses its Kiln Pro wizard (two components deleted); Pro connect returns to the type picker.
- **Generate**: batch form and Data Guide offer extracted for reuse; "New Batch Plan" is "Refine Plan"; the guidance box is smaller; every Data Guide page returns to whoever opened it.
- **Dataset run page**: fork handles use the server's turn boundaries (chain-of-thought safe); no flash of the full conversation mid-fork; the CSV dialog documents the route's task, not the selected one.
- **Load failures**: fine-tune list, create fine-tune, generate landing, prompt optimization.
- **Also in this view**: 45 `tight` → `inline` renames across 18 files, two spacing lines, the e2e spec, five docs.

## Why this PR is where it is

Every user has these pages today. Nothing here is behind the Pro gate: the results page change and the entry-flow change reach single-turn users on the day of release. Read the evals list, the results page and the entry flow first; the generate lane and the load-failure fixes can be skimmed.

## How to comment


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -51,14 +60,20 @@
   let evals_loading = true
   let eval_load_error_count = 0
 
-  // Evals aren't supported for multi-turn tasks, so the page needs the task's
-  // turn mode before it can decide between the table and the unsupported notice.
-  let task: Task | null = null
-  let task_loading = true
-  $: is_multiturn = task?.turn_mode === "multiturn"
+  let settings_loading = true
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:63]: **Behaviour change: multi-turn tasks see the evals table.**

**Where you see it:** the evals list of a multi-turn task.

**As before:** the page loaded the task and, for a multi-turn task, replaced the table with "Evals are not supported for multi-turn tasks."

**What changed:** the task load and the gate are gone; multi-turn tasks reach the evals list and the builder like any other. The state that replaced it: a Copilot availability check and a draft peek, both part of the page's loading gate so the create button is painted with the right label the first time.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -51,14 +60,20 @@
...
+  // The draft peek is in the page's loading gate, not outside it: its answer
+  // is what the create button is LABELLED, so a button rendered before it
+  // lands says "Create Eval" and then rewrites itself a few ms later. Held
+  // with the rest, the button and the body arrive together and the label is
+  // right the first time it is painted.
+  let draft_loading = true
 
-  $: loading = specs_loading || evals_loading || task_loading
-  $: error = specs_error || evals_error
+  $: loading =
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:74]: **Bug: a failed Kiln Copilot check now hides every eval on the page.**

**Where you see it:** the evals list, when the Kiln Copilot check fails or times out.

**What this is:** `settings_error` is folded into the page-level error, and the error branch (line 805) replaces the whole table with one error line. The check it comes from throws on any studio failure and has a 10 s timeout.

**What the user sees:** offline, or with a wedged verify call, the evals list shows "Failed to verify Kiln Copilot API key" and nothing else, though the evals loaded fine. The draft peek three lines below swallows its own failure on purpose ("no draft signal is ever worth an error surface"); the manual form makes the same choice for the same call. No test covers this path.

**Suggested fix:** treat a failed check as "no Copilot" and keep the table.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte
```diff
@@ -183,18 +198,83 @@
...
+    const deadline = new Promise((_, reject) => {
+      timer = setTimeout(
+        () => reject(new Error("draft check timed out")),
+        DRAFT_CHECK_TIMEOUT_MS,
+      )
+    })
+    return Promise.race([peek, deadline]).finally(() => clearTimeout(timer))
+  }
+
+  // Whether the v2 builder has a resumable draft for this task — the
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/+page.svelte:229]: **New: the create button knows whether a builder draft is waiting.**

**Where you see it:** the evals list, the create button.

**What this is:** per task, the page opens the builder's draft store and asks whether the draft is resumable, under a 2 s deadline (an IndexedDB open can hang when another tab holds the database at a different version). The answer labels the button ("Continue Eval Draft") and picks the destination: the builder for a Copilot user with a draft, the type picker otherwise. A failed or slow peek means "no draft". Re-runs when the task changes because the component stays mounted across tasks.

**Note:** the evals list now imports from the builder directory (`builder_draft`), a dependency an existing page did not have.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/+page.svelte
```diff
@@ -880,14 +880,22 @@
...
                         </div>
-                        <button
-                          class="btn btn-sm {current_step_id == 'eval_data'
-                            ? 'btn-primary'
-                            : ''}"
-                          on:click={add_eval_data}
-                        >
-                          Add Eval Data
-                        </button>
+                        {#if eval_split(evaluator, "test")?.source === "eval_input"}
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/+page.svelte:883]: **Behaviour change: no "Add Eval Data" on an eval whose test set the builder minted.**

**Where you see it:** an eval's page, the Add Eval Data button, on evals the builder created.

**What this is:** adding data tags dataset runs under the test split's tag. A builder-made eval's test split is a set of generated cases, not runs, so there is nothing to tag under. The button is replaced by "This eval's data was created by the eval builder and can't be extended here."

**Dead code:** `add_eval_data()` (line 615) still opens with the same check and an alert with different wording ("our new eval dataset format"); its only call site is now inside the else arm, so the guard is unreachable. Same on the compare page (`compare/+page.svelte:1115`). Remove the guards or align the wording.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte
```diff
@@ -57,6 +59,133 @@
   let thinking_dialog: Dialog | null = null
   let displayed_result: EvalRunWithTrace | null = null
 
+  let trace_dialog: Dialog | null = null
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte:62]: **New view: result rows that carry a conversation open it in a dialog.**

**Where you see it:** the eval results page, any row whose run has a stored trace.

**What this is:** the results table for one judge on one run config. A row's stored trace is parsed once (memoised per row) and checked for shape: known roles only, string content on user and assistant turns, well-formed tool calls (the dialog reads `id` and `function.name` unguarded, so a malformed call would throw while opening), and at least one user or assistant turn.

**Decision made:** the row changes shape so a reviewer can open a stored conversation from the results table; the question is which rows. The gate is renderability alone, "whatever the task's turn mode". A single-turn eval whose runs recorded transcripts (tool calls and all) converts too: every row shows the first user message clamped to three lines and a "View Full Trace" link, and the inline Output disappears for those rows. Single-turn evals previously stayed on the flat view even when traces existed; this reverses that, and no comment says why. Reading a model's answer on such a page now means opening the dialog per row.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte
```diff
@@ -306,24 +439,74 @@
...
+                      >
+                        View Full Trace
+                      </button>
+                    </div>
+                  </div>
+                {:else}
+                  <div class="font-medium">Input:</div>
                   <div>
-                    {result.eval_run.reference_answer}
+                    <!-- Nullable: a skipped run whose dataset item is gone.
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/[eval_config_id]/[run_config_id]/run_result/+page.svelte:483]: **Bug fix: a dangling trace no longer renders as a blank output.**

**Where you see it:** the eval results page, a row whose run's trace is missing.

A score record that points at a run whose trace is gone (deleted, or not included in an import) showed an empty Output cell. It now reads "Trace unavailable. It may have been deleted or not included in an import." The column header is "Interaction" when any row renders as a conversation, so a mixed table can show that header above mostly flat rows (pinned by a test).

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte
```diff
@@ -1112,13 +1112,23 @@
...
                                 <div class="text-left">
                                   {#if getEvalDatasetSize(section.eval_id) === 0}
-                                    <button
-                                      class="btn btn-xs mt-1"
-                                      on:click={() =>
-                                        navigateToAddData(section.eval_id)}
-                                    >
-                                      Add Eval Data
-                                    </button>
+                                    {#if eval_split(eval_data_cache[section.eval_id], "test")?.source === "eval_input"}
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/compare/+page.svelte:1115]: **Behaviour change: same "Add Eval Data" replacement on the cross-eval compare page.**

**Where you see it:** the cross-eval compare page, the same button.

Text instead of a button for builder-made evals. `navigateToAddData` (line 673) keeps its own now-unreachable guard and alert (see the eval page, `[eval_id]/+page.svelte:883`).

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/compare_run_configs/+page.svelte
```diff
@@ -471,6 +476,17 @@
         ]}
     action_buttons={action_buttons(evaluator)}
   >
+    <!-- Stored multi-turn conversations can't be regenerated per run config,
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/[spec_id]/[eval_id]/compare_run_configs/+page.svelte:479]: **New banner: stored conversations score the same under every run config.**

**Where you see it:** the compare run configs page, when the set holds stored multi-turn conversations.

A stored multi-turn conversation cannot be regenerated per run config, so the runner judges its saved messages and those rows score identically across configs (2/7 explains why). The score summary reports how many such items the set holds; the banner shows when it is above zero and points at the builder for fresh conversations. Copy in `multi_turn_warning.ts`, four tests.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte
```diff
@@ -1,223 +0,0 @@
-<script lang="ts">
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_workflow/+page.svelte:1]: **Deleted page: "Choose your Eval Creation Workflow".**

**Where you see it:** the old "choose your workflow" URL.

**What it was:** the Pro-versus-manual question between the template picker and the form, with its own redirects (no template back to the picker; tool-bearing tasks straight to manual; "Use Kiln Pro" to the connect page).

**Where users go now:** the question is gone. The picker leads to the builder or the manual form directly; the helper that produced this page's URL now returns the form's URL.

**Known wart:** the old URL falls into the sibling dynamic route and renders "Spec not found". Nothing in the app links to it any more, so it was left as is; the e2e spec pins that behaviour.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/+page.svelte
```diff
@@ -123,155 +66,40 @@
...
-      } catch {
-        has_kiln_copilot = false
-      }
-
-      // Load available models for later API calls
-      await load_available_models()
-
-      // Load the task (used by copilot API calls)
-      task = await load_task(project_id, task_id)
+      // Confirm the route's task exists before showing the form, so a stale
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/spec_builder/+page.svelte:96]: **Removed: the manual form's Kiln Pro wizard.**

**Where you see it:** Create Eval, the manual form.

**As before:** one page ran a multi-state flow: create, questions, analyse, review examples, refine, save with Copilot.

**What changed:** the page is the plain "Create Eval" form. Gone with it: the Copilot availability check, model loading, the tool check, the abort controller, the `workflow` query parameter (a `?workflow=pro` URL is now inert), and roughly 900 lines of flow. The one network call on entry loads the task so a stale link errors here instead of on save. Two components are deleted (`refine_spec.svelte`, `review_examples.svelte`) and the form's submit never reads "Create with Kiln Pro". No dangling references remain; the copilot spec routes are called only from the builder now.

app/web_ui/src/routes/(app)/specs/pro_auth/+page.svelte
```diff
@@ -15,18 +11,12 @@
...
-  $: judge = $page.url.searchParams.get("judge") as V2EvalType | null
+  // A successful connect returns to eval creation at the type picker — or the
+  // home page if ui_state has no current project/task to build a URL from.
   $: success_redirect_url =
-    spec_type && judge && project_id && task_id
-      ? spec_builder_url(project_id, task_id, spec_type, "pro", judge)
-      : project_id && task_id
-        ? `/specs/${project_id}/${task_id}/select_template`
-        : "/"
+    project_id && task_id
```
chiang-daniel, line 17: **Behaviour change: connecting Kiln Pro returns to the type picker.**

**Where you see it:** Kiln Pro connect, where it returns to afterwards.

Before, a successful connect returned to the form in Pro mode with the chosen template preselected. Those params are no longer read; the redirect is the picker (or home when no task is current). The form's Pro mode no longer exists.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte
```diff
@@ -74,10 +76,22 @@
 
   async function handle_routing(req_project_id: string, req_task_id: string) {
     loading = true
+    load_error = null
 
     // Synthetic data generation is single-turn only. For multi-turn tasks
     // don't redirect into the synth/qna flows — show the disabled notice.
-    const loaded_task = await load_task(req_project_id, req_task_id)
+    let loaded_task: Task | null
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/+page.svelte:83]: **Bug fix: a failed task load shows an error instead of an endless spinner.**

**Where you see it:** the /generate landing page when the task fails to load.

The routing awaited `load_task` with no catch; any fetch failure left the landing page spinning. Now the failure lands in `load_error` and an error branch renders (line 221). The routing key stays marked as handled so a failing load is not retried in a hot loop; navigating to another task or remounting retries. Raised in #1650; its one open point is the missing `role="alert"` on the new error container (same on the four sibling error branches added here).

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/kiln_pro_batch_plan.svelte
```diff
@@ -1,43 +1,97 @@
...
   import KilnProPlanSummary from "./kiln_pro_plan_summary.svelte"
   import KilnProPromptsTable from "./kiln_pro_prompts_table.svelte"
+  import SettingsHeader from "$lib/ui/settings_header.svelte"
 
   export let plan: { prompts: string[]; summary: string }
   export let on_generate_inputs: () => void
   export let on_regenerate: () => void
   export let on_delete_prompt: (index: number) => void
   export let summary_out_of_sync = false
+  // Optional override for the generate button's label. The eval builder's
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/kiln_pro_batch_plan.svelte:11]: **Behaviour change on /generate: the plan header is rebuilt and its button renamed.**

**Where you see it:** /generate, the batch plan header and its buttons.

Nine new props with "default keeps /generate unchanged" comments make the surface reusable by the builder. Two things do change for /generate users: the hand-rolled title block becomes the house `SettingsHeader`, and the regenerate button reads "Refine Plan" instead of "New Batch Plan" (the summary panel is "Overview" instead of "Batch Overview"). The confirm-before-regenerate now lives in each parent.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/kiln_pro_batch_form.svelte
```diff
@@ -0,0 +1,71 @@
...
+<div class="flex flex-row items-center gap-4">
+  <div class="flex-grow font-medium text-sm">{count_label}</div>
+  <IncrementUi bind:value={count} max={count_max} />
+</div>
+<FormElement
+  id={guidance_id}
+  label="Guidance"
+  description={`This allows you to control the dataset you are generating. For example, "10% of the dataset should be in Spanish."`}
+  inputType="textarea"
+  height="medium"
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/kiln_pro_batch_form.svelte:61]: **Decision made: the guidance box on /generate is a third of its old height.**

**Where you see it:** /generate, the batch guidance box.

The count row and guidance textarea were extracted into this shared component. It defaults to `height="medium"` (`h-36`); the code it replaced passed `height="xl"` (`h-96`). The change was made for the builder's dialog; it reaches the /generate page through the default. Say if the page should keep the tall box.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_kiln_pro.svelte
```diff
@@ -256,26 +257,17 @@
...
-        height="xl"
-        bind:value={batch_guidance}
-        inline_action={batch_guidance_template &&
-        batch_guidance !== batch_guidance_template
-          ? {
-              handler: () => (batch_guidance = batch_guidance_template),
-              label: "Reset",
-            }
-          : null}
+      <KilnProBatchForm
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/synth_kiln_pro.svelte:260]: **What changed on the batch page: the shared form, and a field moved.**

**Where you see it:** /generate, the batch page, the order of its fields.

The inline count row and guidance box become the shared form (`kiln_pro_batch_form.svelte:61`), and the Data Guide checkbox moves up to sit right after it. Existing users see the reorder and the shorter box.

app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/data_guide/+page.svelte
```diff
@@ -18,6 +18,11 @@
   import { isMacOS } from "$lib/utils/platform"
   import { pending_data_guide_refine_handoff } from "./refine_handoff_store"
   import posthog from "posthog-js"
+  import {
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/generate/[project_id]/[task_id]/data_guide/+page.svelte:21]: **Behaviour change across the Data Guide chain: every page returns to whoever opened it.**

**Where you see it:** every Data Guide page: breadcrumbs, redirects and the "return to" sentences.

Seven pages read a `caller` query key and compute a return target (label, href, view href) from it; every breadcrumb, redirect and "return to ..." sentence uses that instead of a hardcoded "Synthetic Data Generation". The OAuth hop carries the caller in a stash because its URL is fixed. With no key the target must resolve to the synth page for existing users; that helper lives in `lib/utils/data_guide_return.ts` (5/7), so confirm it there. One real fix rides along: a redirect that appended `?draft_failed=1` after an already-query-bearing URL.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/fork_helpers.ts
```diff
@@ -10,42 +10,29 @@ export type ForkTarget = {
...
-  }
-  // Suffix-align the chain against the user messages: the last entry (the
-  // leaf) corresponds to the last user message. A negative offset would
-  // mean the server returned more chain entries than the trace can support
-  // — caller guards against that by capping chain length to turn_count.
-  const offset = total_turns - chain.length
-  if (offset < 0) return result
-  for (let k = 0; k < chain.length; k++) {
-    const entry = chain[k]
+  for (const entry of chain) {
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/fork_helpers.ts:26]: **Bug fix: fork handles land on the right assistant message with chain-of-thought prompts.**

**Where you see it:** the dataset run page, the fork handles on a chain-of-thought task.

**The bug:** turns were located by counting user messages; a chain-of-thought strategy emits two user messages per turn, so the count drifted and fork handles sat on the wrong bubbles or vanished. A broken chain was suffix-aligned, silently mapping the wrong turns.

**The fix:** each chain entry's server-computed `trace_start_index` locates its turn; turn 1 (the root, or the first entry of a broken chain) is never forkable. Tests replaced accordingly.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte
```diff
@@ -702,6 +709,12 @@
   let run_has_children = false
   let chain_loaded_for_run_id: string | null = null
   let fork_target: ForkTarget | null = null
+  // Holds the fork point's truncation across the navigation to the forked run.
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/[run_id]/run/+page.svelte:712]: **Bug fix: forking no longer flashes the whole original conversation mid-send.**

**Where you see it:** the dataset run page, while a fork is being created.

`fork_target` resets the instant the run id changes, so the transcript snapped back to the full conversation while the forked run was being created. A separate `fork_send_trace_index` holds the truncation from send start until the new run renders, with a pending indicator; a failed fork releases it and keeps the composer open at the same point.

app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte
```diff
@@ -1,17 +1,22 @@
 <script lang="ts">
   import { client } from "$lib/api_client"
   import { page } from "$app/stores"
-  import { current_task } from "$lib/stores"
+  import type { TurnMode } from "$lib/types"
   import Dialog from "$lib/ui/dialog.svelte"
 
   export let onImportCompleted: () => void
   export let tag_splits: Record<string, number> | null = null
+  // The route's task turn mode, threaded in by the owning page. The CSV
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/dataset/[project_id]/[task_id]/upload_dataset_dialog.svelte:9]: **Bug fix: the CSV upload dialog documents the task it imports into.**

**Where you see it:** the dataset page, Add Data, the CSV upload dialog.

It read the app-wide current-task store while the import posts to the route's task, so opening Add Data for another task documented the wrong columns (single-turn versus multi-turn). The page now loads the route's task and passes its turn mode in.

app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/+page.svelte
```diff
@@ -43,9 +45,23 @@
     req_project_id: string,
     req_task_id: string,
   ) {
-    const loaded = await load_task(req_project_id, req_task_id)
-    if (req_project_id !== project_id || req_task_id !== task_id) return
-    task = loaded
+    try {
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/fine_tune/[project_id]/[task_id]/+page.svelte:48]: **Bug fix: an error instead of an endless spinner when the task cannot load.**

**Where you see it:** the fine-tune list, create fine-tune and prompt optimization pages when the task fails to load.

Same shape on the create-fine-tune page and the prompt-optimization page: the task load is caught, the spinner condition excludes the error state, and an error branch renders (a "Load failed" message is reworded to "This task may belong to a project you don't have access to"). On the prompt-optimization page the multi-turn notice now waits for a loaded task instead of flashing the single-turn UI from a null default.

app/web_ui/src/routes/(app)/tools/[project_id]/add_tools/code_tool/+page.svelte
```diff
@@ -145,26 +145,21 @@
...
-      // Returning from step 1 after editing schema
-      if (isCodeUnmodified(code, generated_placeholder)) {
-        code = new_placeholder
-        generated_placeholder = new_placeholder
-        schema_changed_hint = false
-      } else if (new_placeholder !== generated_placeholder) {
-        generated_placeholder = new_placeholder
-        schema_changed_hint = true
-      }
+    // Decide the editor contents from the code state itself, not current_step:
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/tools/[project_id]/add_tools/code_tool/+page.svelte:148]: **Bug fix: returning to the code step no longer wipes authored code.**

**Where you see it:** the code-tool wizard, Back from the code step.

The editor's contents were chosen from `current_step`, but browser Back pops the shallow-routing step so it read "define" on a return, and the code-preserving branch was unreachable: coming back replaced the user's code with a fresh placeholder. The decision now keys on the code state itself (a pure helper in 5/7, seen in #1655); the clone seed is consumed once. Open point from that review: the "schema changed" hint also fires for a description-only edit.

sfierro (approved):

leonardmq (approved):

PR #1788, opened 2026-09-16
[Eval V2 Kiln, review-only 7/7] Builder pages: the new five-step wizard and claim review

chiang-daniel:
**Review-only.** The goal of this review: the wizard's decisions about money and correctness are right; the screens themselves were clicked through in the bash. Nothing here merges; fixes land on `dchiang/eb-v2-merge` and ship through #1731. The head is that branch byte for byte; the base is the same tree with the builder's two route directories removed, so "Files changed" is the wizard itself: 15 source files (+9326) and 13 test files. Everything here is behind the Kiln Pro gate. Details are inline; the page is large, so the comments point at the decisions, not every line.

## How it works

The eval builder is one page with five steps held in browser history (Back and Forward walk the steps without unmounting). The entry page asks what the eval should check and which run config to evaluate, then hands over. Step 1 describes, Step 2 asks clarifying questions (the v1 questions form, reused), Step 3 shows the refined name and issue description, Step 4 plans a batch of cases and drives them (multi-turn: synthetic users talk to the agent; single-turn: inputs are minted and run), and Step 5 shows one review card per case where the reviewer agrees or disagrees with each claim. A disagreement offers to improve the judge: the judge prompt is refined from the grades, every case is re-judged, and the review repeats until the grades agree or the reviewer saves without improving. Save writes the spec, the judge that was reviewed, the cases and the reviews. A draft in IndexedDB restores everything up to the plan screen.

```mermaid
flowchart LR
  Entry["select_template: description + run config"] --> S1["1 describe"] --> S2["2 questions"] --> S3["3 refine"] --> S4["4 plan + drive"] --> S5["5 claim review"] --> Save["save"]
  S5 -- "any disagree" --> Improve["improve judge: refine, re-judge all, review again"] --> S5
  S4 -.-> Plan["plan_flow.ts: top-off, preflight, stop screen"]
  S5 -.-> CE["claim_evidence.ts: citations, subset, save payload, calibration"]
  S1 & S3 & S4 -.-> Draft["builder_draft.ts: IndexedDB draft, what restores, what Reset keeps"]
  S5 --> Cards["claim_evidence_review + claim_card + claim_trace_modal"]
```

- **`+page.svelte`** (5492 lines): the state machine; the policy constants and decisions are commented at their lines.
- **`plan_flow.ts`**, **`claim_evidence.ts`**, **`builder_draft.ts`**: the pure logic behind Steps 4, 5 and the draft; the best-tested code here (85, 112 and 80 cases).
- **Step 5 screens**: `claim_evidence_review.svelte`, `claim_card.svelte`, `claim_trace_modal.svelte`, `claim_text.svelte`, `review_intro.svelte`, `progress_count.svelte` (Steve's).
- **`select_template/+page.svelte`**: the entry page, restructured around the description box and the run-config dialog.
- **Small**: `batch_plan_guidance.ts` (the planner prompt policy), `data_guide_flow.ts`, `refine_fields.ts`.

## Why this PR is last

Nothing here is reachable without Kiln Pro, and nothing outside the builder imports from it except the evals list (the draft helpers). It is the largest view and the one with the least prior review; read the decisions first, then the two drive functions (they are the same eight stages twice), then Step 5's policy module.

## How to comment

Mark **Blocker** or **Later**. Blockers are fixed on `dchiang/eb-v2-merge` in one round and this view is regenerated. Inline markers: **Bug** = fix it; **Decision made** = built this way on purpose, say yes or push back; anything else is a note. The page's own test file asserts on the page's source text, not its behaviour; treat the pure modules' tests as the coverage.


app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+  function is_abort_error(error: unknown): boolean {
+    return error instanceof DOMException && error.name === "AbortError"
+  }
+
+  // Deadline on the judge-copilot calls (authoring and the loop's refine):
+  // one full server attempt. The server retries for up to ~20 minutes for
+  // callers that can afford to wait; an interactive flow cannot, so past this
+  // the call is surfaced as failed — authoring as a retryable drive error,
+  // refine as the inline error under the review actions.
+  const JUDGE_COPILOT_DEADLINE_MS = 300_000
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:279]: **Decision made: a 5-minute deadline on the two judge copilot calls.**

**Where you see it:** Steps 4 and 5, the judge authoring and refine calls.

`JUDGE_COPILOT_DEADLINE_MS = 300_000`. The server's own retry budget for a copilot call is around 20 minutes; an interactive step will not wait that long, so the deadline fires a retryable error instead. The number is asserted, not derived; the other long call on the page (`+page.svelte:2731`) has no deadline at all.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+    if (nav.type === "leave") return
+    if (
+      !confirm(
+        "Leave the eval builder? Your answers, generated data, and review progress here will be lost.",
+      )
+    ) {
+      nav.cancel()
+    }
+  })
+
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:396]: **What is saved, where, when: the draft.**

**Where you see it:** every step; this is the draft that "Continue Eval Draft" restores.

Twenty-three fields mirror into IndexedDB under `eval_builder_draft_{project}_{task}_v1` on every change, gated on the restore having run. Restore happens once on mount and never goes past the plan screen (driven and review state depend on disk that can change). Save clears the draft; Reset keeps the batch tags (`+page.svelte:2476`). Write failures are only logged to the console, so a quota failure (a 50-case draft is a few MB) is silent: the user believes their work is saved.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+            task_prompt: task?.instruction ?? "",
+            task_input_schema: task?.input_json_schema ?? "",
+            task_output_schema: task?.output_json_schema ?? "",
+          },
+          target_specification: source,
+        },
+        signal: new_copilot_abort_signal(),
+      })
+      if (error || !data) {
+        questions_error = "Failed to load clarifying questions."
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:784]: **Bug: the server's error is discarded on the questions call.**

**Where you see it:** Step 2, when loading the clarifying questions fails.

`questions_error = "Failed to load clarifying questions."` and the error object is dropped. The sibling call the branch fixed later (`generate_cases`) unwraps the server's code and message and appends them; this site and the batch-plan site (`+page.svelte:1940`) did not get the same treatment. A 422 or a 502 reads as the same generic line.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+  let authored_judge_cache: {
+    spec_text: string
+    task_prompt: string
+    prompt: string
+  } | null = null
+
+  // How many extra authoring calls an unusable prompt is worth. An invalid
+  // authored prompt is model variance on a paid call, so one automatic re-ask
+  // recovers most of them without hiding a persistent failure behind spend.
+  const JUDGE_AUTHOR_RESAMPLE_ATTEMPTS = 1
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:1615]: **Decision made: a judge prompt is always authored, with one automatic resample.**

**Where you see it:** Step 4, judge authoring before the drive.

The builder never falls back to a generic judge: authoring is required, and if the copilot's first attempt is unusable it is asked once more (`JUDGE_AUTHOR_RESAMPLE_ATTEMPTS = 1`). The reviewed judge is what save persists (`+page.svelte:4077`).

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+            // The Data Guide and the grounding sample ride the planner's
+            // data-guide param (multi-turn plans scenarios, not inputs — no
+            // guide there).
+            data_guide: is_multi_turn ? null : single_turn_data_guide(),
+          },
+          signal: new_copilot_abort_signal(),
+        },
+      )
+      if (error || !data) {
+        generation_error = "Failed to draft a batch plan."
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:1940]: **Bug: the server's error is discarded on the batch-plan call.**

**Where you see it:** Step 4, when drafting the batch plan fails.

Same as the questions call (`+page.svelte:784`): `generation_error = "Failed to draft a batch plan."` with the error dropped. This is the surface a stranded-batch 422 (`+page.svelte:2476`) lands on, so the user sees a generic line for a specific cause.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+
+  // Step 4 (multi-turn) part 2 — drive from the approved plan. The approved
+  // prompts become synthetic-user cases in ONE batch call (case i ← prompt i
+  // via generate_cases' case_prompts), then a single multi_turn_pipeline stream
+  // runs [drive → judge → claims] per case — each case flows through
+  // independently, so the plan rows light up as their case progresses.
+  // Ping every lane concurrently (~2s total); the returned failure is the
+  // first in blame order, not race order. Each lane costs about a token —
+  // nothing next to the batch spend a dead lane would waste.
+  async function preflight_lanes(
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:2101]: **Decision made: every lane is preflighted before any spend.**

**Where you see it:** Step 4, before any case is generated.

The synthetic-user model, the agent's run config and the judge model are each checked with a cheap call before cases are generated; the first failure, in blame order, stops the drive with a named lane. A dead key costs nothing.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+            driven_slots.add(slot)
+            driven_cases = [...driven_cases, drive_plan.items[event.case_index]]
+          }
+          // Chains exist on disk under this batch's tag from here on —
+          // record it immediately so an abort can't orphan the batch.
+          if (
+            multi_turn_batch_tag &&
+            !undeleted_batch_tags.includes(multi_turn_batch_tag)
+          ) {
+            undeleted_batch_tags = [
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:2476]: **Bug: the list of batches to delete grows without a cap, against a server limit of 20.**

**Where you see it:** Step 4, the drive after a failed or aborted one.

**What this is:** each drive that put anything on disk records its batch tag; the next drive passes the list so the server deletes the stranded batches. The list rides the draft and survives Reset on purpose (a wiped draft would orphan chains forever).

**The bug:** it is pruned only when a drive completes cleanly, never for an aborted or failed drive, which is exactly the drive that strands a batch. The desktop route declares `replace_batch_tags` with `max_length=20` (verified). After 21 accumulated tags every drive is a 422, shown as the generic batch-plan line, and Reset keeps the tags. Prune on failure too, or cap the list.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+          )
+        for (const r of data.results) {
+          if (r.error) case_failure_messages.push(r.error)
+        }
+        if (inputs.length === 0) {
+          throw new KilnError("No eval data could be written. Try again.")
+        }
+        return inputs
+      }
+      await new Promise((resolve) => setTimeout(resolve, 750))
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:2731]: **Bug: the single-turn minting poll has no deadline.**

**Where you see it:** Step 4, single-turn, while inputs are being minted.

`mint_inputs_from_plan` polls the batch job every 750 ms in a `for (;;)` with no iteration cap or elapsed bound; only Back (which aborts the signal) ends it. A job stuck in `running` server-side holds the screen indefinitely. The judge calls got a deadline (`+page.svelte:279`); this did not. The 750 has no rationale.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+    driven_prompts_json = null
+    replace_step("done")
+  }
+
+  // The page the wizard leaves for once it is finished. The eval itself when
+  // there is one, its list when the save returned no id.
+  $: finished_destination =
+    created_eval_href ?? `/specs/${project_id}/${task_id}`
+
+  function on_advance_to_save() {
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:3331]: **Decision made: any disagreement offers to improve the judge, and the loop has no round cap.**

**Where you see it:** Step 5, after a disagreement, the Improve Judge dialog.

**What this is:** after the review, one or more disagreements open the "Improve Judge with Feedback?" dialog. Accepting refines the judge prompt from the grades, re-judges the whole batch, and reopens the review on a new subset. The only exits are zero disagreements or "Save Without Improving".

**To confirm:** `calibration_rounds_completed` (line 3952) is counted for telemetry and the gate, never compared to a maximum. An oscillating judge loops until the reviewer gives up, each round costing a refine plus a full re-judge. A verdict-only trigger was tried and reverted; any disagreement is the decision.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+  // that failed to re-judge are simply absent. Throws on stream-level
+  // failure (bad response, batch_failed, batch_aborted, or nothing judged) —
+  // partial results from a broken stream are never applied.
+  async function rejudge_all_traces(
+    judge: JudgeConfig,
+  ): Promise<Map<number, RejudgeCaseResult>> {
+    // Every driven case with a durable run id, in plan order. The runner
+    // emits "" for a run without an id — such a case can't be reloaded, so
+    // it sits the round out like a failed case.
+    const entries = trace_claims
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:3781]: **Decision made: a calibration round re-judges the whole batch, not the reviewed subset.**

**Where you see it:** Step 5, during a calibration round.

Fresh verdicts everywhere let the next subset be picked freely (disagreed first, then flipped, then fresh). At 200 cases that is 200 judge calls per round. The alternative, re-judging only the reviewed cases, was rejected because the next subset would then mix old and new verdicts.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte
```diff
@@ -0,0 +1,5492 @@
...
+      // lanes and its cost like every other.
+      void open_drive_settings()
+      return
+    }
+    // Claims are lazy on both arms — gate the advance on the selected
+    // traces being fully resolved (instant when already built).
+    start_claims_gate()
+  }
+
+  // ── Step 6 state — save
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/+page.svelte:4077]: **Decision made: save persists exactly the reviewed judge, the cases and the reviews.**

**Where you see it:** Save, at the end of Step 5.

Multi-turn: the driven conversations are tagged into the eval's splits (train/val/golden); single-turn: the minted inputs and their runs are saved. In both, the judge that was reviewed is the one written, with no fallback; the spec's properties are filtered to the fields the form renders (`refine_fields.ts`) so a model-authored field the user never saw cannot reach disk. The run config is not sent (the save never read it).

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts
```diff
@@ -0,0 +1,1291 @@
...
+  reviews: TraceReview[],
+): number {
+  return traces.filter((t, i) => is_trace_reviewed(t, reviews[i])).length
+}
+
+// The reviewer's call on a trace: the verdict claim's grade IS the call —
+// agree keeps the judge's verdict, disagree flips it. Null while unanswered,
+// and null for a trace with no verdict claim, which the reviewer is never
+// shown. Nothing else records the call.
+export function human_verdict(
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts:845]: **Decision made: the reviewer's overall pass or fail is the verdict claim's grade, and nothing else records it.**

**Where you see it:** Step 5, the last claim on each card.

A card's last claim states the judge's verdict when it has one; agreeing or disagreeing with it is the human verdict. There is no separate verdict control. A case whose claims carry no verdict claim is silently dropped from the review walk (line 913); the reviewer is not told. Confirm both.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts
```diff
@@ -0,0 +1,1291 @@
...
+  const verdict = human_verdict(trace, review)
+  if (verdict === null) {
+    throw new Error(
+      "Cannot read the reviewer's verdict before the trace is graded.",
+    )
+  }
+  return verdict === "pass"
+}
+
+// ── Subset review (both arms) ────────────────────────────────────────────
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts:873]: **Decision made: how many cases the reviewer must grade before saving.**

**Where you see it:** Step 5, how many cases the reviewer must grade before Save.

`min(10, max(1, N/4))`: a quarter of the batch up to ten, flat as the batch grows; during a calibration round, capped by how many cases the round could surface so the gate cannot deadlock on cases the reviewer never saw. The subset is judge-stratified (fails take the odd slot). The doc comment for the round cap (line 889) sits above the wrong function; `calibration_gate_target` itself is undocumented.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts
```diff
@@ -0,0 +1,1291 @@
...
+
+// A resolved citation: the whole [start, end) span, plus where the `from`
+// anchor alone ends. from_end lets a caller that cannot place the whole span
+// fall back to the anchor the model actually wrote, instead of nothing.
+export type CitationSpan = {
+  start: number
+  end: number
+  from_end: number
+}
+
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts:170]: **Known limit: a citation whose quoted text occurs twice anchors to the first occurrence.**

**Where you see it:** Step 5, clicking a citation.

The resolver finds the citation's `from` text as a substring; repeated identical snippets can mis-locate the highlight with no signal to the reviewer. Documented, mitigation is on the model side (quote enough context). Sign off or ask for a heuristic.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts
```diff
@@ -0,0 +1,1291 @@
...
+    : haystack.map[first]
+}
+
+function span_end(haystack: FoldedText, last: number): number {
+  return /\s/.test(haystack.folded[last])
+    ? haystack.map[last]
+    : haystack.map_end[last]
+}
+
+// ── Structured-trace span mapping (MULTI-TURN ONLY) ──────────────────────
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts:414]: **Structural risk: a hand port of the Python transcript formatter.**

**Where you see it:** Step 5, citation highlighting in the trace dialog.

370 lines mirror `libs/core`'s `eval_trace_formatter.py` so a citation into the flattened transcript can be mapped back onto the structured conversation. Protected by a byte guard that yields no highlight on drift, which is a safe failure, but nothing ties the two implementations together at build or test time, and one case (a non-string tool output rendered through Python's `repr`) is documented as unreproducible. Accept, or ask for a shared fixture test.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts
```diff
@@ -0,0 +1,1291 @@
...
+  failed: number,
+  total: number,
+): string | null {
+  if (failed <= 0) return null
+  return `${failed} of ${total} couldn't be re-checked with the improved judge and kept their previous results. They were left out of this review round.`
+}
+
+// A judge prompt/rubric this long is almost certainly runaway model output,
+// not a rubric — reject it rather than persist it into the judge config.
+export const MAX_JUDGE_PROMPT_CHARS = 20000
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_evidence.ts:1252]: **Policy constant: a refined judge prompt over 20,000 characters is rejected.**

**Where you see it:** Step 5, after Improve Judge, if the refined prompt comes back huge.

`MAX_JUDGE_PROMPT_CHARS` guards against runaway model output; the number is not derived from any observed prompt. A rejection fails the calibration round after the refine was paid for. A wrapping code fence is unwrapped once rather than re-paid.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_trace_modal.svelte
```diff
@@ -0,0 +1,238 @@
...
+  // arms: a single-turn run is a conversation of one turn, and its tool calls
+  // and results are the same nodes a multi-turn trace carries.
+  $: use_chat = !!(trace && trace.trace && trace.trace.length > 0)
+
+  function text_for(source: CitationSource): string {
+    if (!trace) return ""
+    return source === "input" ? trace.raw_input : trace.raw_output
+  }
+
+  // Map a citation onto the structured trace so ChatTrace can mark the exact
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/claim_trace_modal.svelte:63]: **Decision made: a citation that cannot be mapped shows no mark, and only the console knows.**

**Where you see it:** Step 5, the trace dialog, when a citation cannot be placed.

Four failure kinds (anchor not found, flattener drift, row not rendered, span across two turns) each log a reason and draw nothing; the reviewer clicking `[3]` cannot tell a missing highlight from a hard-to-see one. "No highlight beats a wrong one" is the stated policy. Confirm, or ask for an on-screen note.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/plan_flow.ts
```diff
@@ -0,0 +1,491 @@
...
+// one-turn run they never picked. Those restore the default instead.
+export function restore_turns_per_case(
+  stored: number | null | undefined,
+  fallback: number,
+): number {
+  if (typeof stored !== "number" || !Number.isFinite(stored)) return fallback
+  return clamp_turns_per_case(stored)
+}
+
+// ── Top-off drive planning ────────────────────────────────────────────────
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/plan_flow.ts:90]: **Decision made: a retry tops off the missing slots instead of re-driving the batch.**

**Where you see it:** Step 4, Retry on the stop screen.

A top-off requires byte-identical items and an unchanged judge, user model and turn ceiling; otherwise it is a fresh batch and the old one is listed for deletion. `drive_lanes_unchanged` compares by `JSON.stringify`, key-order dependent; safe because both sides come from the same literal.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/batch_plan_guidance.ts
```diff
@@ -0,0 +1,99 @@
+// Batch-planner guidance for the eval builder's Step 4 — one function per
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/batch_plan_guidance.ts:1]: **Decision made: the planner's policy, in prose.**

**Where you see it:** Step 4, the batch plan the planner writes.

The batch-planner prompt asks for roughly half expected-pass and half expected-fail cases, ordinary users pursuing their own goal rather than testers probing the spec, and never an input that depends on a record already existing in the system under test. The spec is interpolated raw with no length cap. The user's steer is appended after the base text, which is a strict prefix the arm is identified by.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_template/+page.svelte
```diff
@@ -106,38 +120,288 @@
...
+  }
+
+  // Without Copilot the page opens on the offer rather than the templates.
+  // The Pro-vs-manual question is asked once, up front, instead of partway
+  // through after a template is already chosen. Choosing manual reveals the
+  // same list a Copilot user gets as the page's third section.
+  let chose_manual = false
+  // The offer is showing, as opposed to the picker behind it. Named once so
+  // the page title and the body can never describe different screens.
+  $: show_offer = $kilnCopilotConnected !== true && !chose_manual
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_template/+page.svelte:178]: **Bug: the Kiln Pro offer flashes for Copilot users while the check is in flight.**

**Where you see it:** the entry page, on load, for a user whose Copilot is connected.

`show_offer` is true whenever the connection store is not `true`, and the store is `null` while checking, so a connected user sees the offer and then the picker. The builder page treats `null` as a loading state; this page should too. The page also never initialises the store itself.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_template/+page.svelte
```diff
@@ -106,38 +120,288 @@
...
+    // empty dialog. Hand over as this page did before it asked.
+    if (!task) {
+      goto_builder(null)
+      return
+    }
+    run_config_dialog?.show()
+  }
+
+  // Continue needs a SAVED config: the wizard, the drive and the saved eval all
+  // refer to it by id, and "Custom" is only in the picker's local state. An
```
Comment (chiang-daniel) [inline app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/select_template/+page.svelte:224]: **Decision made: the eval is authored against one saved run config.**

**Where you see it:** the entry page, Write My Eval.

Write My Eval opens the run page's own picker; an unsaved custom selection is saved through the dropdown's path first, and the description and config id go to the builder in the URL (no length cap on the description; the browser's URL limit is the only bound). A task that fails to load means no picker and a silent fallback to the task default. The picker is given the task here while the builder's Generation Settings lane deliberately is not (the tool pickers there write to the app-wide store); only one of the two usages explains itself.

app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/batch_plan_guidance.ts
Comment (chiang-daniel) [reply app/web_ui/src/routes/(app)/specs/[project_id]/[task_id]/builder/batch_plan_guidance.ts:1]: @tawnymanticore this one is yours: what the planner is told to produce (half pass, half fail, ordinary users pursuing their own goal, never an input that depends on a record already existing).

sfierro (approved):

leonardmq (approved):

PR #1796, opened 2026-09-17
eb-v2 bug: eval builder gets stuck after too many failed drives

chiang-daniel:
## What does this PR do?

**TLDR:** Edge case in the eval builder: after twenty-one failed drives on one draft, every further drive was rejected. The cause is a defensive cap on the request, sized by guess. One line removed. Targets `dchiang/eb-v2-merge`.

- **Why there is a list:** a drive saves each conversation as a task run as it goes, like every run in Kiln, so an aborted drive leaves runs on disk under its batch tag. The page keeps a list of those tags, and the next drive hands it to the server to delete once it has produced replacements.
- **The bug:** the request model refused more than twenty tags. The cap was a defensive bound added with the field, sized by guess; the only cost it bounds is one pass over the task's runs per tag, and that only after a successful drive. The list only shrinks after a drive completes, so twenty-one failures in a row meant no drive could ever start again.
- **The fix:** drop the cap. Each stranded batch costs one pass over the task's runs after the next successful drive; twenty-five stranded batches cost twenty-five passes, once.

<details>
<summary><b>Verification</b></summary>

- New test: a drive sent with 25 stranded tags is accepted and deletes all 25, in order. It fails with the cap (422) and passes without.
- Tags are still validated: the 64-character pattern and the rule against listing the drive's own tag are untouched, and their tests pass along with the delete-on-redrive ones. Ruff clean.
- The generated web schema never carried the cap, so nothing regenerates.

</details>

## Checklists

- [ ] New tests have been added to any work in /lib (no library code changed)


app/desktop/studio_server/eval_builder_api.py
```diff
@@ -198,7 +198,6 @@ class ReplaceBatchTagsField(BaseModel):
 
     replace_batch_tags: list[str] = Field(
         default_factory=list,
-        max_length=20,
```
chiang-daniel, line 201: **Removed: the cap on the cleanup list.** A defensive bound sized by guess; the only cost it bounds is one pass over the task's runs per tag after a successful drive, and it made the twenty-first retry a permanent 422. Each tag is still validated below, and the field is shared by both drive requests.

leonardmq (approved):

PR #1801, opened 2026-09-21
Build the builder's eval through the shared spec-eval factory

chiang-daniel:
Refactor, no behaviour change. The eval builder hand-rolled its own `Eval`, duplicating the template, score key, priority, status, golden filter and split shape that `build_spec_eval` already derives for every other spec-eval creation path.

`spec_eval_splits` gains a source per split, typed `ItemSource`, so the factory can express what the builder already does: its test cases are EvalInputs it mints, while train and val are runs in the dataset. The mapping is an exhaustive match, so a third source is a type error.

The saved eval is identical, and the existing save tests pass untouched, including the one asserting the on-disk splits. `spec_api` and `eval_api` take the defaults.

Draft: body to be rewritten by Daniel.


chiang-daniel:
Folded into #1802, which now carries both commits into eb-v2-merge.

PR #1802, opened 2026-09-21
[EB-V2][Bug] Fix test split overlaps with train & val set. 

chiang-daniel:
## What does this PR do?

When I migrated Eval data from TaskRun to EvalInput everything was accidentally tagged as "test" causing an overlap with "train" & "val" set. In addition, a golden (human reviewed one) was tagged on runs whose cases were also sitting in test/train/val. Main never had this issue because golden was taken out before the split. There are 0 overlaps between the cases. 

* Fixed by dealing eval input into exactly one of train / test / val / golden set, followed by 40/25/25/10 split. train and val are EvalInput-backed now. 
* Removed the 25% golden cap and dropped the golden count from `min(10, N/4)` to 6 after discussion with team. 
* Added new protocol "SaveCase" which carries run_id so we know which one is reviewed.  
* Eval builder now creates its eval through `build_spec_eval` like every other creation path. No behavior change.
* Updated tests and comments.


tawnymanticore:
looks fine. only nit is the split is unseeded

>The deal is unseeded. random.Random() with no seed means the same batch saved twice lands differently. Harmless in production, but a seed in the request (or derived from the batch tag) would make saves reproducible, which matters when a loop re-creates an eval.

tawnymanticore (approved):

sfierro (approved):

PR #1803, opened 2026-09-21
fix: explain an unbuilt web UI instead of 500ing on it

leonardmq:
## What does this PR do?

The desktop server serves the web UI from `app/web_ui/build` — a gitignored artifact that nothing in the dev loop builds, refreshes, or validates. When that directory is empty (a fresh clone, or a build that never ran), every non-API request fell through to the 404 handler, which tried to serve a `404.html` that wasn't there either. The result was a 500 carrying a raw file-not-found path:

```
$ curl -s http://localhost:8757/
{"message":"File at path .../app/web_ui/build/404.html does not exist.", ...}
```

This PR serves a short page instead, explaining what this port actually serves, that `make ui` + http://localhost:5173/run is the hot-reloading dev UI, and the one command that builds this one.

the placeholder page

Details:

- **503, not 404.** The route is usually fine; there is just no UI to serve it with.
- **Same no-cache headers** as the rest of the web host, so a page left open picks the app up as soon as the build lands.
- **API paths untouched** — `/api/*` keeps its JSON 404s, covered by a new test.
- **CONTRIBUTING** now says that the Python server's port serves the last `npm run build` output and is not a way to develop against the UI. That was already implicit (the docs send you to :5173) but nothing said what the other port was doing.

### Found while debugging

`make dev` + the desktop port loaded a blank page. The cause turned out to be a *stale* build rather than a missing one: an old `app/web_ui/build` whose HTML bootstrapped `__sveltekit_147c0ot` while its JS chunks read `globalThis.__sveltekit_deqk1n.data`, so `kit.start()` threw before rendering anything. Rebuilding fixed it, and I couldn't reproduce the mismatch across four clean builds, so there is nothing to fix there. But it made clear how little this port tells you when what it's serving is wrong, which is what this PR addresses.

## Checklists


leonardmq:
Kind of useful for when you forget to run `npm run dev` on the UI and try to open the API server instead - but don't want to hijack root route for this

